PR #3, opened 2014-03-20
Allow clamdscan to have a separate config file; use clamd's if not found

otheus:
Very simply, clamdscan needs the possibility of its own configuration file and with a fallback to clamd.conf. Although users can specify this on the command line, it's much nicer for packaging and deployment that it has its own configuration file that can be used out-of-the-box so to speak. 

With this relatively simple code change, we manage that. There is a caveat: I cannot glean from parseopts how to determine if the configuration option was set by the provided command-line option or if the setting comes from the default setting. The work around is commented in the code.


micksola:
While the idea and work here are sound, and this may be a direction we wish to move in the future, it will likely be as part of a larger configuration re-work as we move closer to a 1.0 release. For those reasons, and given the time that has passed (which requires a revisiting of clamdscan options to ensure completeness), it is unlikely we will be including this PR in upstream.

PR #6, opened 2014-08-05
Add missing const keyword in cl_update_hash()

antchan2:
This aligns the signature for _cl_update_hash()_ with OpenSSL's _EVP_DigestUpdate()_ used by the implementation.


micah-at-talos:
Applied in https://github.com/Cisco-Talos/clamav-devel/commit/09d896fe686b15b5ddd3fca861b96a3717738a92

PR #8, opened 2014-09-05
Initial work to provide better-usable fanotify for on-access scanning under Linux..

mgjani:
Modifies two existing configuration options and adds a third to allow
user to specify whether or not file access should be prevented based
on on-access scan results.


vrtadmin:
This has been fixed in our Internal GIT

PR #18, opened 2015-03-11
iso9660: return CL_VIRUS if any part fails

MarshallMiller:
return CL_VIRUS if any part fails rather than if the last scanned file or directory fails
honor the SCAN_ALL setting


vrtadmin:
This has been fixed in our Internal GIT

PR #22, opened 2015-06-18
Adds a module to use Prelude and connect to a prelude manager

ningirsu:
To enable prelude compile with ./configure --enable-prelude

In ClamAV configuration file set Prelude Enable on yes and choose an analyzer name with PreludeAnalyzerName (default ClamAV).

You need to have a prelude manager to use this module.


vrtadmin:
This has been fixed in our Internal GIT.

PR #27, opened 2015-09-25
Bugfix for Exclude UID settings

atomicturtle:
This corrects the behavior of OnAccessExcludeUID to accurately exclude the UID from OnAccess scanning (Ex: OnAccessExcludeUID 0)


micah-at-talos:
Sorry this PR sat idle for so long.  We chose to resolve the UID 0 issue a different way.

PR #43, opened 2015-11-22
Ported solution and source code to Visual Studio 2015 (Bug 11378)

keithjjones:
Fixed bug 11378

https://bugzilla.clamav.net/show_bug.cgi?id=11378

Broken: Setup projects

Invalidates pull requests #40 #41 #42


vrtadmin:
This has been fixed in our internal GIT

keithjjones:
Is that internal Git pushed here?

vrtadmin:
We generally sync the GIT on releases and builds.

vrtadmin:
But I believe this has been synced in the 0.99.3 branch.

PR #51, opened 2016-04-21
Allow matching credit cards only and add additional card ranges.

user-none:


libclamav/dlp.c
```diff
@@ -128,42 +128,43 @@ static int ssn_max_group[MAX_AREA+1] = { 0,
...
+    {222100, 272099, 16, 16, 1, 1, "Mastercard 2016"},
+    {300000, 305999, 14, 16, 1, 1, "Diner's Club - Carte Blanche"},
+    {309500, 309599, 14, 16, 1, 1, "Diner's Club International"},
+    {340000, 349999, 15, 15, 1, 1, "American Express"},
+    {352800, 358999, 16, 16, 1, 1, "JCB"},
+    {360000, 369999, 14, 16, 1, 1, "Diner's Club International"},
+    {370000, 379999, 15, 15, 1, 1, "American Express"},
+    {380000, 399999, 16, 16, 1, 1, "Diner's Club International"},
+    {400000, 499999, 16, 16, 1, 1, "Visa"},
+    {500000, 509999, 16, 16, 1, 0, "Maestro"},
```
AlexanderS: The `0` should be the is_cc field? Isn't the 6th field in iin_map_struct the `luhn` field?

user-none:
The struct was updated with new members.

ROCKNROLLKID:
Any chance to get ranges for Capital One and Amazon, too? (Or am I just not seeing them?)

user-none:
What do you mean by Capital One and Amazon ranges?

Capital One's cards are Visa or MasterCard. Amazon is a Chase Visa card. Both Visa and MasterCard are included in the patch. The ranges provided are for the card brands themselves which the issuers (Capital One, Chase...) issue cards on behalf of. It's not necessary to include a break down per issuer because the entire range is reserved for the given card brand. While it would be nice to have a full breakdown of issuer's sub ranges it's very difficult to get this info and can/does change weekly. Also, a file broken down to the issuer level is going to be massive.

When you see things like Amazon or PayPal credit card it's though a partnership with a bank issuing cards for a given card brand. It's just branding. The Amazon card for example, is a Chase Visa with Amazon's logo on it. It's still a Chase Visa card and falls within Visa's bin range because it's a Visa card.

ROCKNROLLKID:
Ah alright. Nevermind then.

micksola:
Hi, could you provide a compelling use case for this feature that outweighs the inherent potential for misuse?

user-none:
> Hi, could you provide a compelling use case for this feature that outweighs the inherent potential for misuse?

I don't understand your question. ClamAV currently supports searching for credit cards and blocking them. This patch is an enhancement to that functionality because the current code detects many things that aren't credit cards and blocks legitimate messages.

I don't see how making detection of an existing feature has potential for misuse.

This change set does two main things:

1. Adds better bin range detection to determine if a number could really be a credit card. This decreases false positives. The main reason for this change.

2. By default the patch blocks credit and other financial cards. An option was added to only match on credit cards and not other financial cards. Some types of card, such as gift, do not fall under PCI which is one of the big reasons people want to block credit cards. So, someone who wants to block credit cards because of PCI may not want to block gift cards. Hence making additional card type checks optional.

As for misuse with #2, an admin would need to actively configure allowing non-credit cards through.

micahsnyder:
Hi @user-none , we just merged an updated version of this PR into the 0.103 development branch (dev/0.103): https://github.com/Cisco-Talos/clamav-devel/commit/a6a355629d5add17d0af9b66d95aeaefa74d0a8b

Sorry this took so long!  Thank you for contributing this new feature.

PR #52, opened 2016-05-06
xml.m4 : improve buildability on multi-arch OS distros

jimklimov:
The original m4 snippet was unusable on systems where libraries are not necessarily in verbatim "/lib/" subdirectories (e.g. a 64-bit build on Solaris).

NOTE: Many other m4's in this directory would welcome similar rehaul, but it can wait - they happen to be reasonable (vs. Solaris build tools at least), when dependency's installation prefix is verbatim "/usr".

It also helps that 64-bit binary directories are prioritized in PATH in the build recipe (so proper `*-config` scripts are found to apply correct architecture settings) - and now `xml.m4` can newly benefit from this too.

### DIFF
--- a/m4/reorganization/libs/xml.m4
+++ b/m4/reorganization/libs/xml.m4
@@ -2,52 +2,87 @@
 want_xml="auto"
 AC_ARG_ENABLE([xml],
 [AS_HELP_STRING([--disable-xml], [do not include DMG and XAR support])],
-want_xml=$enableval, want_xml="auto")
+    [want_xml=$enableval], [want_xml="auto"])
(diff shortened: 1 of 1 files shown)


m-sola:
Some functionality here was doubled up on in https://github.com/Cisco-Talos/clamav-devel/commit/3f0ed8d46095090e8d0a5f06009153e145d348f5 but the rest was added via https://github.com/Cisco-Talos/clamav-devel/commit/1a5a89e9eecab34817d444e04e884648508fdffa to work with pr #103 

Thanks for your work on this

PR #53, opened 2016-05-06
Spelling fix

ka7:
here patches for spelling-errors (comments, man-pages - not in variables, ..) , as found via misspell-fixer.
(there are some more in 3rd party libs, ( zlib..), but that has to be fixed upstream)

also did run a 'make' and 'make check' to ensure not to break too much...
Testsuite summary for ClamAV devel
- TOTAL: 13
- PASS:  6
- SKIP:  7
- XFAIL: 0
- FAIL:  0
- XPASS: 0
- ERROR: 0


vrtadmin:
This has been committed in our internal Git repo and will be public in the next release. (0.99.3)

PR #56, opened 2016-05-17
Allow freshclam downloadmanager to return FC_UPTODATE if no update was required

rkubik:
Useful for callers of this method to differentiate between a successful update or no update required.

### DIFF
--- a/freshclam/manager.c
+++ b/freshclam/manager.c
@@ -2893,5 +2893,5 @@ downloadmanager (const struct optstruct *opts, const char *hostname,
     if (newver)
         free (newver);
 
-    return 0;
+    return updated ? 0 : FC_UPTODATE;
 }


rkubik:
I see that this has been added in 0.99.3, thanks.

rma-x:
Russ, was it intentional that this does not only change the return code of downloadmanager(), but also the exit code of freshclam?

See also my comment on:
https://github.com/Cisco-Talos/clamav-devel/commit/5a0b539ca2

rkubik:
@rma-x Yes, changing freshclam's exit code was intentional however it was not captured in the release notes or man page - this was an oversight. Thank you for bringing this to our attention.

In version 0.101, the exit codes for freshclam will be reverted back to the original behaviour (prior to this patch - the libfreshclam downloadmanager changes will stay). The maintainers have expressed interest in reserving exit code 0 for success conditions and non-zero for errors. These changes will be documented in the release notes for visibility.

micah-at-talos:
The change for 0.101 to report 0 for success, non-zero for failure was just sync'd to Github. 
https://github.com/Cisco-Talos/clamav-devel/commit/d0b738a873cf136f6b012cce04b81ff3afa05df1

PR #57, opened 2016-06-29
clamscan: Add --jobs option for parallel scanning

michal42:
When --jobs=N is given, build a list of files first and then spawn N
children to scan the files in parallel. The main process writes indexes
of to-be-scanned files to a pipe and the children all read from the
other end of the pipe and scan the given files. No attempt is made to
synchronize the output of the children, so the --jobs option is best
used together with --infected.

Signed-off-by: Michal Marek mmarek@suse.com


michal42:
Superseded with PR#78

PR #64, opened 2016-10-28
Update the man page to reflect exit code behavior

oozie:
`freshclam` returns status 1, not 0, when the database is up-to-date as per the following line:  
https://github.com/vrtadmin/clamav-devel/blob/master/freshclam/manager.c#L2896


micah-at-talos:
Hi @oozie.  You may have noticed, we're quite behind on public PR's at the moment but are trying to rectify that.  I'm not certain if freshclam returning 1 instead of 0 was intentional.  We discussed the issue and reversed that behavior change in v0.101 so it returns 0 in both up-to-date and updated success cases.

PR #70, opened 2017-01-09
Fix broken zlib version detect

rbgarga:
A new zlib version 1.2.10 was released and is being detected as broken
due to the way version is checked. Add extra quote to the end of search
string to match entire version from zlib.h and fix false positive

### DIFF
--- a/configure
+++ b/configure
@@ -18718,9 +18718,9 @@ then
     as_fn_error $? "Please install zlib and zlib-devel packages" "$LINENO" 5
 else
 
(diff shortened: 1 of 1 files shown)


AlexanderS:
The autoconf source is already fixed in f0bcd186190fe6e67b3f0eaaceb7a99aa6a98865, but `configure` was not updated.

micah-at-talos:
configure eventually did get updated the next time autoreconf was run.

PR #73, opened 2017-02-21
Transforming the PNG checker into a PNG exploit searcher

gbmaster:
I removed the PNG validating parts from png.c and kept only the ones that I think are really useful to an antimalware use case. I wrote the detection for a couple of exploits I was aware of for this format.

Also, it scans the potential overlay inside a PNG file.


gbmaster:
Most likely, the two cli_append_virus calls can be removed. However, this code may still be useful to spot overlays

micahsnyder:
I meant to close this the day it merged... but we merged it! 

See: https://github.com/Cisco-Talos/clamav-devel/compare/9317e7872a69...ffd0f1357fc4

We ended up keeping the cli_append_virus calls.  I may tweak the name of the one which is missing "PNG" a little later, but otherwise I think they're good.  The overlay detection does seem like a great feature.

Thanks so much @gbmaster

PR #74, opened 2017-03-03
Making the GIF parser more tolerant and supporting GIF overlays

gbmaster:
I modified the already existing (but not used) GIF parser in order to make it more tolerant to GIF anomalies. As far as I could see, the most of the malicious content carried with GIF files is contained in the overlay: that's why I added the code in charge of scanning it.

In case something goes wrong during the parsing, instead of immediately quitting, the rest of the file is treated and scanned as overlay.


micahsnyder:
Hey @gbmaster!  I hope you're still doing well.  I know we spoke offline about this, but I wanted to thank you again publicly for the awesome work you've added.  Sorry this has taken this long for us to merge in your contribution.  We'll try to stay more on top of things in the future if you choose to continue working with ClamAV. :)

We merged an updated version of this GIF PR earlier last week: https://github.com/Cisco-Talos/clamav-devel/compare/50455664a715...f55acb436e7d

Updated versions of the GIF and PDF PRs are still in editing/review internally but are about down and should be merged any day now. 

Best,
Micah

PR #76, opened 2017-04-28
Move help before config file check, so it actually works

swelljoe:
Currently, if there isn't a clamd.conf in the default location (which there never will be, if following the documentation for setting up clamd, since it suggests setting up a per-service daemon), it'll print this error:

```bash
ERROR: Can't parse clamd configuration file /etc/clamd.conf
```

Moving help option above that check fixes the problem.

### DIFF
--- a/clamdscan/clamdscan.c
+++ b/clamdscan/clamdscan.c
@@ -74,6 +74,12 @@ int main(int argc, char **argv)
 	return 2;
     }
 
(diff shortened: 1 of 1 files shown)


m-sola:
Thanks for this, added to dev/0.102 via https://github.com/Cisco-Talos/clamav-devel/commit/f6bd5bf564a5d3c54c88bbce25fd9d6d87d9f2b1

PR #79, opened 2017-08-04
Make milter to respect $THREAD_LIBS instead of force -lpthread

rbgarga:
On generated configure clamav-milter is being forced to link using `-lpthread` instead of respecting a possible different thread library defined on `$THREAD_LIBS`. A current example of OS that doesn't support `-lpthread` is FreeBSD, that uses `-lthr`.


micah-at-talos:
Your pull-request was merged in here: https://github.com/Cisco-Talos/clamav-devel/commit/68afd226a895ffe22ec5230ad09bb24f017fe427

Thanks for submitting the improvement!

PR #80, opened 2017-08-11
This PR fixes linking for clam utils depending on libxml2

Redfoxmoon3:
Fixes building with no-undefined failure because the libclamav makefiles doesn't explicitly link with libxml2


micah-at-talos:
Merged this PR here: 
https://github.com/Cisco-Talos/clamav-devel/commit/1189b1fde3ea3090ba8bd918fad7554b27c9dcca

Thanks @Redfoxmoon3 !

Redfoxmoon3:
Apologies, but shouldn't the Makefile.am commit be merged, in case of autoreconf being run, re-generating the Makefile.in files?

micah-at-talos:
Yup, that's the parent commit: https://github.com/Cisco-Talos/clamav-devel/commit/d0872326cf9399f3d0d6f4101cd99e8ea9310555 I probably should've linked to both.

Redfoxmoon3:
Thanks! Sometimes my eyes are in the way:-)

PR #81, opened 2017-12-04
B11965:  Correct size of erroroffset on 64bit

candrews:
`erroroffset` is a pointer to `PCRE2_SIZE` type which according to the header file `pcre2-10.30/src/pcre2.h` is `#define PCRE2_SIZE size_t`. 

`size_t` on the amd64 architecture is 64 bits, so `erroroffset` cannot be an `int`.

Change `erroroffset` to `size_t` so it's always correct for all architectures.

Originally reported at https://bugs.gentoo.org/638932
Patch created by Jiří Moravec qjim@volny.cz


micah-at-talos:
Sorry about the delay.  Thanks for submitting the pull request. 

I was reviewing pull requests now that 0.100 is out and we're working on development towards 0.101. I found that this one is already covered by a similar commit.
https://github.com/Cisco-Talos/clamav-devel/commit/eaf0a436c19cf44fc93c5dad60da31ca85beb763

PR #84, opened 2018-02-16
libclamav: explicitly use top_srcdir .git dir

clandmeter:
This will prevent git from picking up a ref from parent git repository
and setting this as libclamav REVISION/version.


micah-at-talos:
@clandmeter - just finally merged this PR.  Sorry it took so long.  Here's the commit:
https://github.com/Cisco-Talos/clamav-devel/commit/1f4cec6197e

PR #88, opened 2018-07-13
Include filename in config file line reports

jsoref:
Not knowing the config filename is frustrating...

### DIFF
--- a/shared/optparser.c
+++ b/shared/optparser.c
@@ -1169,9 +1169,8 @@ struct optstruct *optparse(const char *cfgfile, int argc, char **argv, int verbo
 struct optstruct *optadditem(const char *name, const char *arg, int verbose, int toolmask, int ignore,
                             struct optstruct *oldopts)
 {
-	int i, err = 0, sc = 0, lc=0, line = 0, ret;
+	int i, err = 0, sc = 0, lc=0, ret;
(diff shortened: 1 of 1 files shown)


m-sola:
This fails to compile since the cfgfile var isn't available to the optadditem function.

micah-at-talos:
It looks to me like `optadditem()` has a definition and 2 declarations (`libfreshclam.h`, `optparser.h`), but is never actually used anywhere. 

@m-sola , do you object to removing `optadditem()` from the codebase?

micah-at-talos:
I semi-retract my remarks.  It must exist for libfreshclam users.  I wonder if it's actually used.

micah-at-talos:
Your approach works, @jsoref, thoug I would be happier removing `cfgfile` and `line` from `optadditem()`, honestly.  There is no file and there are no line numbers.

micah-at-talos:
I pulled in your PR here:
https://github.com/Cisco-Talos/clamav-devel/commit/57b0a260e1d2af1b2d3b179cfe0f76f9cd12e657
And made a slight modification here:
https://github.com/Cisco-Talos/clamav-devel/commit/76d0d93d4f11a43f237cce495765b0f95d4352d1

PR #90, opened 2018-12-05
Add configure option to support mempool while cross compiling

rickwang7712:
- Some ARM machines only have small memory like 256MB.
      If mempool function isn't enabled, the engine initialization
      time will be very long due to memory fragmentation.
      The fragmentation problem will make heap grow very fast and then use a lot of swap,
      which will slow down initialization process.
        - The initialization time difference:
           1hr(with mpool) vs. 5hrs(without mpool)

### DIFF
--- a/configure
+++ b/configure
@@ -976,6 +976,7 @@ with_libprelude_prefix
 with_libcurl
 enable_strni
 enable_largefile
+enable_mmap_for_cross_compiling
 '
       ac_precious_vars='build_alias
 host_alias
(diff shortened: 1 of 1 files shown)


micah-at-talos:
Thanks for the pull-request.  I noticed that you edited `configure`, which is an autogenerated file.  We use autoconf & automake, so you'll have to make the necessary changes to configure.ac instead.

rickwang7712:
Got it. Sorry for that.
I will change it.

PR #91, opened 2018-12-06
Add configure option to force enable mmap support while cross compiling

rickwang7712:
- If cross-compiling platform like ARM, the mmap support will be
      disabled directly, which indirectly disabled mempool function, too.
      Without mempool, the engine initialization time will be very long
      due to memory fragmentation. The fragmentation problem will make
      heap grow very fast, and results in using a lot of swap while
      running on low RAM machines. It will slow down initialization
      and scanning process.
    - Test result:
      The initialization time of using memory pool on 256MB ARM machine is
      faster than without mempool by 5-hour.

### DIFF
--- a/libclamav/scanners.c
+++ b/libclamav/scanners.c
@@ -3695,7 +3695,7 @@ static cl_error_t scan_common(int desc, cl_fmap_t *map, const char *filepath, co
         if (FSTAT(desc, &sb))
             return CL_ESTAT;
 
-        if ((size_t)(sb.st_size) > (size_t)(INT_MAX - 2))
+        if ((unsigned long long)(sb.st_size) > (unsigned long long)(INT_MAX - 2))
             return CL_CLEAN;
     }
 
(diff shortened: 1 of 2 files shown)


rickwang7712:
Hi @micah-at-talos,
We had discussed about this at https://github.com/Cisco-Talos/clamav-devel/pull/90
Now I changed the modification on m4 file which will be used to generate configure script.
Is it okay?

micah-at-talos:
I think so!  Will investigate shortly.  Thanks Rick!

rickwangsyno:
Hi @micah-at-talos,
Is there any update about this one?
Thanks!

rickwang7712:
Hi @micah-at-talos,
Is there any update about this one?
Thanks!

micah-at-talos:
Here ya go! https://github.com/Cisco-Talos/clamav-devel/commit/481cf18696f565266f470a6e05580b181548c837

PR #94, opened 2018-12-23
remove private header cltypes.h from clamav.h

pkrul:
cltypes.h should not be included with the libclamav API

### DIFF
--- a/libclamav/clamav.h
+++ b/libclamav/clamav.h
@@ -65,8 +65,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
-#include "cltypes.h"
-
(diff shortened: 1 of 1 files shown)


micah-at-talos:
Hi pkrul,

Unfortunately it isn't as simple as removing cltypes.h from clamav.h, as clamav.h now depends on types from stdint.h / inttypes.h included by cltypes.h.

As per https://blog.clamav.net/2018/12/libclamav-missing-headers-issue.html, we have a fix coming shortly in a patch release to replace cltypes.h with clamav-types.h.in which will generate clamav-types.h at configure time.

PR #95, opened 2019-01-09
Fix compilation without deprecated OpenSSL 1.1 APIs

neheb:
Initialization is implicit and not supported anymore.

### DIFF
--- a/libclamav/crypto.c
+++ b/libclamav/crypto.c
@@ -53,6 +53,10 @@
 
 #include <openssl/evp.h>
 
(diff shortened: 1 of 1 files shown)


micah-at-talos:
@neheb We just merged this change into the upstream private repo (github is a mirror).  Here's there's the commit: https://github.com/Cisco-Talos/clamav-devel/commit/57f4d1fd10eff859e1f6ca88e842bd721f21d0d4

Thanks for the fix so users don't have to enable deprecated OpenSSL APIs!

PR #96, opened 2019-01-29
fix failure of large file detection on 32-bit platforms

rickwang7712:
With FILE_OFFSET_BITS=64, st_size is 8-byte.
In this case, casting st_size to size_t, which is 4-byte on 32-bit machine,
will make the comparison go wrong.

Signed-off-by: rickwang <rickwang@synology.com>


micah-at-talos:
Rick, we just merged this in the upstream private repo (github is a mirror).
Here's the commit: https://github.com/Cisco-Talos/clamav-devel/commit/1dd0e3355a534adf8dda453088d96af605d4e12b

Thanks for your bug fix!

PR #97, opened 2019-02-17
Update sigtool.c

beita1:
- fix bug: fd open but no close,it makes handle is occupied

- The other problem which is worth discussing is  line1748  in sigtool.c
```
    if (!(ctx = convenience_ctx(fd))) {
        close(fd);
        free(dir);
        return -1;
    }
    if (cli_ole2_extract(dir, ctx, &vba)) {
        destroy_ctx(-1, ctx);
        cli_rmdirs(dir);
        free(dir);
        close(fd);
        return -1;
    }
```
i think cli_ole2_extract need to be judged by specific requirement clearly because when ctx is null  it will step in if branch and try to destroy null that will cause crash. it not a good design.Although it is be confirmed not null in line1743.

### DIFF
--- a/sigtool/sigtool.c
+++ b/sigtool/sigtool.c
@@ -1749,6 +1749,7 @@ static int vbadump(const struct optstruct *opts)
         destroy_ctx(-1, ctx);
         cli_rmdirs(dir);
         free(dir);
+        close(fd);
         return -1;
     }
     destroy_ctx(-1, ctx);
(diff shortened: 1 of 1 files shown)


beita1:
my english is rookie.sorry,hahahha

micah-at-talos:
I agree the error handling is not a good design.  However, if ctx == NULL, it should step in if branch on line 1743 and return on line 1746.

micah-at-talos:
Merged here:

https://github.com/Cisco-Talos/clamav-devel/commit/1e574d5422e2c686e8a79da0649cf7d6f08f902e

I will continue to investigate the related bugzilla ticket you submitted.  Thanks Chips!

PR #100, opened 2019-04-15
systemd - set TimeoutStartSec=240 for clamav-daemon

opoplawski:
clamd can take quite a long time to start - especially on VM guests with multiple guests starting up at once.  This has been sufficient for me.

### DIFF
--- a/clamd/clamav-daemon.service.in
+++ b/clamd/clamav-daemon.service.in
@@ -11,6 +11,7 @@ ExecStart=@prefix@/sbin/clamd --foreground=true
 # Reload the database
 ExecReload=/bin/kill -USR2 $MAINPID
 StandardOutput=syslog
+TimeoutStartSec=240
 
 [Install]
 WantedBy=multi-user.target


micah-at-talos:
Without setting TimeoutStartSec, the timeout defaults to DefaultTimeoutStartSec from the manager configuration file (90s / 1.5 minutes), so this I guess is extending to 4min to be safe.

4 minutes is probably generous for a system running the standard clamav signature set.  However, I believe that for users who pile on as many third party signature sets as they can, the signature load time can be upwards of 3-5 minutes.  It may be very excessive for most users, but I'd be tempted to go with 7 minutes ( `TimeoutStartSec=420` ) just to be extra safe.

micksola:
We'd very much like to include this (crediting you properly) with an increased timeout if possible. If 420 is too high, somewhere between 5-7 minutes would be preferred

opoplawski:
I don't see any trouble with setting this to be very long (7 minutes) - definitely want to have it cover all reasonable cases.

micksola:
Awesome! Are you okay with me making the modification to the patch on your behalf as a fixup commit?

opoplawski:
Absolutely

m-sola:
Merged here: https://github.com/Cisco-Talos/clamav-devel/commit/a4c2d3b7181de8e7c805634acd6bea5dcfe2b394

Thanks again

PR #101, opened 2019-05-05
add Hygon Dhyana support for clamav

hygonsoc:
To add Hygon Dhyana support in Clamav, we add in following parts:
* add dhyana detection in function sys::getHostCPUName() and X86Subtarget::AutoDetectSubtargetFeatures()
* add dhyana define in llvm/lib/Target/X86/X86.td

### DIFF
--- a/libclamav/c++/llvm/lib/System/Host.cpp
+++ b/libclamav/c++/llvm/lib/System/Host.cpp
@@ -291,6 +291,14 @@ std::string sys::getHostCPUName() {
     default:
       return "generic";
     }
+  } else if (memcmp(text.c, "HygonGenuine", 12) == 0) {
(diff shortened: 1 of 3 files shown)


micah-at-talos:
I have no way of testing if this is correct.  All I can do is build it on another system and verify that the build and behavior remain unchanged. 
Do you have a similar patch for the main LLVM project that has been verified/accepted that I can reference?

micksola:
Need an update (as above) to verify patch works before we can move ahead with inclusion.

micahsnyder:
Closing due to lack of activity, and because the PR doesn't align with future plans to drop our internal copy of LLVM and update to make use of a modern versions of system-installed LLVM.

PR #102, opened 2019-05-26
Fix buffer length check

flowerysong:
It is possible for `bm->offset` to be <= 0 and `(offset + bm->offset)` to be > 0, in which case the bounds check was incorrectly skipped, sometimes resulting in a segfault.

```
    Program terminated with signal SIGSEGV, Segmentation fault.
    #0  0x00007fea90598db0 in cli_bcomp_compare_check (
        f_buffer=0x7fea5c9e3a3e <error: Cannot access memory at address 0x7fea5c9e3a3e>, f_buffer@entry=0x7fea5c98c1ba "\001\030\001\030",
        buffer_length=buffer_length@entry=2590, offset=<optimized out>,
        bm=bm@entry=0x7fea7289f9c8) at matcher-byte-comp.c:720
```

### DIFF
--- a/libclamav/matcher-byte-comp.c
+++ b/libclamav/matcher-byte-comp.c
@@ -583,16 +583,13 @@ cl_error_t cli_bcomp_compare_check(const unsigned char *f_buffer, size_t buffer_
     opt      = bm->options;
 
     /* ensure we won't run off the end of the file buffer */
-    if (bm->offset > 0) {
-        if (!((offset + bm->offset + byte_len <= length))) {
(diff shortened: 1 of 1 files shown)


micah-at-talos:
@flowerysong thanks for the pull request.  We'll merge this in soon.  Fortunately there is no immediate crisis as the signature which triggered this issue has been pulled for now.

micah-at-talos:
We just merged your change here: https://github.com/Cisco-Talos/clamav-devel/commit/6c5805e47413034c92398ca919be903d80e63f87

Thanks @flowerysong !

PR #103, opened 2019-08-10
Avoid libxml checks if --disable-xml is used

cotequeiroz:
Cross compilation currently fails because libxml2 is picked up from the host even when using `configure --disable-xml`.

Signed-off-by: Eneas U de Queiroz <cotequeiroz@gmail.com>

### DIFF
--- a/m4/reorganization/libs/xml.m4
+++ b/m4/reorganization/libs/xml.m4
@@ -12,87 +12,85 @@ if test "X$want_xml" != "Xno"; then
                     @<:@default=/usr/local or /usr if not found in /usr/local@:>@]),
     [with_xml_val=$withval]
   )
-fi
 
-AS_IF([test "x$with_xml_val" = "xno"], [XML_HOME=""],
-  [test "x$with_xml_val" = "xyes"], [XML_HOME="/usr/local"],
-  [XML_HOME="$with_xml_val"])
+  AS_IF([test "x$with_xml_val" = "xno"], [XML_HOME=""],
+    [test "x$with_xml_val" = "xyes"], [XML_HOME="/usr/local"],
+    [XML_HOME="$with_xml_val"])
 
-AS_IF([test "x$XML_HOME" != "x"], [
-   AS_IF([test ! -x "$XML_HOME/bin/xml2-config"], [XML_HOME=""])
-   ])
+  AS_IF([test "x$XML_HOME" != "x"], [
+     AS_IF([test ! -x "$XML_HOME/bin/xml2-config"], [XML_HOME=""])
+     ])
 
-AS_IF([test "x$XML_HOME" = "x" -a "x$with_xml_val" = "xyes"], [
-   AS_IF([test -x "/usr/bin/xml2-config"], [XML_HOME="/usr"])
-   ])
+  AS_IF([test "x$XML_HOME" = "x" -a "x$with_xml_val" = "xyes"], [
+     AS_IF([test -x "/usr/bin/xml2-config"], [XML_HOME="/usr"])
+     ])
 
-if test "x$XML_HOME" != "x"; then
-  AC_MSG_RESULT([$XML_HOME])
-else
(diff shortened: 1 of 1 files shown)


micah-at-talos:
It looks like there is another PR to address similar issues: https://github.com/Cisco-Talos/clamav-devel/pull/52/files  Can you please review?

cotequeiroz:
Yes, I saw that PR, but it was open in 2016 and left unmaintained.  It changes a lot more than it needs to build.  I came up with a simpler solution:  I merely guarded the libxml checks with the `if test "X$want_xml" != "Xno"` line, by moving its `fi` to the very end, before the `AM_CONDITIONAL`, following the instructions in the `AM_CONDITIONAL` [manual](https://www.gnu.org/software/automake/manual/html_node/Usage-of-Conditionals.html): "If AM_CONDITIONAL is run conditionally (e.g., in a shell if statement), then the result will confuse automake." Then I've removed an extra check for `want_xml` that became redundant.  The rest of the huge diff is just indentation shift.
It's easy to see that with `git diff -U6 --ignore-all-space @~1`, and I should have posted it with my description to make this easier to review:
```diff
--- a/m4/reorganization/libs/xml.m4
+++ b/m4/reorganization/libs/xml.m4
@@ -8,15 +8,14 @@ XML_HOME=""
 if test "X$want_xml" != "Xno"; then
   AC_MSG_CHECKING([for libxml2 installation])
   AC_ARG_WITH([xml],
     AS_HELP_STRING([--with-xml@<:@=DIR@:>@], [path to directory containing libxml2 library
                     @<:@default=/usr/local or /usr if not found in /usr/local@:>@]),
     [with_xml_val=$withval]
   )
-fi

   AS_IF([test "x$with_xml_val" = "xno"], [XML_HOME=""],
     [test "x$with_xml_val" = "xyes"], [XML_HOME="/usr/local"],
     [XML_HOME="$with_xml_val"])

   AS_IF([test "x$XML_HOME" != "x"], [
      AS_IF([test ! -x "$XML_HOME/bin/xml2-config"], [XML_HOME=""])
@@ -81,18 +80,17 @@ if test "$working_xml" = "yes"; then
     AC_SUBST(XML_CPPFLAGS)
     AC_SUBST(XML_LIBS)
     AC_MSG_NOTICE([Compiling and linking with libxml2 from $XML_HOME])
   else
     if test "$want_xml" = "yes"; then
       AC_MSG_ERROR([****** Please install libxml2 packages!])
     else
-    if test "$want_xml" != "no"; then
       AC_MSG_NOTICE([****** libxml2 support unavailable])
     fi
-  fi
     XML_CPPFLAGS=""
     XML_LIBS=""
     AC_SUBST(XML_CPPFLAGS)
     AC_SUBST(XML_LIBS)
   fi
+fi

 AM_CONDITIONAL([HAVE_LIBXML2], test "x$HAVE_LIBXML2" = "xyes")
```
I'm no expert in `autoconf`, but I can give #52 a try.

cotequeiroz:
I've looked at #52.  TLDR: #52 has the same goal as 3f0ed8d, while this PR fixes a bug introduced by the latter.

#52 does not apply anymore because of 3f0ed8d.  In fact, #52 is more like it than like this PR; they both tried to accomplish the same goal: to find where the xml2 libs are located using standard paths.  This PR, OTOH, fixes a bug in 3f0ed8d that nullifies the `--disable-xml` option if libxml2 is found in the system.

One of nice things about #52 is that it tries to locate `xml2-config` in the default PATH.  This feature should be added, as the libxml2 setup should just work out of the box without having to make assumptions, if the build system is setup correctly.  3f0ed8d's problem is that it looks for xml2 libraries even when `--disable-xml` is given, while #52 took proper care of it.

m-sola:
This had been brought into the cobebase as of https://github.com/Cisco-Talos/clamav-devel/commit/ecfcb2fc1ff7f90e37dd57a64882d15cbc2d28da

Thanks for your work here

PR #104, opened 2019-08-19
Adds clamscan start and end dates on summary

jhon287:
It could be great to have the start and the end dates of the scan on the summary.

### DIFF
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -62,6 +62,7 @@ int main(int argc, char **argv)
     int ds, dms, ret;
     double mb, rmb;
     struct timeval t1, t2;
+    char buffer[26];
 #ifndef _WIN32
     sigset_t sigset;
 #endif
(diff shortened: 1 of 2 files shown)


clamscan/clamscan.c
```diff
@@ -184,6 +185,10 @@ int main(int argc, char **argv)
         rmb = info.rblocks * (CL_COUNT_PRECISION / 1024) / 1024.0;
         logg("Data read: %2.2lf MB (ratio %.2f:1)\n", rmb, info.rblocks ? (double)info.blocks / (double)info.rblocks : 0);
         logg("Time: %u.%3.3u sec (%u m %u s)\n", ds, dms / 1000, ds / 60, ds % 60);
+        strftime(buffer, 26, "%Y:%m:%d %H:%M:%S", localtime(&t1.tv_sec));
```
micah-at-talos: It would be better to use sizeof(buffer) here (and in the line below) rather than `26`, just in case there is need to change the size of `buffer`.

micah-at-talos:
This does seem like a nice feature for the summary.
I seems like this feature could also go into `clamdscan.c`.

jhon287:
All done :)

clamscan/clamscan.c
jhon287: You're right !

micah-at-talos:
Hi @jhon287 sorry about the delay.  

We did end up merging this, here: https://github.com/Cisco-Talos/clamav-devel/commit/83d6f64469213edba9aa39de4d19f554634330e5

Thanks for the submission!

PR #107, opened 2019-09-08
onas_...cleanup function return void pointer remove

kolbma:
The functions 
onas_cleanup()
onas_context_cleanup()
doesn't return anything so we need type void and not void*.


micah-at-talos:
Merged your commit, here: https://github.com/Cisco-Talos/clamav-devel/commit/8043cd45696b61d3f1d032e23292a651a868e8ab

Thanks for helping out!

PR #108, opened 2019-09-09
Fix using wrong linker for clamsubmit

kolbma:
Changing from Objective-C linker to C linker for clamsubmit

### DIFF
--- a/clamsubmit/Makefile.in
+++ b/clamsubmit/Makefile.in
@@ -679,7 +679,7 @@ installcheck-binPROGRAMS: $(bin_PROGRAMS)
 
 clamsubmit$(EXEEXT): $(clamsubmit_OBJECTS) $(clamsubmit_DEPENDENCIES) $(EXTRA_clamsubmit_DEPENDENCIES) 
 	@rm -f clamsubmit$(EXEEXT)
-	$(AM_V_OBJCLD)$(OBJCLINK) $(clamsubmit_OBJECTS) $(clamsubmit_LDADD) $(LIBS)
+	$(AM_V_CCLD)$(LINK) $(clamsubmit_OBJECTS) $(clamsubmit_LDADD) $(LIBS)
 
 mostlyclean-compile:
 	-rm -f *.$(OBJEXT)


micah-at-talos:
Hi @kolbma 

The file you changed ( clamsubmit/Makefile.in ) is a generated file.  This change would not have any lasting effects.   Can you describe what issue you're facing that prompted you to make the change?

kolbma:
@micah-at-talos

It is the wrong variable to build C code. This is for ObjectiveC (Mac).
And clamsubmit is C code.

If you set different compilers with environment variables CC and CXX,
e.g.
CC="ccache cc"
CXX="ccache g++"
then the build fails.

kolbma:
Or more concrete...

The OBJCLINK command line format doesn't work here... there is an error about to specify a tag.

OBJCLINK = $(LIBTOOL) $(AM_V_lt) $(AM_LIBTOOLFLAGS) $(LIBTOOLFLAGS) \
        --mode=link $(OBJCLD) $(AM_OBJCFLAGS) $(OBJCFLAGS) \
        $(AM_LDFLAGS) $(LDFLAGS) -o $@

Also if you set additional CFLAGS it is ignored in this OBJCLINK command.

This works...

LINK = $(LIBTOOL) $(AM_V_lt) --tag=CC $(AM_LIBTOOLFLAGS) \
        $(LIBTOOLFLAGS) --mode=link $(CCLD) $(AM_CFLAGS) $(CFLAGS) \
        $(AM_LDFLAGS) $(LDFLAGS) -o $@

micah-at-talos:
If I might ask, what system are you building for?  We haven't run into any issues like this in our own testing, though we also don't use ccache.

clamsubmit and libfreshclam on Mac make use of https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/shared/mac/cert_util_mac.m to obtain the certificates from the KeyChain for HTTPS cert validation. 

This line enables Objective C support for the project: https://github.com/Cisco-Talos/clamav-devel/blob/dev/0.102/configure.ac#L163 

For reasons unknown to me, autoconf does not allow the AC_PROJ_OBJC macro to be used inside a conditional (eg. "if OS == Mac").  I don't know of another way to get autoconf to enable Objective C support for just just Mac builds. 

I am curious though, do you have autotools ( autoconf, automake, m4, libtool, pkg-config ) available on your machine?  If so, can you try changing configure.ac, line 163 to this:

```
m4_ifdef([AC_PROG_OBJC], [AC_PROG_OBJC])
```

Then run `autoreconf -ifv` and then try the build again.

kolbma:
Might also be related because I do `--enable-static --disable-shared` builds.
The system is an openSUSE 15.1.

**libfreshclam** works, because there is set `AM_LIBTOOLFLAGS = --tag CXX` in `Makefile.am`.

This would be here also a solution for **clamsubmit**.

micah-at-talos:
@kolbma Those are good insights.  I'll try it out and let you know how it goes.

micah-at-talos:
I was able to reproduce the issue using `./configure CC="ccache cc" CXX="ccache g++"` on Mint Linux.  The static/shared thing didn't appear to make a difference.  Adding `AM_LIBTOOLFLAGS = --tag CXX` to `clamsubmit/Makefile.am` and running `autoreconf` before re-building seemed to resolve the *clamsubmit* build issue as you had hoped.  

I am presently running an internal pull-request through our build-acceptance test framework to make sure it doesn't cause any adverse side effects before we merge it in.

PR #109, opened 2019-09-20
etc: cleanup config files for standard terminal width

bleve:
Also use safe example.com addresses in examples.


micah-at-talos:
@bleve can I trouble you to make matching changes to the configs found under `win32/conf_examples/`?

bleve:
Now they should be in sync.

micah-at-talos:
> Now they should be in sync.

Thank you @bleve !

PR #110, opened 2019-10-02
configure.ac: fix building of clamav milter

bleve:


micah-at-talos:
Replaced by https://github.com/Cisco-Talos/clamav-devel/pull/111

PR #111, opened 2019-10-02
configure.ac: fix building of clamav milter

bleve:


micah-at-talos:
Thanks Tuomo!  We'll test this out for inclusion into 0.102.1!

prlw1:
Just wondering about the word "test" - the patch is clearly correct by inspection:
m4/reorganization/milter/enable.m4 defines the variable have_milter
m4/reorganization/milter/check.m4 uses the variable have_milter
configure.ac had better include enable.m4 before check.m4.

micah-at-talos:
Hi @bleve @prlw1 
We ended up fixing this in a slightly different way than this PR because the issue affected more than just milter.  

The issue is fixed for dev/0.103 here: https://github.com/Cisco-Talos/clamav-devel/commit/4e585c89eb7665cbc095a71a3e156899f78b95a3 and also in an identical commit in an internal branch for 0.102.1 that will be public once the security patch is released.

PR #112, opened 2019-10-10
test(1) uses = for string equality. (== is a bashism)

prlw1:


micah-at-talos:
Hi @prlw1 thanks for the PR. 
Is there a specific shell we should test to verify that this fixes broken behavior?

prlw1:
Isn't it the other way around: check that your bash still works after the patch?
(Read the test(1) man page otherwise)

prlw1:
ping? (This really is a trivial patch...)

micahsnyder:
Sorry @prlw1 , we've just been focused on other things.  Will put in an internal copy of this pr and run it through the test suite before merging.

micahsnyder:
We just merged an internal copy of your PR into dev/0.103.  See the commits here: 
https://github.com/Cisco-Talos/clamav-devel/compare/d5a733ef9031...54c766d1843e

@prlw1 Thanks for submitting this fix.

PR #78, opened 2017-06-20
clamscan: Add --jobs option for parallel scanning (v2)

michal42:
When --jobs=N is given, build a list of files first and then spawn N
children to scan the files in parallel. The main process writes indexes
of to-be-scanned files to a pipe and the children all read from the
other end of the pipe and scan the given files. No attempt is made to
synchronize the output of the children, so the --jobs option is best
used together with --infected.

Signed-off-by: Michal Marek <mmarek@suse.com>

Changes in v2:
 * Rebased onto current master
 * Fixed a memory leak in error paths in run_jobs()

PR #113, opened 2019-11-05
clamscan: Add --multiscan option for parallel scan

ematsumiya:
Adds -m/--multiscan=N option for clamscan to spawn N children to scan files in
parallel.

This is a rebase of PR #78 from Michal Marek (@michal42).

Signed-off-by: Enzo Matsumiya <ematsumiya@suse.de>

### DIFF
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -56,6 +56,7 @@ void help(void);
 struct s_info info;
 short recursion = 0, bell = 0;
 short printinfected = 0, printclean = 1;
+int multiscan_jobs;
 
 int main(int argc, char **argv)
 {
@@ -153,6 +154,8 @@ int main(int argc, char **argv)
         exit(2);
     }
 
+    multiscan_jobs = optget(opts, "multiscan")->numarg;
+
     memset(&info, 0, sizeof(struct s_info));
 
     gettimeofday(&t1, NULL);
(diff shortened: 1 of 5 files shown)


micah-at-talos:
Cool thanks @ematsumiya! Will investigate.

ematsumiya:
Hi @micahsnyder ! Any news here? Did you have the chance to review this PR? Thanks.

ematsumiya:
@micahsnyder can you provide an update here please? We really look forward to have this running internally, but we want to make sure upstream will accept it.

micahsnyder:
> @micahsnyder can you provide an update here please? We really look forward to have this running internally, but we want to make sure upstream will accept it.

Sorry @ematsumiya you've been waiting so long.  I haven't had much chance to review your PR yet.

micahsnyder:
Conceptually I like this feature.  However, I don't think this is the right implementation.  Using `fork()` to spawn off subprocesses to scan each file results in a huge amount of memory usage.  

In local testing on my laptop, I used `-m 4` so it would fork up to 4 subprocesses, though I was only scanning 3 files. 
The original processes uses some 870MB or so of RAM (just loading the signature databases). 
Each of the 3 subprocesses for scanning my 3 target files consumed about 430MB.  I believe this is because they each duplicated most of the loaded signatures in their own process spaces.

Eg:
image
 
I repeated this test with `-m 10` for a larger collection of files and see the same thing, which each subprocess using anywhere from 420MB -> 610MB or more of RAM.  

image

A good solution would use new threads instead of process forks and would share the scanning engine with each thread so as not to exhaust system memory. 

I will have to decline this PR.  If you have the time to re-engineer this using multithreading I'd be happy to review the new version.

michal42:
Using threads would indeed reduce the memory footprint a bit (the original implementation does not spawn a subprocess for each file, though). But guess what, after almost 3 years, I lost interest ¯\_(ツ)_/¯

micahsnyder:
@michal42 I don't blame you for losing interest.  Work on this project has been a constant game of playing catch-up while juggling requests for new features.  I'm sorry about the long wait.

PR #114, opened 2019-11-25
build: Use pkg-config to find libxml2

Polynomial-C:
Otherwise we get unneccesary linkage to icu and other unwanted stuff

Bug: https://bugzilla.clamav.net/show_bug.cgi?id=12202
Gentoo-bug: https://bugs.gentoo.org/661328
Signed-off-by: Lars Wendler <polynomial-c@gentoo.org>

### DIFF
--- a/m4/reorganization/libs/xml.m4
+++ b/m4/reorganization/libs/xml.m4
@@ -4,85 +4,29 @@ AC_ARG_ENABLE([xml],
 [AS_HELP_STRING([--disable-xml], [do not include DMG and XAR support])],
 want_xml=$enableval, want_xml="auto")
 
-XML_HOME=""
-xmlconfig=""
 if test "X$want_xml" != "Xno"; then
-  AC_MSG_CHECKING([for libxml2 installation])
-  AC_ARG_WITH([xml],
-    AS_HELP_STRING([--with-xml@<:@=DIR@:>@], [path to directory containing libxml2 library
-                    @<:@default=/usr/local or /usr if not found in /usr/local@:>@]),
-    [with_xml_val=$withval]
-  )
-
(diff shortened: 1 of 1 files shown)


micahsnyder:
Sorry about the long delay on this.  Will look at it for inclusion in 0.103.

micahsnyder:
@Polynomial-C It appears this change would break the build on many systems that don't have pkg-config.  I'd be more inclined to go the other way and try to adopt GNOME's official m4 files for finding dependencies like libxml2 (https://gitlab.gnome.org/GNOME/libxml2/blob/master/libxml.m4)

In local testing w/ my mac, I found that libxml2 detection with your PR (after running `autoreconf -ifv`) fails to detect a locally installed libxml2 in favor of one apparently found in /usr/local, despite using the `./configure --with-xml=<path/to/my/local/install>` option. 

Do you have a specific use case where adding pkg-config support to libxml2's m4 file fixes your build?

chewi:
I think a compromise would be reasonable here such as using pkg-config unless a non-yes/no value is given to `--with-xml`. Overlinking causes unnecessary breakage when libraries like ICU are updated. pkg-config also helps enormously when cross-compiling and in other odd situations.

micahsnyder:
@chewi That sounds like a good compromise to me.  I want to improve our m4 library detections but I can't break the build for downstream users on various platforms (again, heh).  

Do you or @Polynomial-C have the time to put together a new PR (or update and re-open this PR) with the described functionality?

orlitzky:
If pkg-config is there and if libxml2 supplies a *.pc file, you might as well use it. If you want to keep support for systems without it, the entire existing check can be copy/pasted into the "fail" branch of `PKG_CHECK_MODULES` that gets run when pkg-config isn't available or there's no *.pc file for libxml2.

PR #118, opened 2020-04-08
Scan embedded files when allmatch is specified

adepasquale:
Per the TODO note, I think this is the expected behavior when the "allmatch" option is specified.

### DIFF
--- a/libclamav/scanners.c
+++ b/libclamav/scanners.c
@@ -2818,10 +2818,8 @@ static int cli_scanraw(cli_ctx *ctx, cli_file_t type, uint8_t typercg, cli_file_
     ret = cli_fmap_scandesc(ctx, type == CL_TYPE_TEXT_ASCII ? 0 : type, 0, &ftoffset, acmode, NULL, refhash);
     perf_stop(ctx, PERFT_RAW);
 
-    // TODO I think this causes embedded file extraction to stop when a
-    // signature has matched in cli_fmap_scandesc, which wouldn't be what
-    // we want if allmatch is specified.
-    if (ret >= CL_TYPENO) {
+    // If allmatch is specified, we need to scan embedded files as well.
+    if ((ret >= CL_TYPENO) || (ctx->options->general & CL_SCAN_GENERAL_ALLMATCHES)) {
         perf_nested_start(ctx, PERFT_RAWTYPENO, PERFT_SCAN);
         ctx->recursion++;
         fpt = ftoffset;


micahsnyder:
Hi Andrea,

Unfortunately the problem isn't quite so simple.  The issue is that `cli_fmap_scandesc()` was set up to return:
- either a `cl_error_t` enum 
- a file type `cli_file_t` enum

If one or more file types are detected, then we need to scan it with one of the file type parsers. The problem is that `cli_fmap_scandesc()` isn't capable of returning both `CL_VIRUS` and a file type.  You may have noticed  this function is responsible for a compile time warning about mixing enums.  The proper solution would be to change the arguments  to `cli_fmap_scandesc()` to have the the `cli_file_t` file type match as an out argument instead of sharing it with the `cl_error_t` return value.  Then this if-statement will have to be modified to something like:

```c
if (((CL_SUCCESS == ret) || SCAN_ALLMATCHES) && (CL_TYPE_ANY != detected_filetype)) {
```

micahsnyder:
Do you want to keep working on this, or should I close the PR?

adepasquale:
Well, I can give it a try :smile:

Would you rather have the additional file type returned as:

- a new struct containing both `cl_error_t` and `cli_file_t`; or
- a new `cli_file_t *detected_ftype` function parameter?

Thanks

micahsnyder:
>     * a new `cli_file_t *detected_ftype` function parameter?

This sounds good to me. 👍

paulroyal:
Hey Micah,

Would you let me know your thoughts on the following alternative fix? It aims to be simple while navigating the concerns you raised in your comment. 

```
@@ -2818,9 +2818,11 @@
     ret = cli_fmap_scandesc(ctx, type == CL_TYPE_TEXT_ASCII ? 0 : type, 0, &ftoffset, acmode, NULL, refhash);
     perf_stop(ctx, PERFT_RAW);
 
-    // TODO I think this causes embedded file extraction to stop when a
-    // signature has matched in cli_fmap_scandesc, which wouldn't be what
-    // we want if allmatch is specified.
+    /* if allmatch is specified, keep scanning even if we already have a hit */
+    if ((ctx->options->general & CL_SCAN_GENERAL_ALLMATCHES) && (ftoffset != NULL)) {
+        ret = ftoffset->type;
+    }
+
     if (ret >= CL_TYPENO) {
         perf_nested_start(ctx, PERFT_RAWTYPENO, PERFT_SCAN);
         ctx->recursion++;
```

micahsnyder:
Hi @adepasquale any luck with this?

micahsnyder:
@adepasquale will close out this PR for now to tidy up.  Feel free to file a new one if you solve it! 🙂

PR #119, opened 2020-05-18
Update clamonacc.c

ninoCan:
The clamonacc command doesn't present a `--debug` flag, but according to your blog https://blog.clamav.net/2019/09/understanding-and-transitioning-to.html the correct flag should be `--verbose`:
"[...]This is akin to clamd’s or clamscan’s --debug option, but isn’t quite so noisy as either of those. By default, clamonacc does not print any output after daemonizing, so you will have to pair this option with --log or --foreground to use it.[...]"

### DIFF
--- a/clamonacc/clamonacc.c
+++ b/clamonacc/clamonacc.c
@@ -64,7 +64,7 @@ static void onas_clamonacc_exit(int sig)
 {
     logg("*Clamonacc: onas_clamonacc_exit(), signal %d\n", sig);
     if (sig == 11) {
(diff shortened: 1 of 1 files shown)


micahsnyder:
Hi @ninoCan, sorry the the delay responding to your PR.  Nice find!  I'll stage your PR on our internal repo and merge it shortly.

micahsnyder:
Merged!  Here are the commits: https://github.com/Cisco-Talos/clamav-devel/compare/25105ad67b19...82e1c2a587a1

PR #120, opened 2020-05-26
RFC libclamav: use libarchive instead of libclamunrar

sebastianas:
This is PoC for libarchive support to access .rar files with the help of
libarchive. The version 3.4.2 was used for testing.
From ClamAV's testsuite the file
- clam-v2.rar
  failed to unpack.

- clam-v3.rar
  unpacked properly


micahsnyder:
Thanks Sebastian! I'll try to find some time soon to play with it with a larger set of RAR files and to step through the code.

micahsnyder:
Sorry @sebastianas , I didn't mean to close this. 

I deleted `dev/0.103` because it was merged and is done.  We're going to stop using the `dev/*` branches for feature version development in the future. So future PR's can all target `main` (replacement for `master`) at that time.  

I'm still up for having an alternate backend for RAR support, though it's taken me forever to find time to work on it.  I dont' seem to have an option to re-target this for `dev/0.104` and then re-open it.  Is that something github will let you do?  If not, I can cherry-pick the commit hash and make a new branch/PR for you.  That's fine too.

sebastianas:
Oh I was so happy :)
Don't worry. I will try to rebase the patch next weekend.
If you keep the libunrar folder, it would be nice if you could make it "independent" of everything else and have a custom configure script (or cmake) and everything. I meant to fix this since it broke again but I don't get around…

micahsnyder:
Haha if you don't want to work on it / want to be done on it -- I totally get that.  This task is hard for me to justify working on, as you can tell -- and adding alternative configurations exponentially increases the permutations that _should_ be tested.  So if you want to drop it for now, I get that.

By the way I'm planning to drop the autotools build tooling soon so I can do build system changes in CMake without having to support autotools too.  My main driver for this was for Rust toolchain support.  We'd already figured out the basics for doing it in autotools and cmake. But when looking into additional features like vendoring rust libraries at configure-time (so people can do offline builds) I decided it didn't make sense to continue supporting autotools.  I believe the cmake tooling is stable in `dev/0.104` now. I have a branch that removes autotools ready now, waiting for my team to review it. Hoping I get to merge it soon 🤞.

I'm not entirely sure I understand what you mean regarding an "independent" way to build libunrar.  If libarchive is an option, then a way to build libclamunrar_iface separately from the rest of ClamAV doesn't seem needed.

PR #121, opened 2020-05-31
freshclam: Fix execute() function on Windows in daemon mode

loblolly986:
Currently, freshclam's "OnUpdateExecute", etc., features will fail when running on Windows in daemon mode if the supplied command contains any arguments. This is because of incorrect usage of spawnlp() in the execute() function. This patch simply replaces it with a call to system(), which also makes it consistent with non-daemon mode on Windows and both modes on other platforms.

### DIFF
--- a/freshclam/execute.c
+++ b/freshclam/execute.c
@@ -53,7 +53,7 @@ void execute(const char *type, const char *text, int bDaemonized)
     }
 
 #ifdef _WIN32
-    if (spawnlp(_P_NOWAIT, text, text, NULL) == -1) {
+    if (system(text) == -1) {
         logg("^%s: couldn't execute \"%s\".\n", type, text);
         return;
     }


micahsnyder:
That's a nice find @loblolly986  and a very simple fix.  I'll give it a try.

micahsnyder:
Local testing before & after the test looks good to me. I tested using a simple `echo "hello world"` command, and saw it occur properly at each update interval..

I see that in non-daemonize mode it uses system().  I wonder why spawnlp was used before.  🤷 

I'll stage the PR on our internal git repo so we can merge it.

loblolly986:
I'd guess whoever implemented it that way thought it would be good for the function to not wait for the command to finish before returning (`system()` uses the OS command shell to launch the command and waits for the shell to terminate). But this is inconsistent with the behavior on other platforms, and they still did not read the documentation of `spawnlp()` carefully enough.

Note that because `system()` uses the command shell, it allows one to use shell-provided commands (`echo` being one), chain commands with `&&`, etc. If one is concerned about a command holding up freshclam too long, then on Windows you can use `start %COMSPEC% /C "<command>"` so that cmd.exe will not wait for it (note that the simpler `start <command>` only works properly if `<command>` does not need to be quoted and is not a shell-provided command or batch file).

micahsnyder:
Merged! Here are the commits:
https://github.com/Cisco-Talos/clamav-devel/compare/e01ba94e36f1...84caca8150c4

PR #122, opened 2020-05-31
libfreshclam: Fixes/improvements to progress display

loblolly986:
Despite the changes in commit f907e4b4d7e47b3992bd0613bbb71bd1185194d9, the freshclam progress display still tends to spill over to the next line in 80-column Windows consoles for anyone without a very fast connection (and perhaps even for them at times). I found that reducing `totalNumDots` further to 25 resolved this well enough, but I decided to improve the display further. This patch makes it also utilize format string field widths for consistent alignment (no shifting around as information changes), and more intuitive Xm XXs (or Xh XXm) time displays for values of 60 seconds or more (I kept the decimal seconds for values under 60 seconds). With my changes, it is practically assured that the progress display will not overflow to the next line in an 80-column console.

Other changes:

- Simplified a few calculations. The bytes display now also properly changes units at 1024 B/KiB instead of 2048 B/KiB.
- Fixed the progress bar not showing anything when `numDots` is 1.
- Replaced an instance of `printf(" ")` in the progress function with `fprintf(stdout, " ")` to be consistent with the other output calls.

### DIFF
--- a/libfreshclam/libfreshclam_internal.c
+++ b/libfreshclam/libfreshclam_internal.c
@@ -160,16 +160,30 @@ struct xfer_progress {
     CURL *curl;
 };
 
-static void printBytes(curl_off_t bytes)
+static void printTime(double seconds)
 {
-    if (bytes / (1024 * 1024) > 1) {
-        double megabytes = bytes / (double)(1024 * 1024);
-        fprintf(stdout, "%.02fMiB", megabytes);
-    } else if (bytes / 1024 > 1) {
-        double kilobytes = bytes / (double)(1024);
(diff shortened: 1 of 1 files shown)


micahsnyder:
Just tested out your changes and read over the diff.  It looks great to me.  Way cleaner than before.  Good idea!  I'll stage the PR on our internal git server. 

 Thanks for the contribution @loblolly986

micahsnyder:
Merged!  Here are the commits:
https://github.com/Cisco-Talos/clamav-devel/compare/84caca8150c4...3e94a1b7e6ce

PR #123, opened 2020-06-01
multithreaded database reloading

okias:
> What it does is offloading the DB load to a separate thread and only replace the current engine instance afterwards.
>
> That means that while reload is pending:
> - existing scan requests use the old db (this is unchanged)
> - new scan requests are honored instead of blocked and they also use the old db (this is new)
>
> After the reload is complete:
> - existing scan requests use the old db (this is unchanged)
> - new scan requests use the new db (this is unchanged)
>
> The existing engine is refcounted so it'll be eventually freed when no longer in use.
>
> Reload requests while reload is pending are silently ignored (i.e. I never fork more than a single reload thread). I think this is perfectly acceptable as the reload time is a few seconds (10-60) while the db update frequency is several hours. If someone has stupidly short release cycles they can still rely on SelfCheck.

Closes: https://bugzilla.clamav.net/show_bug.cgi?id=10979

@jendis

### DIFF
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -552,7 +552,7 @@ int main(int argc, char **argv)
             break;
         }
 
-        if ((ret = statinidir_th(dbdir))) {
+        if ((ret = statinidir(dbdir))) {
             logg("!%s\n", cl_strerror(ret));
             ret = 1;
             break;
@@ -744,7 +744,7 @@ int main(int argc, char **argv)
             break;
         }
 
-        ret = recvloop_th(lsockets, nlsockets, engine, dboptions, opts);
+        ret = recvloop(lsockets, nlsockets, engine, dboptions, opts);
 
     } while (0);
 
--- a/clamd/server-th.c
+++ b/clamd/server-th.c
@@ -63,12 +63,33 @@
 
 #define BUFFSIZE 1024
 
+typedef enum {
+    RELOAD_STAGE__IDLE,
+    RELOAD_STAGE__RELOADING,
+    RELOAD_STAGE__NEW_DB_AVAILABLE,
+} reload_stage_t;
+
+struct reload_th_t {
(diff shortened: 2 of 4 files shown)


micahsnyder:
Hi @okias we haven't forgotten about https://bugzilla.clamav.net/show_bug.cgi?id=10979

I fully intend to merge some variant of the patches provided.  I promise!  The reason we haven't simply merged one of the solutions is that I believe that the patch will cause problems for people who are limited on RAM and cannot afford for clamd to load the signature databases into RAM twice.  The current signature set uses around 910MB of RAM.  A multithreaded reload results in twice that, nearly 2GB of RAM used.  Here's an example multithreaded reload viewed in activity monitor on my laptop.  I didn't capture the highest peak because the activity monitor view isn't particularly accurate, but I think it conveys the point:
image

I believe we should provide a clamd engine option that will allow users to choose to block during a reload to reduce RAM requirements.  This should help folks trying to run clamd in a Docker container in a cloud or on say a Raspberry Pi. 

My plan is to start work on this option in the coming weeks.

As a side note, your commit appears to be Alberto Wu's work and if we were to accept such a PR it should give him credit as primary author and give credit to those who worked on the issue before him and provided similar patches in the past.  

Also, IIRC there is a race condition that may or may not cause a feature test failure when you run `make check`.  I don't see any changes to the check tests to account for it.  See the patch attached to the bugzilla bug from Arjen de Korte for details. 

Anyways, please continue to be patient.  I fully intend to include this feature improvement in the 0.103 version.  It's one of the top remaining items on our task list.

okias:
Thanks for explanation and care.

For now our company (Excello) will keep using this patch, but when you have new version more close to stuff which gets accepted, we're able to give you testing on our high-load ClamAV instances.

micahsnyder:
Thanks @okias will do!

PR #124, opened 2020-06-06
libfreshclam: Avoid division-by-zero + further optimization in progress function

loblolly986:
I didn't notice it in time to include it in my last pull request, but it looks like division by zero could occur where the value of `remtime` is calculated and assigned in `xferinfo()`. I haven't run into any divide-by-zero errors myself, but considering that the function already handles `fractiondownloaded` being less than or equal to 0.0 when printing the time values, I'm assuming it's possible. This patch simply moves the calculation/assignment so it will only be done if `fractiondownloaded` is greater than 0.0 (`remtime` is otherwise unused anyway).

While I was at it, I also removed the handling of `NowUploaded`/`TotalToUpload` and the `if` statements altogether where it calls `printBytes()`. The rest of the function is not coded to handle uploads, and the function already returns early if `TotalToDownload` is less than or equal to 0.0, so this code was simply reducing its efficiency.

### DIFF
--- a/libfreshclam/libfreshclam_internal.c
+++ b/libfreshclam/libfreshclam_internal.c
@@ -215,7 +215,6 @@ static int xferinfo(void *prog,
     curl_easy_getinfo(curl, TIMEOPT, &curtime);
 
     xferProg->lastRunTime = curtime;
-    remtime               = (curtime / fractiondownloaded) - curtime;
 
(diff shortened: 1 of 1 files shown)


micahsnyder:
Looks good to me.  Good catch, and I agree it makes sense to get rid of the upload bit.

PR #125, opened 2020-06-07
clamdtop: Fix header label alignment; tweaks to better fit 80-column windows

loblolly986:
Currently, the `DBVER` and `DBTIME` header labels are no longer correctly aligned, because of current (as of 0.100.x) ClamAV version numbers being wider. This patch fixes this.

This patch also tweaks the printing of primary thread counts and header values so they fit in 80-column windows (as best as possible; the `h` suffix of the `DBTIME` hour value still doesn't fit, but at least the value does). Previously, an 83-column window was required for the information to fit. (As a side effect, the alignment of the thread counts relative to the bar underneath looks cleaner.)

### DIFF
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -505,12 +505,15 @@ static void print_con_info(conn_t *conn, const char *fmt, ...)
     va_list ap;
     va_start(ap, fmt);
     if (stats_head_window) {
-        char *buf = malloc(maxx);
+        char *buf = malloc(maxx + 1);
+	char *nl = NULL;
         OOM_CHECK(buf);
-        memset(buf, ' ', maxx);
-        vsnprintf(buf, maxx - 1, fmt, ap);
+        memset(buf, ' ', maxx + 1);
+        vsnprintf(buf, maxx + 1, fmt, ap);
+	if ((nl = strchr(buf, '\n')) != NULL)
+            *nl = ' ';
         buf[strlen(buf)] = ' ';
-        buf[maxx - 1]    = '\0';
+        buf[maxx]    = '\0';
         wattron(stats_head_window, ERROR_ATTR);
         mvwprintw(stats_head_window, conn->line, 0, "%s", buf);
         wattroff(stats_head_window, ERROR_ATTR);
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -130,6 +130,7 @@ static void exit_program(enum exit_reason reason, const char *func, unsigned lin
     } while (0)
 
 static struct global_stats global;
+static SCREEN *curses_scr  = NULL;
 static int curses_inited   = 0;
 static int maxystats       = 0;
 static int detail_selected = -1;
@@ -290,7 +291,12 @@ static void init_ncurses(int num_clamd, int use_default)
 {
     int default_bg = use_default ? DEFAULT_COLOR : COLOR_BLACK;
     int default_fg = use_default ? DEFAULT_COLOR : COLOR_WHITE;
-    initscr();
+
+    /* newterm() allows us to free curses-allocated memory with delscreen() */
+    if (!(curses_scr = newterm(NULL, stdout, stdin))) {
+        fprintf(stderr, "Failed to initialize curses\n");
+        exit(EXIT_FAILURE);
+    }
     curses_inited = 1;
 
     start_color();
@@ -427,6 +433,7 @@ static void cleanup(void)
         }
         rm_windows();
         endwin();
+        delscreen(curses_scr);
     }
     curses_inited = 0;
     for (i = 0; i < global.num_clamd; i++) {
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -895,7 +895,7 @@ static void output_queue(size_t line, ssize_t max)
             if (filtered_tasks[i].line + 15 > cmde)
                 cmd[cmde - filtered_tasks[i].line] = '\0';
             if (filstart) {
-                size_t oldline = (line += i);
+                size_t oldline = line;
                 char *nl = strrchr(++filstart, '\n');
                 if (nl != NULL)
                     *nl = '\0';
@@ -908,6 +908,7 @@ static void output_queue(size_t line, ssize_t max)
                 line = getcury(stats_window);
                 if (line > oldline)
                     max -= line - oldline;
+                line++;
             }
         }
     }
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -510,7 +510,7 @@ static void print_con_info(conn_t *conn, const char *fmt, ...)
         OOM_CHECK(buf);
         memset(buf, ' ', maxx + 1);
         vsnprintf(buf, maxx + 1, fmt, ap);
-        if ((nl = strchr(buf, '\n')) != NULL)
+        if ((nl = strrchr(buf, '\n')) != NULL)
             *nl = ' ';
         buf[strlen(buf)] = ' ';
         buf[maxx]        = '\0';
@@ -880,12 +880,10 @@ static void output_queue(size_t line, ssize_t max)
     wattron(stats_window, COLOR_PAIR(queue_header_color));
     mvwprintw(stats_window, line++, 0, "%s", queue_header);
     wattroff(stats_window, COLOR_PAIR(queue_header_color));
-    if (max >= j)
-        max = j;
-    else
+    if (max < j)
         --max;
     if (max < 0) max = 0;
-    for (i = 0; i < max; i++) {
+    for (i = 0; i < j && i < max; i++) {
         char *cmde;
         assert(tasks);
         cmde = strchr(filtered_tasks[i].line, ' ');
@@ -897,20 +895,26 @@ static void output_queue(size_t line, ssize_t max)
             if (filtered_tasks[i].line + 15 > cmde)
                 cmd[cmde - filtered_tasks[i].line] = '\0';
             if (filstart) {
-                ++filstart;
+                size_t oldline = (line += i);
+                char *nl = strrchr(++filstart, '\n');
+                if (nl != NULL)
+                    *nl = '\0';
                 if (detail_selected == -1 && global.num_clamd > 1)
-                    mvwprintw(stats_window, line + i, 0, "%2u %s", filtered_tasks[i].clamd_no, cmd + 1);
+                    mvwprintw(stats_window, line, 0, "%2u %s", filtered_tasks[i].clamd_no, cmd + 1);
                 else
-                    mvwprintw(stats_window, line + i, 0, " %s", cmd + 1);
-                mvwprintw(stats_window, line + i, 15, "%10.03fs", filtered_tasks[i].tim);
-                mvwprintw(stats_window, line + i, 30, "%s", filstart);
+                    mvwprintw(stats_window, line, 0, " %s", cmd + 1);
+                mvwprintw(stats_window, line, 15, "%10.03fs", filtered_tasks[i].tim);
+                mvwprintw(stats_window, line, 30, "%s", filstart);
+                line = getcury(stats_window);
+                if (line > oldline)
+                    max -= line - oldline;
             }
         }
     }
     if (max < j) {
         /* in summary mode we can only show a max amount of tasks */
         wattron(stats_window, A_DIM | COLOR_PAIR(header_color));
-        mvwprintw(stats_window, line + i, 0, "*** %u more task(s) not shown ***", (unsigned)(j - max));
+        mvwprintw(stats_window, maxystats - 1, 0, "*** %u more task(s) not shown ***", (unsigned)(j - max));
         wattroff(stats_window, A_DIM | COLOR_PAIR(header_color));
     }
     free(filtered_tasks);
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -506,11 +506,11 @@ static void print_con_info(conn_t *conn, const char *fmt, ...)
     va_start(ap, fmt);
     if (stats_head_window) {
         char *buf = malloc(maxx + 1);
-	char *nl = NULL;
+        char *nl  = NULL;
         OOM_CHECK(buf);
         memset(buf, ' ', maxx + 1);
         vsnprintf(buf, maxx + 1, fmt, ap);
-	if ((nl = strchr(buf, '\n')) != NULL)
+        if ((nl = strchr(buf, '\n')) != NULL)
             *nl = ' ';
         buf[strlen(buf)] = ' ';
         buf[maxx]    = '\0';
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -513,7 +513,7 @@ static void print_con_info(conn_t *conn, const char *fmt, ...)
         if ((nl = strchr(buf, '\n')) != NULL)
             *nl = ' ';
         buf[strlen(buf)] = ' ';
-        buf[maxx]    = '\0';
+        buf[maxx]        = '\0';
         wattron(stats_head_window, ERROR_ATTR);
         mvwprintw(stats_head_window, conn->line, 0, "%s", buf);
         wattroff(stats_head_window, ERROR_ATTR);
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -637,9 +637,9 @@ static int make_connection_real(const char *soname, conn_t *conn)
 {
     int s = -1;
     struct timeval tv;
-    char *port       = NULL;
-    char *pt         = NULL;
-    char *host       = pt;
+    char *port = NULL;
+    char *pt   = NULL;
+    char *host = pt;
     struct addrinfo hints, *res = NULL, *p;
     int err;
     int ret = 0;
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -1072,7 +1072,7 @@ static int output_stats(struct stats *stats, unsigned idx)
     i   = 0;
     if (sel && !stats->stats_unsupp) {
         memset(line, ' ', maxx + 1);
-        snprintf(line, maxx - 1, "Details for Clamd version: %s", stats->version);
+        snprintf(line, maxx + 1, "Details for Clamd version: %s", stats->version);
         line[maxx]         = '\0';
         line[strlen(line)] = ' ';
         wattron(win, COLOR_PAIR(queue_header_color));
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -55,8 +55,6 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/un.h>
-#endif
-#ifndef _WIN32
 #include <sys/time.h>
 #endif
 #include <assert.h>
@@ -629,7 +627,7 @@ static int make_connection_real(const char *soname, conn_t *conn)
     struct timeval tv;
     char *port       = NULL;
     char *pt         = NULL;
-    const char *host = pt;
+    char *host       = pt;
     struct addrinfo hints, *res = NULL, *p;
     int err;
     int ret = 0;
@@ -709,11 +707,11 @@ static int make_connection_real(const char *soname, conn_t *conn)
         break;
     }
 
-
     if (p == NULL) {
         ret = -1;
         goto done;
     }
+
 end:
     conn->sd = s;
     gettimeofday(&conn->tv_conn, NULL);
@@ -877,6 +875,7 @@ static int recv_line(conn_t *conn, char *buf, size_t len)
 static void output_queue(size_t line, ssize_t max)
 {
     ssize_t i, j;
+    int tasks_truncd            = 0;
     struct task *tasks          = global.tasks;
     struct task *filtered_tasks = calloc(global.n, sizeof(*filtered_tasks));
     OOM_CHECK(filtered_tasks);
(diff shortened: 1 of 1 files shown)


micahsnyder:
Hm this does seem like a slight improvement:
image

@loblolly986 you may have noticed that we like to give credit to contributors in our NEWS.md, which will also be included our release notes blog post.  Do you want it to continue to read "loblolly986" or is there another name you would prefer?

loblolly986:
@micahsnyder - you may use Zachary Murden in the credits.

Your screenshots pointed out to me another instance of a line being unnecessarily truncated two characters short of the screen edge (the "Details for Clamd version" line), so for consistency I've added a correction for it to this pull request. This particular fix won't be so noticeable in ordinary release builds, but we might as well make the behavior consistent.

I also ran into yet another instance of premature truncation (one character short) in the `print_con_info()` function, so I corrected that and also fixed its handling of messages with newlines when printing to the header (the entire line will now properly use the red background). You can see this by making clamd reload the virus database while clamdtop is running.

loblolly986:
I've revised the `print_con_info()` fix to use `strrchr()` instead of `strchr()` for (ostensibly) slightly better efficiency.

While we're at it, why not fix the awful handling of wrapped lines in the task list, as seen in the first screenshot above? A patch for this has also been added to this pull request.

Note that if not all tasks will fit and the last line in the list is wrapped text, the "`*** X more task(s) not shown ***`" bar will print over the first part of that line. I'm not sure if this looks "clean enough", or if we should make the bar print over the entire line, or move the bar elsewhere (e.g. to the status bar or task list header). Thoughts?

loblolly986:
My patch for the task list had a bug of its own. Corrected.

I also corrected the label alignments in the task list header, which previously were inconsistent between single- and multi-clamd modes (and partially off either way).

loblolly986:
Just thought I'd add that I recognize I've stretched this pull request a bit beyond its original purpose with the additional fixes. Unless you want me to revise the "X more task(s) not shown" popup (as discussed above), or you run into any more mistakes of mine that didn't show up in my own testing, I won't be adding any more to it.

micahsnyder:
No worries about stretching the PR a bit.  The UI tweaks are appreciated.  I see what you mean regarding the line wrap issue.  Do you know if we can disable line wrapping for the scan target files?  

Also, it appears another issue may have crept in with this PR.   The "host" field appears as `(null)`.  I tested with both Mac and Windows:

image

image

Unrelated: Have you been doing any testing with clamdtop on Windows?  We only just added clamdtop support for Windows in the dev/0.103 branch.  I find that clamdtop occasionally disconnects and fails to reconnect, but I haven't had time to really figure out why. 

Eg:
image

loblolly986:
@micahsnyder: We could truncate the lines at the screen edge instead of wrapping to the next line...  but arguably the line wrapping (when properly accommodated) improves usability of clamdtop in smaller windows much more than any of my other tweaks here. Not wrapping the lines without truncation would require implementing additional code for horizontal scrolling of the task list; IMO it would be a better idea to implement vertical scrolling, so that all tasks could be viewed even in a smaller window. We could print the COMMAND values in boldface (maybe also the QUEUEDSINCE values) to improve delineation between the tasks; the screenshot below illustrates this.

Oddly enough, I've been doing all my work/testing thus far on Windows, with personal 0.102.3 builds of the great unofficial Windows port by @sherpya (Gianluigi Tiesi) (https://github.com/clamwin/clamav-win32, https://github.com/clamwin/clamav/tree/0.102.3), which already includes clamdtop as of the 0.102.1 release. I prefer it because 32-bit builds are fully functional on all Windows versions back to NT 4.0 (frankly what has me interested in ClamAV in the first place), and it has a few compelling additional features (also I like the logo/icon much more than the official one :-) ). It is largely consistent with the official source, but has some additional Windows compatibility fixes and the additional features. Any fixes I submit here are things that apply to official ClamAV as well. Cisco's official Windows binaries work on my Vista system, but I'm not inclined to go acquire a crappy recent version of M$VC (and Windows) just to rebuild those, when with this I can use the free MinGW-w64 and gain much broader OS support.

I've now made a custom version of sherpya's repo rebased on the current 0.103 tree, plus my patches. The null HOST value isn't due to my patches, but due to a bug in commit 944ee2c024b6bb929b90ce0989aea30724ce5f12. I've added a fix to this PR (and also made the `curses_inited` value initially 0 as it should be, something that remains unfixed here). Wouldn't surprise me if this also fixes the reconnection problems you illustrate. My clamdtop 0.102.3 builds have seemed pretty stable, and clamdtop 0.103.0 (with this patch) doesn't seem any worse thus far.

clamdtop-0 103-linewrap-boldcmd

micahsnyder:
Hi @loblolly986 sorry about the delay -- I recently relocated and things have been very busy for me since. Nice find regarding the source of the null HOST bug.  Whoops!  

Gianluigi Tiesi recently gave us the thumbs up to incorporate some of his features into the official clamav project, such as the memory scanning and windows service features.  Unfortunately, our team doesn't have any interest in testing to maintain compatibility with older OS versions though we welcome contributions to support older versions or uncommon platforms/architectures.  For Windows we only go as far back as Windows 7 32bit.

In ClamAV 0.104, we will be writing some components in Rust - which I expect may cause some headaches building for uncommon operating systems/architectures.  We believe Rust is the right direction to write more secure code, however. 🤷

In other news, I'm actively working on adding CMake support so we can drop Visual Studio project files some day.  Here is my work-in-progress, if you're interested in giving it a try: https://git.vrt.sourcefire.com/micasnyd/clamav-devel/tree/CLAM-889-cmake-tooling-2  It currently builds on Windows and Posix systems, but I've broken the feature to dynamically determine the config & database directories on Windows.  Hoping to fix that this weekend.  I haven't done any testing with MinGW-w64 or Cygwin.  If you have any interest in validating my work, I would appreciate all the help I can get.  If I can get it stable next week (or at least so I haven't broken the autotools and Visual Studio builds) then we should be able to include it in 0.103 as an experimental build system.

micahsnyder:
I just played a bit with clamdtop with your latest changes and it looks very good.  The way it line wraps now is I think quite readable.

image

loblolly986:
@micahsnyder: No worries about the delay; I saw your mention of your move in another PR's comments and definitely understand.

I noticed that in the time since my last update here, the null HOST bug was also "fixed" in commit 4d6ca3e6d844d50c1f99d4f12f86348d7e34f400. I think my fix is better, because with mine the IP-storing code is placed under the "end" label and only executed upon success, whereas the other one still has it under the "done" label where it will be executed even upon failure. I'm going to try to update this PR so that my fix will replace the other one instead of conflicting with it.

Recently, I tried `clamdtop`'s multi-clamd mode for the first time, leading me to make a few additional UI tweaks/improvements that I will also be posting here.

Nice to hear that some of sherpya's enhancements will be finding their way to official ClamAV; does this include the "Loading virus database" spinning status indicator in `clamscan` and the graceful Ctrl+C termination?

CMake would certainly help with portability compiler-wise. Unfortunately, your link refuses to load for me (private repo?). It is worth mentioning that sherpya's port also uses CMake as of 0.102.1, and may be worth consulting, although note that it uses sherpya's own Windows compatibility files instead of the official ones (to some extent probably an artifact of his port having existed much longer than official ClamAV support for Windows) and does not bother with complexities of the official build system such as autotools or Mussels - it builds most of the library dependencies (apart from OpenSSL) itself.

I expect the introduction of Rust into the project would indeed introduce compatibility challenges with older OSes; from a little research this would chiefly be due to some usages of newer APIs in the Rust standard library, wherein the affected features would need to be modified to use more compatible implementations. (The Rust developers only seem to care about compatibility with the latest/officially-supported OSes, as is all too typical.) Nevertheless, I appreciate your willingness to accept contributions to improve compatibility with older systems and would begin looking into contributing improvements in that regard if I could get official ClamAV building on my end (via CMake and MinGW/MinGW-w64).

micahsnyder:
> CMake would certainly help with portability compiler-wise. Unfortunately, your link refuses to load for me (private repo?). It is worth mentioning that sherpya's port also uses CMake as of 0.102.1, and may be worth consulting, although note that it uses sherpya's own Windows compatibility files instead of the official ones (to some extent probably an artifact of his port having existed much longer than official ClamAV support for Windows) and does not bother with complexities of the official build system such as autotools or Mussels - it builds most of the library dependencies (apart from OpenSSL) itself.

Sorry about that!  I posted the a link to the wrong repo.  Here's the same branch on my public fork: https://github.com/micahsnyder/clamav-devel/tree/CLAM-889-cmake-tooling-2 

I wasn't aware that sherpya's port was using CMake now. I will take a look to get some ideas.  I certainly would like to do away with autotools.  Mussels is definitely not a required tool, but makes it easier to build dependencies without vendoring them in or relying on 3rd party lib downloads or system installed libs.  

I look forward to seeing your updates on clamdtop.  I staged your current work and was preparing to merge it, but will hold off until you update it again.

micahsnyder:
@loblolly986 Are you done/ready for review again?

loblolly986:
Yes; I've been a bit busy lately myself and didn't have time to get a comment up right after the commits.

I got my fork updated, but rebasing it (as opposed to merging) didn't turn out to be a good way to do it, as it reset the dates and IDs of my existing commits. Hopefully this won't cause much trouble on your end.

The latest addition addresses these issues:

- In multi-clamd mode, formerly the multi-clamd version of the task list header was being used both in the summary view and (incorrectly) the individual detail views. Now it loads both versions and uses the correct one for the current view.
- I replaced `#` with `NO` in the multi-clamd task header to be consistent with the clamd list header.
- Fixed the presence of an extraneous space between the "previous clamd" and "next clamd" status bar elements.
- Using clamdtop with an additional IP address highlighted the inadequacy of the existing position of the `HOST` field: the field is only sized big enough to contain a string the size of `localhost:3310` or `127.0.0.1:3310`, and anything bigger messes up the alignment of the fields that follow. My solution makes `HOST` the rightmost field; if a value is too long to fit, it overprints an ellipsis at the end of that line to signal to the user that they need a larger window.
- Doing the above also made the existing length of the `DBTIME` field more problematic, as it seemed desirable to me that everything fit within 80 columns at least when the user is simply monitoring a local clamd. The only decent solution I could come up with here is to display the time value in ISO 8601 format, which reduced the length by the necessary one character.
- I removed the `#ifndef _WIN32` statements around the `endwin()` call in `cleanup()`; I can't think of a good reason to not call this at exit on Windows. I'm assuming that this was simply a workaround for `curses_inited` erroneously being initialized to 1. Certainly it works OK here with my builds based on sherpya's port. Calling `delscreen(SP)` here might also be a good idea, to explicitly free memory allocated by the curses library (`endwin()` simply exits curses mode, which can be resumed with `refresh()`), but I'm unsure about compatibility with ncurses (might need term.h included for the `SP` variable?).

I'm also curious what you think about my idea to print the command names in the task list in bold text. This seems to me to enhance the readability a bit when there are wrapped lines, so I went ahead and added it here. I suspect it won't be so noticeable in your Windows Terminal tests, however, between the PDCurses "WinCon" port's limited text attribute support and the muted colors shown in your screenshots. But I'll mention that I've been thinking of proposing (in a separate pull request) that we use the "WinGUI" port provided by [Bill Gray's fork of PDCurses](https://github.com/Bill-Gray/PDCurses), which is what sherpya uses. WinGUI creates its own window to display the curses screen, allowing better compatibility with curses features on other platforms such as real bold/italic text, full support for window resizing, better color support, etc., and on all Windows versions back to 95/NT 4.0 as well.

clamdtop-multi-clamd-test-1
clamdtop-multi-clamd-test-2

micahsnyder:
No worries about the rebase -- we do a lot of rebasing. I've gotten very comfortable with git magic to clean up / update pull-requests. 

I agree that the bold print makes it easier to read.  
Putting the IP address on the other side makes sense. Good idea!

I am also unsure about the `#ifndef _WIN32` statements around the `endwin()` call. 🤷 

In local testing on mac, linux, and windows your changes all look really good. '

I'm definitely up for switching to Bill Gray's PDCurses.  That sounds much better.  Any gotya's I should know about when switching over the Mussels recipes?  Any code changes that would be required?

Mostly unrelated:  While testing this I discovered a fairly major issue in 0.102.4 and dev/0.103 wherein a clamscan (without `-r`) or clamdscan of a directory fails entirely because a check to get the realpath (`cli_realpath`) on Win32 uses `_wopen()` via our `safe_open()` function which cannot open a directory handle, and wouldn't work to get the realpath even if it could.  Looks like I'll be fixing that bug tomorrow or soon thereafter.  😣  

Anyways -- your changes all look good to me.  One request, there's a warning about the snprintf statement on line 1063, pre-existing, regarding a float format string (observed with gcc on ubuntu 18.04).

```
  CC       clamdtop.o
clamdtop.c: In function 'main':
clamdtop.c:1063:37: warning: '%7.3f' directive output truncated writing between 7 and 314 bytes into a region of size 6 [-Wformat-truncation=]
         snprintf(mem, sizeof(mem), "%7.3f", s);
                                     ^~~~~
In file included from /usr/include/stdio.h:862:0,
                 from clamdtop.c:39:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:64:10: note: '__builtin___snprintf_chk' output between 8 and 315 bytes into a destination of size 6
   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        __bos (__s), __fmt, __va_arg_pack ());
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  CCLD     clamdtop
```

There's another on line 1101, introduced in this PR, regarding comparing unsigned & signed integers (observed on macOS):
```
  CC       clamdtop.o
/Users/micasnyd/workspace/clamav-micasnyd-2/clamdtop/clamdtop.c:1101:13: warning: comparison of integers of different signs: 'int' and 'unsigned int' [-Wsign-compare]
    if (len > maxx) {
        ~~~ ^ ~~~~
```

By chance can I trouble you to fix these two warnings while you're cleaning up clamdtop?  😃

loblolly986:
Y'know, I was going to bring up the `cli_realpath()` issue but was hesitant as my testing was with sherpya's port (which uses his own `safe_open()` implementation that calls `_open()` instead of `_wopen()`, not to mention his own code for `cli_get_filepath_from_filedesc()` that I've now mostly rewritten to make it work properly on all NT versions) -- and it only just occurred to me before bed that I should try the new official 0.102.4 binaries. :-) If it helps, I could just put my fix up in a pull request, wherein we call `CreateFileW()` with `FILE_FLAG_BACKUP_SEMANTICS` to get a Windows handle to a file *or* directory, then pass this to `_open_osfhandle()` to get a C runtime file descriptor.

I see what's causing the clamdtop warnings and will add another commit to fix them.

Switching to WinGUI should be pretty straightforward, but a few things to note:

- It is expected that Bill's PDCurses fork will undergo a bit of a name change very soon to better distinguish it from the official PDCurses project (this is not a "fork" simply in the superficial GitHub-style sense!). I'm sure you will appreciate the easier differentiation as well.
- If possible, I highly recommend using the latest master code and not the old 4.1 release. There have been a number of fixes and improvements to WinGUI recently, some from yours truly (with a few more to come, albeit mostly not significant for clamdtop); notably, window resizing/moving now works without hanging the curses application for the duration. Plans are to publish a new 4.2 release soon, however.
- There are a few calls specific to PDCurses (and Bill Gray PDCurses) that need to be added to make clamdtop work optimally with WinGUI, but they can be wrapped with `#ifdef PDCURSES` so that ncurses builds won't be affected. I was going to supply these with my planned separate pull request.
- When a user runs a WinGUI program linked as a console application (which is what we want with clamdtop) from the Explorer or a desktop shortcut, they'll be stuck with both the WinGUI window and a console window (which is probably not what we want). To address this, my builds of clamdtop currently contain some additional Windows-specific code to "free" (detach from) the console after calling `init_ncurses()` and display any "abnormal program termination" errors thereafter in a message box. But I have an idea of how console-freeing (and re-allocating) functionality could be implemented in WinGUI itself as part of cleaning up its `endwin()` and `delscreen()` backends, which I will be working on and proposing to Bill's project (and which would render my aforementioned additional code in clamdtop mostly superfluous). I'll leave it up to you whether you'd want to go with my existing solution here for now or wait for my planned changes to WinGUI to be merged.

loblolly986:
OK, the first commit should take care of the warnings. The handling of the `MEM` value in the clamd list was quirky: it appears the intent was to display a decimal value if the value is small enough (< 100), but the format string's specified field width (which triggered the warning) caused the decimal value to always be truncated. So I've reworked it to round to value appropriately to fit the actual field width (and seeming original intent). I also ended up revising the handling of the memory stats in general to store the original `double` values (now only multiplying by 1000 and assigning to an integer for the `show_bar()` function) and round them to an integer for display. To me, rounding to the nearest value makes more sense than just truncating (which results in, say, 1.9 displaying as 1).

On another note, to follow up on my mention of `delscreen()` earlier: it looks like to use this in an ncurses-compatible fashion, we'd use `newterm(NULL, stdout, stdin)` instead of `initscr()`, saving the returned `SCREEN *` to a global variable, then pass that to `delscreen()` at exit. I've gone ahead and implemented this in the second commit. If this causes problems on Linux or macOS then I'll revert it. Admittedly, I've heard that most curses programs, as with clamdtop currently, just use `initscr()` and `endwin()` and let the OS clean up the rest, but I do like the thought of freeing what memory I can myself.

micahsnyder:
> If it helps, I could just put my fix up in a pull request, wherein we call `CreateFileW()` with `FILE_FLAG_BACKUP_SEMANTICS` to get a Windows handle to a file _or_ directory, then pass this to `_open_osfhandle()` to get a C runtime file descriptor.

I'm fairly certain I tried this elsewhere to use the directory HANDLE to try to determine the "real path" and that I was unable to use `GetFinalPathNameByHandleA` with a directory HANDLE to get the real path.  Instead of my actual directory, I only got `"C:\Users"`, IIRC.  I should take a closer look at this, I don't think that we actually need the real-path of a directory anyways as we don't provide verdicts on directories, we provide verdicts on files.  Ergo, we actually need the real-path for each file being scanned.  TL;DR, don't worry about it -- I will have to try to figure this out as soon as I'm able (hopefully I can work on it this weekend or at least next week). 

> I see what's causing the clamdtop warnings and will add another commit to fix them.

I built with your fixes, described in your follow-up comment.  Looks good.  No issues compiling or running on mac or linux as far as I can tell -- and the warnings are gone.  I agree with your assessment Re: truncating vs rounding :D.  I skimmed the code changes and they seem to make sense, but I didn't delve in line by line to follow the math.  Basic testing on mac/linux/windows all seems like everything is right though.  Very nice!  That was quite a bit of work, thank you! 🎉

> Switching to WinGUI should be pretty straightforward, but a few things to note:
>...
>  I'll leave it up to you whether you'd want to go with my existing solution here for now or wait for my planned changes to WinGUI to be merged.

Let's wait for the next release of Bill Gray's fork w/ name changes, etc, and include that in 0.104+.  I love what you've done in this PR, and I think it's in a good state for 0.103.  

I have to go do real-life things for a bit, but will re-stage your work in an internal PR to run through our test suite so I can merge it.  Will update/close this PR once it merges.

loblolly986:
Hmmm... while it wouldn't surprise me if `GetFinalPathNameByHandleA()` (which is documented as supporting directory handles) has been broken by yet another instance of Microsoft's poor quality control in the crap-show that is Windows 10, with my fix it seems to work fine on Windows Vista; sherpya's `cli_get_filepath_from_filedesc()` code for Vista and later (with a little tweak from me to use `VOLUME_NAME_DOS`) works pretty much the same as the official implementation, except it doesn't remove the `\\?\` prefix -- using escaped absolute paths (and the Unicode versions of Win32 functions that take a path as a parameter) is what you want on Windows NT OSes to bypass `MAX_PATH` and other restrictions. My rewrite of the XP-and-earlier code uses `NtQueryObject()` to get the NT object path and manually translates it to Win32-compatible format, and it works here as well (on both NT 4.0 and Vista).

If you'd like, I could put up the pull request anyway, and/or share one of my 0.103 builds based on sherpya's port for you to play with to see if it also fails to work on your end, and go from there. It's out of my wheelhouse to judge whether checking directory paths is needed from a security standpoint, but perhaps the easiest fix would leave in such checking,

micahsnyder:
Ok just merged this one, squashed into this commit: https://github.com/Cisco-Talos/clamav-devel/commit/ee60f31602ae8ff763908333d7221d092ee016c2

> Hmmm... while it wouldn't surprise me if `GetFinalPathNameByHandleA()` (which is documented as supporting directory handles) has been broken by yet another instance of Microsoft's poor quality control in the crap-show that is Windows 10, with my fix it seems to work fine on Windows Vista; sherpya's `cli_get_filepath_from_filedesc()` code for Vista and later (with a little tweak from me to use `VOLUME_NAME_DOS`) works pretty much the same as the official implementation, except it doesn't remove the `\\?\` prefix -- using escaped absolute paths (and the Unicode versions of Win32 functions that take a path as a parameter) is what you want on Windows NT OSes to bypass `MAX_PATH` and other restrictions. My rewrite of the XP-and-earlier code uses `NtQueryObject()` to get the NT object path and manually translates it to Win32-compatible format, and it works here as well (on both NT 4.0 and Vista).
> 
> If you'd like, I could put up the pull request anyway, and/or share one of my 0.103 builds based on sherpya's port for you to play with to see if it also fails to work on your end, and go from there. It's out of my wheelhouse to judge whether checking directory paths is needed from a security standpoint, but perhaps the easiest fix would leave in such checking,

If it's no trouble, I'd be interested to see how you do it and will definitely test it out.  As you may notice, my Win32-skills are mediocre.   Still, I am trying to bring Win32 support in line with Linux (hence adding clamsubmit in 0.102, and clamdtop in 0.103).  I'm also happy to take your patches to improve compatibility with older OS versions though we don't have the resources to develop & test for every OS version ourselves.

For security purposes the goal is to make sure the file scanned matches the file we alert on / the file we move or remove.  Checking the real path of directory before the scan wouldn't really matter because we really only care about it when we traverse the path from root with the assumption that our previous cli_realpath call has eliminated all symlinks, meaning that if we encounter a symlink -- someone has swapped out the directory between scan and quarantine.

Anyways, closing this PR.  I look forward to seeing your next one! 😜

PR #126, opened 2020-06-10
For review: clamd multi-threaded reload 0.103

micahsnyder:
I'm providing this multi-threaded reload PR for review, as per the request in https://github.com/Cisco-Talos/clamav-devel/pull/123

Associated bugzilla ticket: https://bugzilla.clamav.net/show_bug.cgi?id=10979

In addition to the patch crafted by Alberto Wu that provides the default non-blocking clamd database reload improvement, this PR includes an option to block on the reload to provide the behavior like you see in 0.102 and prior versions.  This option probably won't be very popular, but may be needed on some machines to reduce RAM usage. 

To try the blocking-reload, set `ConcurrentDatabaseReload no` in your clamd.conf config file.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,24 @@ ClamAV 0.103.0 includes the following improvements and changes.
 
 ### Major changes
 
+- Clamd can now reload the signature database without blocking scanning.
+  This multi-threaded database reload improvement was made possible thanks to
+  a community effort.
+
+  Non-blocking database reloads are now the default behavior. Some systems that
+  are more contrained on RAM may need to disable non-blocking reloads as it will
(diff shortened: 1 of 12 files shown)


okias:
slightly modified version (for 0.102.3) is deployed on one instance.

okias:
On second server (clamav 0.102.3; higher traffic & load) we noticed that clamav stopped removing /tmp files probably at some circumstances. Re-testing to ensure it's not caused by patch.

micahsnyder:
Just merged this feature into dev/0.103.

https://github.com/Cisco-Talos/clamav-devel/compare/cdbc833...34c3441bcd3a

PR #129, opened 2020-06-24
pdf: do not override pdf->fileIDlen if there is no new fileID.

clem1:


micahsnyder:
Hi @clem1 this bug-fix makes sense to me.  I am curious though - how did you encounter this issue?  It seems very specific!  

Is there a test case I can use to show fixed behavior?

micahsnyder:
https://github.com/Cisco-Talos/clamav-devel/compare/bbbbac807986...8db3476dfed2

Merged.  Thanks again :)

Do let me know how you found the bug, if you're allowed to share.  I am quite curious.

clem1:
Le ven. 17 juil. 2020 à 01:42, Micah Snyder <notifications@github.com> a
écrit :

> Hi @clem1 <https://github.com/clem1> this bug-fix makes sense to me. I am
> curious though - how did you encounter this issue? It seems very specific!
>
> Is there a test case I can use to show fixed behavior?
>

Hi Micah, thanks for the merge!

The issue was encountered in a real PDF which I cannot share unfortunately.
I managed to create this PDF which seems to be reproducing the issue on my
side.

clem1:
Sorry. Here is the [pdf](https://github.com/Cisco-Talos/clamav-devel/files/4937543/bla.pdf).

PR #130, opened 2020-07-09
Correct log output for AlertEncryptedArchive

jbiggar:
I believe there's a small but potentially confusing mismatch between log output and the options actually applied when `AlertEncryptedArchive` is set to `yes`.


micahsnyder:
Hi @jbiggar nice find.  Thanks for submitting this.

micahsnyder:
https://github.com/Cisco-Talos/clamav-devel/compare/619877890337...bbbbac807986

Merged!

PR #133, opened 2020-08-21
Add newlines at end of sources to satisfy strict warnings…

jimklimov:
…and reduce noise in build logs (people say C/C++ standard even requires/required them)


micahsnyder:
Thanks for the fix!

PR #136, opened 2020-09-03
Fix systemd error message.

sritd:
Fix systemd error message. It is recorded in journald on my linux system and can be corrected by two modifications.

Recorded error messages:
usr/lib/systemd/system/clamav-freshclam.service:13: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether.
/usr/lib/systemd/system/clamav-daemon.service:13: Standard output type syslog is obsolete, automatically updating to journal. Please update your unit file, and consider removing the setting altogether.

systemd version: systemd 246.3-1

Signed-off-by: Sven Rueß <github@sritd.de>


micahsnyder:
Thanks @sritd 

I just merged your fix here for inclusion 0.104: https://github.com/Cisco-Talos/clamav-devel/compare/e4e3149368d2...8909465ecbff

PR #138, opened 2020-09-18
Fix ck_assert_msg() call

opoplawski:


micahsnyder:
Thanks for the PR, good fix. :) I'll stage this to be merged on our internal repo.

micahsnyder:
@opoplawski just merged this: https://github.com/Cisco-Talos/clamav-devel/compare/2b46876dcccd...dce04e58fe3e

Thanks again!

rma-x:
It looks like this fix only made it into 0.104, but I think it should also be in 0.103.
And maybe it would have been better to use ck_abort_msg() instead of calling ck_assert_msg() with a constant expression.

PR #140, opened 2020-10-13
fix: typo in error message

loopingz:
Nothing really to explain just a simple typo in an error message


micahsnyder:
Merged here: https://github.com/Cisco-Talos/clamav-devel/commit/0bc61351e4781978bad670f9fca68425b9819d5e

PR #142, opened 2020-11-16
fix onaccess scan queue

golovach:
Hi
I faced a problem with tar application hang in case clamonacc is active. Tar is extracting a lot of small files from an archive and stops at some arbitrary point. The problem is not stable to reproduce.

I can explain it in the following way:

1. Consumer thread is waiting for condition variable which indicates that there are files in the queue. Once the condition is satisfied, the consumer starts dispatching the files to the worker threads in the thread pool.
2. While the consumer thread processes the files in the queue, producer can put few more items in the queue and fire the condition variable. 
3. Consumer thread stucks on waiting for condition variable nevertheless there are items in the queue. However, new items are not coming from tar application because it is waiting on verdict regarding the items in the queue.

The solution is to:
1. use a single mutex to guard the queue and the condition variable
2. signal condition variable before releasing mutex in producer thread
3. check if there are events in the queue before waiting for condition

Besides, some small things worth to mention:
1. pthread_testcancel() call is not required as pthread_cond_wait() does the check for thread cancelation.
2. lock seems to be not required in onas_scan_queue_exit() as by this time no one supposed to access the event queue
3. memory allocation in onas_queue_event() is done without locking the queue
4. dispatch an item to a worker thread is done without locking the queue (calling thpool_add_work())


micahsnyder:
Hi @golovach

Though I too wasn't able to reproduce the deadlock problem you encountered, I think this is some very nice work.

I played with your PR a bit and everything seems to work well. 

Some notes:
- Good observations Re: `pthread_testcancel()`. 
- I agree that the node allocation in `onas_queue_event()` doesn't need to be inside the mutex lock, just when it's added to the event queue.  
- I also agree that you don't need to lock the queue for the `thpool_add_work()` call.  The call to `jobqueue_push()` has it's own lock for the queue anyways.
- I'm not certain about your claim that `pthread_mutex_lock()` isn't needed in `onas_scan_queue_exit()`.  It would seem to me that other theads adding to the queue (eg the ddd thread) could still be active when the `onas_scan_queue_exit()` is invoked by the `pthread_cancel()` in `onas_clamonacc_exit()`. I can see an argument for removing the `pthread_mutex_unlock()` after destroying the thread pool, however, because no one should access the queue after it's gone.  Thoughts?

Off topic from this PR:  I wanted to ask if you've encountered [clamonacc crashes when removing large directory trees](https://bugzilla.clamav.net/show_bug.cgi?id=12625).  This issue causes the most trouble for me when using clamonacc.  We presently don't have time budgeted to work on it over here and you seem to have a knack for solving thread resource/mutex issue in clamonac.  I am wondering if you have time to look into this one for us.

golovach:
Hi
For the ddd thread you seems to be right. There is a chance that event is already fetched from fanotify, user then presses Ctrl+C, which causes the queue to be destroyed, and then ddd thread puts the event into the queue. I think a simple fix for this will be change the order of thread teardown in clamonacc.c (I will commit it shortly)

micahsnyder:
Merged here: https://github.com/Cisco-Talos/clamav-devel/compare/9ce1b70cd7b4...6bc4078f235c

Thanks @golovach

PR #143, opened 2020-11-26
Added %f to VirusEvent

vasile-papp:
Hello,
While configuring the event trigger in clamd.conf, the only documented variable was "%v" for the virus name.
I considered it necessary to include also the file name, to avoid having to search for it in the log file. By looking into the code I found that there are 2 environment variables created which can be used in the VirusEvent configuration, they were just not documented in the example config file.
In this pull request I created a "%f" option (clamd source, test and doc) that can be used in the VirusEvent string as a placeholder for the file name and added the documentation for the 2 environment variables in the example config file.
Thank you for your consideration,
Vasile

### DIFF
--- a/unit_tests/check_common.sh
+++ b/unit_tests/check_common.sh
@@ -383,7 +383,7 @@ test_clamd2() {
 
 test_clamd3() {
     test_start $1
-    echo "VirusEvent $abs_srcdir/virusaction-test.sh `pwd` \"Virus found: %v\"" >>test-clamd.conf
+    echo "VirusEvent $abs_srcdir/virusaction-test.sh `pwd` \"Virus found: %v in %f\"" >>test-clamd.conf
     echo "HeuristicScanPrecedence yes" >>test-clamd.conf
     start_clamd
     # Test HeuristicScanPrecedence feature
@@ -403,7 +403,7 @@ test_clamd3() {
     # Test VirusEvent feature
     run_clamdscan_fileonly $TOP/test/clam.exe
     test -f test-clamd.log || sleep 1
-    grep "Virus found: ClamAV-Test-File.UNOFFICIAL" test-clamd.log >/dev/null 2>/dev/null ||
+    grep "Virus found: ClamAV-Test-File.UNOFFICIAL in .*/clam.exe" test-clamd.log >/dev/null 2>/dev/null ||
 	{ cat test-clamd.log || true; die "Virusaction test failed"; }
 
     test_end $1
--- a/clamd/clamd_others.c
+++ b/clamd/clamd_others.c
@@ -108,7 +108,7 @@ void virusaction(const char *filename, const char *virname,
     const struct optstruct *opt;
     char *buffer_file, *buffer_vir, *buffer_cmd, *path;
     const char *pt;
-    size_t i, j, v = 0, len;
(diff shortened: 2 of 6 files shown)


micahsnyder:
Thanks for submitting this pr @vasile-papp 

This change makes sense to me, and I appreciate that you added the environment variables to the example config file. 

I, or one of my team members, will test/review this soon for inclusion in 0.104.

micahsnyder:
Hi @vasile-papp 

Your work seems good.  I only found one issue:

```c
-    grep "Virus found: ClamAV-Test-File.UNOFFICIAL" test-clamd.log >/dev/null 2>/dev/null ||
+    grep "Virus found: ClamAV-Test-File.UNOFFICIAL in $TOP/test/clam.exe" test-clamd.log >/dev/null 2>/dev/null ||
```

This won't work because `$TOP` may be a relative path while `clamd` now reports real paths.  When I test, the `$TOP/test/clam.exe` grep-string is actually something like this:

```
Virus found: ClamAV-Test-File.UNOFFICIAL in /home/micah/clamav-devel/build/unit_tests/../test/clam.exe
```

Instead of this, which is what ClamD VirusEvent should print:
```
Virus found: ClamAV-Test-File.UNOFFICIAL in /home/micah/clamav-devel/build/test/clam.exe
```

We could either drop the above change so it isn't tested, or find a way to check for the real path in the test instead of using `$TOP`.  As a heads up, we're in the process of switching from autotools to CMake.  CMake doesn't yet have test support, but that's going to merge in the coming week(s). You can preview it here: https://github.com/micahsnyder/clamav-devel/tree/CLAM-34-unit-test-overhaul-windows-support  Ideally if you do want the `%f` feature to be tested, the test should work with CMake+CTest too.  Our CTest support will use Python instead of bash, so we should be able to get the realpath pretty easy for that test.

Let me know what you want to do. 

-Micah

unit_tests/check_common.sh
```diff
@@ -403,7 +403,7 @@ test_clamd3() {
     # Test VirusEvent feature
     run_clamdscan_fileonly $TOP/test/clam.exe
     test -f test-clamd.log || sleep 1
-    grep "Virus found: ClamAV-Test-File.UNOFFICIAL" test-clamd.log >/dev/null 2>/dev/null ||
+    grep "Virus found: ClamAV-Test-File.UNOFFICIAL in $TOP/test/clam.exe" test-clamd.log >/dev/null 2>/dev/null ||
```
micahsnyder, line 406: This won't pass, because $TOP may be a relative path.

micahsnyder (changes requested):

unit_tests/check_common.sh
vasile-papp: Considering that the test is going to change in the future, I updated this piece to not actually check the complete path accuracy, just match the file name at the end of a path.

micahsnyder (commented):
This works for me.  Since we presently work on an internal git repo and github hosts our public mirror -- I've imported your work, rebased it, and am running it through our CI.  A local test on my machine passed after your update and I expect I'll merge it later this afternoon.

micahsnyder:
Merged here: https://github.com/Cisco-Talos/clamav-devel/compare/56969286a121...978fea6788d5

PR #144, opened 2020-12-05
autoit: Added 0x43 "VOLATILE" to autoit op 0x0  keywords

cw2k:
Before it was listed as 'UNKNOWN_43' VOLATILE
about 'UNKNOWN_00' : 
currently Autoit interpreter 3.3.14.5 will turn keyword_ID 0x0 into an empty String.
However it is maybe good to keep it to "UNKNOWN_00" to see when keyword_ID 0x0 gets used.

### DIFF
--- a/libclamav/autoit.c
+++ b/libclamav/autoit.c
@@ -454,7 +454,7 @@ const char *autoit_functions[] = {
     "WINWAITNOTACTIVE"};
 
 const char *autoit_keywords[] = {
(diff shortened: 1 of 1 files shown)


micahsnyder:
Looks like all of these changes are in https://github.com/Cisco-Talos/clamav-devel/pull/145, which @recvfrom is helping review.  Closing this PR as a duplicate.

PR #145, opened 2020-12-05
autoit: Decompiled directives are double prefixed

cw2k:
Example: #NoTrayIcon will result in ##NoTrayIcon. 
Background: Prefixbyte is already included in the string.
Added two other TODO at issues I noticed when doing a autoit decompiler in Python.
I can't test these changes nor I know how clamav scans the decompiled autoit script for malicious content.
So I leave it to you it implement these / or not. 
The plus-bug is probably not very critical 
however some incorrect quoted / unclose string can be a create some serve problem for a parser (you may apply later to the decompiled output)

### DIFF
--- a/libclamav/autoit.c
+++ b/libclamav/autoit.c
@@ -1416,7 +1416,7 @@ static int ea06(cli_ctx *ctx, const uint8_t *base, char *tmpd)
                         //TODO: Fix Autoit plus bug
                         // if (op == 0x49) /* + */ and next op ==0x05 /*int32*/ and that int32 is negative...
                         // skip next line (and don't add "+")
-                        // Background: "$a= -4" gets incorrect compiled. Decompiled it will be get "$A= + -4"
+                        // Background: "$a= (-4)" gets incorrect compiled. Decompiled it will be get "$A= (+ -4)"
                         // That doesn't effects the interpreter however when recompiling decompiled output that will result in a syntax error
 
                         UNP.cur_output += snprintf((char *)&buf[UNP.cur_output], 4, "%s ", opers[op - 0x40]);
--- a/libclamav/autoit.c
+++ b/libclamav/autoit.c
@@ -454,7 +454,7 @@ const char *autoit_functions[] = {
     "WINWAITNOTACTIVE"};
 
 const char *autoit_keywords[] = {
-    "UNKNOWN_0",
+    "UNKNOWN_0", // "".
     "AND",
     "OR",
     "NOT",
@@ -497,7 +497,7 @@ const char *autoit_keywords[] = {
     "FALSE",
     "DEFAULT",
     "NULL",
-    "UNKNOWN_43",
+    "VOLATILE",
     "ENUM",
 };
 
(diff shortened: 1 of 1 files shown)


recvfrom:
Thanks for submitting a PR with these fixes, @cw2k!

Regarding `#NoTrayIcon` showing up as `##NoTrayIcon`, I verified that the fix in this PR addresses the issue in a bunch of AutoIt EXEs I have (and has no obvious regressions).  00163b439172ff537183f616228e25a4b205df1544d572807b7d2f34ec5cd83e is a test case for this.

For `UNKNOWN_43` being `VOLATILE`, I couldn't find any samples that uses this... I even tried to make one, but all my attempts just resort to `VOLATILE` being stored as text instead of the `UNKNOWN_43` identifier (even at the highest and lowest compression levels). @cw2k do you have an executable that uses this that you could share?

Regarding the "mask double quote" comment, after looking around I wonder if ClamAV is using `"` where it should be using `'` sometimes... Scanning 01f605cfefdb9904c0595ba6d7afa2532951e69969c549e52e42f9c46c7fe09c yields the following with ClamAV:
```
RUN ( "explorer.exe" & " "" & $S [ 0x00000001 ] & $S [ 0x00000002 ] & $S [ 0x00000003 ] & """ 
```
but with another AutoIt extractor, the following is extracted:
```
RUN ( "explorer.exe" & ' "' & $S [ 1 ] & $S [ 2 ] & $S [ 3 ] & '"' )
```
We should investigate a bit more...

Regarding the `Fix Autoit plus bug` comment, I identified a sample that exhibits this behavior - 254f3ad0a3af39792d18f716713c37a5934c407a90fb9de6b2aa71f5053285b0.  ClamAV extracts:
```
SHELLEXECUTE ( @COMSPEC , CHRW ( 0x00000051 - 0x00000022 ) <truncated> & CHRW ( 0x00000025 + - 0x00000005 ) & "0" & <truncated> CHRW ( 0x00000086 - 0x00000012 ) , "" , "" , @SW_HIDE )
```
the other AutoIt extractor I tried outputs the same thing, so I wonder if this is just how the AutoIt compiler stores certain values:
```
SHELLEXECUTE ( @COMSPEC , CHRW ( 81 - 34 ) <truncated> & CHRW ( 37 + - 5 ) & "0" & <truncated> CHRW ( 134 - 18 ) , "" , "" , @SW_HIDE )
```

micahsnyder:
I have this PR staged to merge, but would like to give you credit in the acknowledgements section for the 0.104 release notes (NEWS.md file).  Is there a name we can use?  Should I put "cw2k"?

micahsnyder:
I just merged this, which you can find here: https://github.com/Cisco-Talos/clamav-devel/commit/a1e66c1a3760f581b8faa9a7501c4ef55bc3301d

If you get back to us Re: you name for the acknowledgements, I'd be happy to update it in our news document before the release. 

Thank you for the contribution to ClamAV!

cw2k:
> I have this PR staged to merge, but would like to give you credit in the acknowledgements section for the 0.104 release notes (NEWS.md file). Is there a name we can use? Should I put "cw2k"?

Yeah just cw2k.
That's fine. :)

PR #146, opened 2020-12-11
Deprecation of Pidfile Parameters

goshansp:
Dear Colleagues and ClamAV-Users
This proposed PR might break your deployment, as it aims for removing PidFile options completely from the project. It is my first PR and I might have missed things. 

The handling of PID-files should be taken care of by the init system / service manager. The Fedora/EPEL RPMS does apparently not rely on ClamAV-Managed PID-files, hence PID-files are completely removed in this proposed PR. OpenRC should also able to handle PID-files so there seems to be no need for ClamAV to handle any PID-Files.

This PR will do the opposite of the original proposal by @orlitzky :
[Bug 12595 - Add -p (--pidfile) flags for clamonacc, clamav-milter, and clamd](https://bugzilla.clamav.net/show_bug.cgi?id=12595)

This PR reduces the codebase by 155 lines. It will make handling for users easier, as the PidFile is gone from the conf and freshclam parameter.

This PR may break packaging. Please check if this PR integrates with your stack if if you need support doing so. I'll be happy to support you. My sincere thanks to @orlitzky and @micahsnyder for the motivating replies to my questions.

Best regards,
Hanspeter

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@ ClamAV 0.104.0 includes the following improvements and changes.
 
 ### Major changes
 
-- PidFile Option completely removed from codebase closing https://bugzilla.clamav.net/show_bug.cgi?id=12595
+- PidFile handling removed from codebase closing https://bugzilla.clamav.net/show_bug.cgi?id=12595
 
 ### Notable changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,8 @@ ClamAV 0.104.0 includes the following improvements and changes.
 
 ### Major changes
 
+- PidFile Option completely removed from codebase closing https://bugzilla.clamav.net/show_bug.cgi?id=12595
+
 ### Notable changes
 
 ### Other improvements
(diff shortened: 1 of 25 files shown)


goshansp:
after posting to the clamav-users mailing list i have not received any evidence of this PR breaking existing deployments as of yet. alternative/legacy service scripts may simply use something like follows to create a pid file while demonizing:
```
& echo $! > /var/run/program.pid
```
please provide any evidence for this PR to break your deployment and i'd be happy to look into it.

goshansp:
Shouldn't there be any priviledge-drop / fork at all? Isn't that something that the service manager should take care of as well? Can we deprecate the User-option also? I found the user option confusing, because i remember the user was mentioned as clamav in one place and clamscan in another in doc. I would like to investigate deprecating the user or can you explain to me why we would want a priviledge-drop/fork?

micahsnyder:
As per discussion on https://bugzilla.clamav.net/show_bug.cgi?id=12595, removing the PID file option and/or User option would break things for many users, particularly for systems with simpler service managers that lack PID and run-as-user features. 

Instead, we should add the `--pid=FILE` command line option to `clamd` and `clamonacc`, like `freshclam` has.  We can recommend the new `--pid` command line options over the PidFile config options in the manpages and in clamd.conf.sample.

PR #147, opened 2020-12-16
Fix wrong cmake target name for static libclamav ICONV link

bdb:


micahsnyder:
Hi @bdb thanks for the PR.  I actually already have a much larger update to the CMake build system in the works to add fuzzer support and a CTest test suite that included a fix for this:

https://github.com/micahsnyder/clamav-devel/blob/CLAM-34-unit-test-overhaul-windows-support/libclamav/CMakeLists.txt#L638
Fix in this commit: https://github.com/micahsnyder/clamav-devel/commit/8dd9de170bf1b0e02918a57b75c9478131ffaa88

My CMake update with this fix should merge next month some time after we finish swapping out autotools for CMake in our build & test pipelines and iron out any wrinkles.

bdb:
OK, I have may have more patches coming up (besides the 2 already) should I just not submit PRs for those?

e.g. libclamav and libfreshclam both cause build system failures when SHARED=OFF and STATIC=ON because their respective aliases are not defined for static lib when SHARED=OFF.

e.g here's a patch for libclamav
`--- CMakeLists.txt	2020-12-14 16:48:44.000000000 -0600
+++ CMakeLists.txt.tmp	2020-12-14 16:46:46.000000000 -0600
@@ -616,6 +616,9 @@
     endif()
 
     add_library( ClamAV::libclamav_static ALIAS clamav_static )
+    if(NOT ENABLE_SHARED_LIB)
+        add_library( ClamAV::libclamav ALIAS clamav_static )
+    endif()
 endif()`

bdb:
Ah, I see in your patch you caught the static lib only build issue. Cool. Thanks for the info.

micahsnyder:
> Ah, I see in your patch you caught the static lib only build issue. Cool. Thanks for the info.

Yup! The static-only build should work when my patch merges. Thanks for looking at the new CMake build. Feel free to reach out about and additional improvements 👍

PR #149, opened 2020-12-17
clamd: Export StreamMaxLength in the output of VERSIONCOMMANDS

markpizz:
An application that wants to use clamd's INSTREAM command can determine
the availability of the INSTREAM command by using the VERSIONCOMMANDS.
However to reliably use the INSTREAM it must also know the
StreamMaxLength parameter.


micahsnyder:
@markpizz Apologies for the delay in reviewing your work.  I've spent some time with it today and am not convinced that this is a good design decision.

With this PR, the VERSIONCOMMANDS response looks like this:
```bash
❯ printf 'zVERSIONCOMMANDS\0' | nc -U /tmp/clamd.socket
ClamAV 0.104.0-devel-20210620| COMMANDS: SCAN QUIT RELOAD PING CONTSCAN VERSIONCOMMANDS VERSION STREAM END SHUTDOWN MULTISCAN FILDES STATS IDSESSION INSTREAM StreamMaxLength=26214400 DETSTATSCLEAR DETSTATS ALLMATCHSCAN⏎
```

`StreamMaxLength=26214400` isn't a command, and doesn't belong in this list. 

I can understand that knowledge of StreamMaxLength is necessary if interacting with a ClamD server that is customized to allow streaming larger files. I'm open to other ideas for how to query the max stream length.

micahsnyder (commented):
If you want to propose a better way to query the max stream length I'd be happy to discuss it or review a new PR. I promise to respond much faster next time. 
I'm declining the PR for this implementation.

markpizz:
I see 3 choices:
1) What I proposed initially - you don't like that one
2) A completely different command that returned just the StreamMaxLength 
3) A variant on the initial proposal which returns ```COMMANDS: SCAN QUIT RELOAD PING CONTSCAN VERSIONCOMMANDS VERSION STREAM END SHUTDOWN MULTISCAN FILDES STATS IDSESSION INSTREAM INSTREAM-StreamMaxLength=26214400 DETSTATSCLEAR DETSTATS ALLMATCHSCAN``` Note the additional INSTREAM-StreamMaxLength=nnn

My initial proposal was based on the idea that a connecting clamd client would ignore things it didn't understand and only use things it specifically cared about.  Correct use of INSTREAM really needs to know the max length.  The client may not have any other access to clamd configuration details on the system serving these requests.  A client that attempts to send more than the clamd configured data size currently fails in unusual ways.  The client may be busy sending all the data it wants to send and not in a position to take a response until it finishes sending all that data.  The response may end up getting lost when clamd ultimately drops the connection since it was too long and the client may never see it, and this approach wastes bandwidth and cycles on both the client and server to handle it.

If you agree on a workable approach to conveying the StreamMaxLength info I'll be glad to submit a new PR against the latest codebase that implements it for future inclusion.

BTW. I was the one who originally proposed what was ultimately called INSTREAM.  That was way back in 2006.  My proposal was described in Bug 99 of the old bug system and the command I proposed was called CHUNKEDSTREAM.

micahsnyder:
You're probably right that most clients would ignore things they don't understand.  I just don't like putting something that isn't a command in the list of commands. 

Option 2 feels more correct to me. Maybe something like `INSTREAMMAXLENGTH`. An application could query `VERSIONCOMMANDS` first to find out if `INSTREAMMAXLENGTH` is available before calling it, but wouldn't need to. Querying `INSTREAMMAXLENGTH` without first checking if the command is available should also be fine. An older version would return `UNKNOWN COMMAND`. 

Ahh I see: https://bugzilla.clamav.net/show_bug.cgi?id=99 
Great to see that you're still working with ClamAV almost 15 years later!

PR #151, opened 2021-02-05
fix invalid unsigned arithmetic.

ihsinme:
I believe your checks are not correct.
in my opinion they are equivalent to checks `!=`.
I suggest a simple fix.


micahsnyder:
Thanks @ihsinme, I think you're right.  Nice find!

micahsnyder:
I merged this and https://github.com/Cisco-Talos/clamav-devel/pull/150 here: https://github.com/Cisco-Talos/clamav-devel/compare/f55863ae3352...56969286a121

Thanks again!

PR #157, opened 2021-05-03
fix logg with time in foreground

fallaha:
when we use the mprintf function and pass the time string as the first parameter, the mprintf function cannot recognize specific character (such as ! ^ * ,...) so does not replace it with the appropriate word (such as ERROR, WARNNING)
I separated the time string from the error message and print they separately with call twice mprintf.

### DIFF
--- a/common/output.c
+++ b/common/output.c
@@ -434,10 +434,10 @@ int logg(const char *str, ...)
                 cli_ctime(&currtime, timestr, sizeof(timestr));
                 /* cut trailing \n */
                 timestr[strlen(timestr) - 1] = '\0';
-                mprintf("%s -> %s", timestr, buff);
-            } else {
-                mprintf("%s", buff);
-            }
+                mprintf("%s -> ", timestr);
+            } 
+             mprintf("%s", buff);
+            
         }
     }
 


micahsnyder:
Hi @fallaha 

A quick review of your code feels correct, but it is causing test failures when I run `ctest` (or `ctest -V -R freshclam\$` for just the freshclam test).  

I see the failure in github actions and also if I test locally on my laptop:
image

It is printing:
```
Sun Jun 20 18:05:38 2021 -> Sun Jun 20 18:05:38 2021 -> Sun Jun 20 18:05:38 2021 -> This means that you have been rate limited by the CDN.
```
Instead of 
```
Sun Jun 20 18:05:38 2021 -> WARNING: FreshClam received error code 429 from the ClamAV Content Delivery Network (CDN).
Sun Jun 20 18:05:38 2021 -> This means that you have been rate limited by the CDN.
```

The code is:
```c
                    logg("^FreshClam received error code 429 from the ClamAV Content Delivery Network (CDN).\n");
                    logg("This means that you have been rate limited by the CDN.\n");
```

micahsnyder (changes requested):
Please find a solution that passes the tests.

fallaha:
> Please find a solution that passes the tests.

I have no idea :/

micahsnyder:
Yeah I don't either. 

I've been wanting to fix this issue for years, but it will take some fussing with it to find the right solution. 

To be honest, I've been extremely tempted to change the `logg()` function to take a `level` parameter as the first argument with a new enum and then replace all uses of `logg()` to use that instead of the stupid character prefix ( `~`/`!`/`^`/etc. ). The enum would be something simple like:
```c
enum log_level {
  LOG_DEBUG,
  LOG_INFO,
  LOG_WARN,
  LOG_ERROR
}
```

This seems like the logic would be waaay less fussy than trying to parse the log-level character from the log message. 

We could go further and also wrap `logg(level, msg, ...)` with functions or macros: `log_debug()`, `log_info()`, `log_warn()`, and `log_error()`.  That would make the function calls a little tidier.

fallaha:
This is a good way
Why don't you implement it?

micahsnyder:
@fallaha I'd like to, I (or someone else) simply need to find time to work on it.

PR #160, opened 2021-05-16
clamd/CMakeLists.txt: missing  clamav-daemon.socket

akuster:
clamav-daemon.service requires clamav-daemon.socket but
it is missing from the build and install process.

Signed-off-by: Armin Kuster <akuster808@gmail.com>


micahsnyder:
Thanks for the fix @akuster . It looks good.

I rebased and merged this here: https://github.com/Cisco-Talos/clamav/commit/9e013248a40ff0f435d3949fcb4e05e3e7e8bfaa

PR #161, opened 2021-05-18
FreshClam: fixed wrong variable in debug message

sherpya:
Simple fix, wrong variable is used (copy/paste error)


micahsnyder:
Nice find.  Thanks @sherpya,  Will merge shortly.

PR #164, opened 2021-05-29
Fix configuration error when ENABLE_EXTERNAL_MSPACK is ON

yasuhirokimura:
Fix configuration error when ENABLE_EXTERNAL_MSPACK is ON.

This fixes bug 12759.

### DIFF
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -68,7 +68,7 @@ if(ENABLE_APP)
             tomsfastmath
             bytecode_runtime
             JSONC::jsonc
-            ClamAV::libmspack
+            ${LIBMSPACK}
             OpenSSL::SSL
             OpenSSL::Crypto
             ZLIB::ZLIB
@@ -106,7 +106,7 @@ if(ENABLE_APP)
             tomsfastmath
             bytecode_runtime
             JSONC::jsonc
-            ClamAV::libmspack
+            ${LIBMSPACK}
             OpenSSL::SSL
             OpenSSL::Crypto
             ZLIB::ZLIB
@@ -156,7 +156,7 @@ target_link_libraries(check_clamav
         tomsfastmath
         bytecode_runtime
         JSONC::jsonc
-        ClamAV::libmspack
+        ${LIBMSPACK}
         OpenSSL::SSL
         OpenSSL::Crypto
         ZLIB::ZLIB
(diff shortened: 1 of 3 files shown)


micahsnyder:
Wow thanks for putting together a PR @yasuhirokimura 
I'm presently working on some tests for our internal suite to try building on various operating systems with an external libmspack. After seeing your email earlier this week I wanted to make sure we test for this and a couple other build configurations.

At a glance your work looks like it would fix the problem but it's a bit heavy handed with the copy-paste.
You should be able to call `target_link_libraries` more than once, so you can have an extra call to add libmspack for each target that needs it. I'll add some comments to try to demonstrate.

unit_tests/CMakeLists.txt
```diff
@@ -328,17 +393,31 @@ if(WIN32)
         # Have CMake invoke itself to performa a local install for our test suite.
         #
         if(ENABLE_APP)
-            add_custom_target(test_install
```
micahsnyder, line 331: Instead of this, we can also simplify by using `add_dependencies()` for the `test_install` target after we add the custom target. Eg.
```cmake
add_custom_target(test_install
    ALL
    "${CMAKE_COMMAND}"
    -D CMAKE_INSTALL_CONFIG_NAME:string=$<CONFIG>
    -D CMAKE_INSTALL_PREFIX:string=$<TARGET_FILE_DIR:check_clamav>
    -P "${CMAKE_BINARY_DIR}/cmake_install.cmake"
    DEPENDS
        check_clamav check_clamd check_fpu_endian
        ClamAV::libclamav ClamAV::libfreshclam ClamAV::libunrar ClamAV::libunrar_iface 
        clambc clamd clamdscan clamdtop clamscan clamsubmit clamconf freshclam-bin sigtool
)

if(ENABLE_EXTERNAL_MSPACK)
    add_dependencies(test_install MSPack::mspack)
else()
    add_dependencies(test_install ClamAV::libmspack)
endif()
```

micahsnyder (changes requested):

yasuhirokimura:
@micahsnyder Thanks for feedback. I pushed revised commit so please check it.

micahsnyder:
@yasuhirokimura your idea to use a variable instead is far better than what I recommended. Very nice.
I haven't yet finished the new tests to verify that external libmspack works on each OS so it may take a little bit longer while I fuss with that. 

Thanks so much for the help.

micahsnyder:
Everything worked great.  Thanks for the fix.  Here's the merged commit: https://github.com/Cisco-Talos/clamav/commit/fb479870cc8e343df5521712fd354bcea5b71f27

PR #165, opened 2021-06-01
Adds SECURITY.md policy file

micahsnyder:
We would like to switch from Bugzilla to Github Issues. This will make
issue reporting more accessible (more folks have a Github account than a
bugzilla.clamav.net account) in addition to the benefits of a more
modern issue tracker.

However, GitHub Issues reports are always public. Vulnerability reports
will have to go somewhere else. The preferred option is to report them
to Cisco PSIRT after which PSIRT will coordinate with the ClamAV team
and the reporter to resolve the issue.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,131 @@
+# ClamAV Security Policy
+
+## What constitutes a security issue / vulnerability?
+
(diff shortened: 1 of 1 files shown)


micahsnyder:
Updated with minor grammatical fixes.

micahsnyder:
> [psirt@cisco.com](mailto:psirt@cisco.com) is correct.

Hahaha I don't even know how I did that.  

> May want to add a sentence about "your issue will be promptly removed"

Sounds good to me.

I just force-pushed an update. I will run it by PSIRT next for a proof-read before I merge it in the internal repo.

PR #167, opened 2021-06-11
Clam 1481 clamsubmit cookie 0.103.3

micahsnyder:
ClamSubmit: Fix __cfduid cookie failure

    Cloudflare deprecated the __cfduid cookie which caused ClamSubmit
    failures on systems that stopped receiving the cookie.

    This commit removes support for the __cfduid cookie.
    Also made the session cookie optional, in case that disappears too.

    Changed error messages over to use the logg() function like our other apps.

    Tidied up some of the logic, and changed "cleanup" label to "done" to
    match other code.


micahsnyder:
Forgot to close this after we merged it in the old upstream bitbucket repo

PR #169, opened 2021-06-17
Fix bytecode match evaluation for PDF scans

micahsnyder:
Bytecode signatures targeting PDF files (target 10) fail to evaluate
match conditions. This occurs because the raw scan step during a magic
scan currently occurs _after_ the PDF scan, thus at the time the hook is
triggered, no matches have been performed and the logical sig eval is
always negative.

This commit fixes the issue by relocating the PDF parsing step to occur
after the raw scan, thereby enabling the bytecode lsig evaluation to
match and the signature to execute.


micahsnyder:
This PR was replicated from a PR in our internal backlog where it was pending additional research & testing to see if there isn't a better solution.

Prior conversation, for reference:

> Andrew: It's interesting that in that top section of scan calls there's one for ELF, and yet the ELF unpacker hook gets evaluated correctly... It might be worth looking at how that works to verify whether moving the scanpdf code to the lower section is the correct fix.
> 
> Micah: Perhaps it is picked up by the `ret = cli_unpackelf(ctx);` in the bottom section? I think I'd have to debug it to find out. Can you share a sig and sample to test with so I can find out?
> 
> Andrew: Here's my test ELF bytecode sig, which should match on any ELF file: 
>   < Snip. See: `clamav-coverage-test-cases/test-case-elf-unpacker.bc` in Andrew's `testing` repository on the internal gitlab. >

micahsnyder:
I'm pretty sure this is the best fix, at least for now. The reason the elf and macho parser can run before the raw scan and have the unpacker hook run _after_ the raw scan is that the elf parser doesn't provide metadata to the bytecode signatures. By contrast, the PE parser only runs after the raw scan, similar to the PDF parser now will -- both of which provide metadata to the bytecode signatures for evaluation. For the PDF parser, this happens with:
```c
    cli_bytecode_context_setpdf(bc_ctx, phase, pdf->nobjs, pdf->objs, &pdf->flags, pdf->size, pdf->startoff);
```
On idea if we really wanted to parse things before the raw-scan and then run hooks after, would be to start carrying file parsing metadata in the ctx in some sort of allocated opaque structure, to be evaluated when running the hook depending on the file type. I'm not really sure if there is much of an advantage in running the raw-scan after the parsers. I imagine we could try moving it up and doing some benchmarking to find out.  🤷‍♀️  A problem for another day. 

Anyways, TL;DR, I think this PR is good.

ragusaa (approved):

PR #170, opened 2021-06-17
Adds file inspection callback and example code

micahsnyder:
libclamav callbacks can be used to access embedded file content at each
layer of extraction during the course of a scan. The existing callbacks
only provide access to the file descriptor and a guess at the file type.

This patch adds a new callback for the purposes of file/archive
inspection that provides additional insight into the embedded file.
This includes:
- ancestors: an array of parent file names
- parent file size: the size of the direct parent layer
- file name: current layer's filename, if any.
- file buffer (pointer)
- file size: size of file buffer
- file type: just a guess at the current file's type
- file descriptor: may be -1 if the layer is in-memory only.
- layer attributes: a flag field. see LAYER_ATTRIBUTE_* defines in clamav.h

Two new example apps are added that are automatically built when
compiling under CMake:
- ex2 demonstrates the prescan callback.
- ex3 demonstrates the new file inspection callback.

The examples are now installed if enabled, so you can test them in the
Docker image, and so that they'll be colocated with the DLLs so you can
test them on Windows. The installed examples should also be able to find
the UnRAR library at run time, without having to set LD_LIBRARY_PATH.

This commit also sets the fmap->name in an fmap-scan using the basname
of the provided filename if the caller provided the filename and the
provided fmap does not have the name set.


micahsnyder:
This PR was replicated from an internal backlogged PR.  It is pending additional research & testing.  I'm not yet certain this new API offers the best parameters.  Will have to work with other teams to demonstrate its value.

micahsnyder:
I forgot to note that to test the example program ex3, you need to tell it where to load libclamunrar_iface for RAR support to work. Eg:
```bash
LD_LIBRARY_PATH=/home/micasnyd/workspace/clamav/build/libclamunrar_iface ./examples/ex3_file_inspection ~/Downloads/test-files.zip
```

micahsnyder:
I've updated this PR once more, making it so the "is normalized" and "was encrypted" booleans are provided in an attributes flag field that may be extended without breaking the API. 
I've also changed the "parents" parameter to an "ancestors" parameter that is now an array of filenames. the filenames may be NULL for layers that don't have names. 

I'm content with how this works now. Ready for review.

libclamav/matcher.c
```diff
@@ -802,6 +802,8 @@ cl_error_t cli_scan_desc(int desc, cli_ctx *ctx, cli_file_t ftype, uint8_t ftonl
     if (NULL != new_map) {
         funmap(new_map);
     }
+    // Clear the next-layer attributes so we don't accidentally apply them to subsequent layers.
+    ctx->next_layer_attributes = 0;
```
shutton, line 806: Per discussion, it would be safer to convert this to a parameter and pass around as an argument (possibly burying it in a parameter struct if the argument list gets onerous).

libclamav/scanners.c
```diff
@@ -134,9 +134,9 @@ cl_error_t cli_magic_scan_dir(const char *dir, cli_ctx *ctx)
     DIR *dd           = NULL;
     struct dirent *dent;
     STATBUF statbuf;
-    char *fname                      = NULL;
-    unsigned int viruses_found       = 0;
-    bool processing_normalized_files = ctx->next_layer_is_normalized;
+    char *fname                    = NULL;
+    unsigned int viruses_found     = 0;
+    uint32_t next_layer_attributes = ctx->next_layer_attributes;
```
shutton, line 139: Alignment not consistent.

libclamav/scanners.c
```diff
@@ -5629,6 +5748,9 @@ cl_error_t cli_magic_scan_file(const char *filename, cli_ctx *ctx, const char *o
     if (fd >= 0) {
         close(fd);
     }
+    // Clear the next-layer attributes so we don't accidentally apply them to subsequent layers.
+    ctx->next_layer_attributes = 0;
```
shutton, line 5752: I noticed you defined a constant for this (`LAYER_ATTRIBUTES_NORMAL`) -- might as well use it, or name it so it better reflects the cleared/initial state (e.g., `LAYER_ATTRIBUTES_NONE`).

shutton (changes requested):
Not familiar enough to follow the logic, so mainly commenting on style.

Also need to consider the risk level of this feature for 0.105. *Presumably* regression testing should vet it, but...

libclamav/scanners.c
micahsnyder: Good call, I like NONE

micahsnyder:
Just rebased to update it.

shutton (approved):
Ship it!

micahsnyder:
Test results look good.  Merging.

This new callback API will likely be annotated as "unstable" in the upcoming feature release in case we are required to adjust the parameters in the next feature release.

PR #171, opened 2021-06-17
Clam 1333 update unrar 6.0.6 and max dictsize

micahsnyder:
- Update to UnRAR v6.0.6

- Patch UnRAR: allow skipping files in solid archives

    This is a cherry-pick of commit 24f225c21ffa77dbd4b2a3c0377b9b3f86547ce6

    Modification to unrar codebase allowing skipping of files within
    Solid archives when parsing in extraction mode, enabling us to skip
    encrypted files while still scanning metadata and potentially
    scanning unencrypted files later in the archive.

- RAR: limit archive extraction by dictsize

    In addition to unpacked size, RAR archive file entries also have a
    dictionary size. This commit checks the dictionary size when peeking at
    the next file entry so that it can be skipped early if the dictionary
    size is larger than our max filesize, like we do with the unpacked size.


micahsnyder:
Replacing this with https://github.com/Cisco-Talos/clamav/pull/210

PR #172, opened 2021-06-21
libclamav: Increase max PE section count to 65535

recvfrom:
Windows XP had a maximum section count of 96, and this has been
the max for ClamAV forever as well. Raising this prevents malicious
executables from being able to evade certain ClamAV signatures by
having 97 or more sections.


micahsnyder:
I didn't observe any suspicious results in a 4 million sample regression test.  I'm happy with the test results, will merge.

PR #173, opened 2021-06-22
docker: Update docker registry URL

oliv3r:
The URL registry.hub.docker.com was apparently deprecated for a while,
and started to give 404 errors as of today for some repo's. The correct
URL is index.docker.io, so lets use that instead.


micahsnyder:
Woah! Thanks for the fix. I just merge this upstream (which I hope is the last time I have to tell someone that before we start merging directly here).  Here is the commit: https://github.com/Cisco-Talos/clamav/commit/b5e4d953738ad8c804f297653369ba2b458d126e

PR #174, opened 2021-06-22
Clam 1493 Windows services

kang-grace:
Adds Windows service options for freshclam and clamd.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,6 +67,11 @@ ClamAV 0.104.0 includes the following improvements and changes.
   Special thanks to Olliver Schinagl for his excellent work creating ClamAV's
   new Docker files, image database deployment tooling, and user documentation.
 
+- `clamd` and `freshclam` are now available as Windows services. To install
+  and run them, use the `--install-service` option and `net start [name]` command. 
+  
+  Special thanks to Gianluigi Tiesi for his original work on this feature.
+
 ### Notable changes
 
 The following was added in 0.103.1 and is repeated here for awareness, as
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -75,6 +75,10 @@
 #include "shared.h"
 #include "scanner.h"
 
+#ifdef _WIN32
+#include "service.h"
+#endif
+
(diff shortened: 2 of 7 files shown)


unit_tests/freshclam_test.py
```diff
@@ -263,8 +263,8 @@ def test_freshclam_05_cdiff_update(self):
         # start with this CVD
         shutil.copy(TC.path_source / 'unit_tests' / 'test_db' / 'test-1.cvd', TC.path_db / 'test.cvd')
 
-        # update to this CVD
```
micahsnyder, line 266: It appears that your commit somehow includes bits of a previous commit, I think as a result of a merge.  Can I trouble you to remove the changes to freshclam_test.py?

micahsnyder:
@kang-grace we should add a description of this new feature in the `NEWS.md` file for the 0.104 release, and credit @sherpya for his work in the news burb and also in the Acknowlegdements section.

@sherpya if you have any concerns about how we're importing your work, your review will be appreciated.

unit_tests/freshclam_test.py
kang-grace: Oops - should be fixed now.

micahsnyder:
The Jenkins pipeline looks great.  *Everything* passed. 

Here's the main clamd service test output:
image

And the main freshclam service test output:
image

sherpya:
> @kang-grace we should add a description of this new feature in the `NEWS.md` file for the 0.104 release, and credit @sherpya for his work in the news burb and also in the Acknowlegdements section.
> 
> @sherpya if you have any concerns about how we're importing your work, your review will be appreciated.

looks ok, a see a new flag "service-mode", I was always calling svc_stuff, they seems harmless, anyway better to call only if in service mode

micahsnyder:
Everything look good to me. 

Still seeing some odd failures on windows with:
`Wrong reply for command zMULTISCAN \nonexistentfilename.`
`Wrong reply for command SCAN \nonexistentfilename`
But these aren't your fault. 

I'll squash & merge it now. 
Nice work!

PR #181, opened 2021-06-30
CLAM-1404 CLAM-1508 ClamDScan leaks & ExcludePath issues

micahsnyder:
This PR includes a feature test that should suffice, but it's pretty trivial to reproduce the issue manually as well with valgrind and the test files. 

### ClamDScan: --fdpass/--stream leak; ExcludePath issues

ClamDScan will leak the memory for the scan target filename if using `--fdpass` or using `--stream`. This commit fixes that leak.
Resolves: https://bugzilla.clamav.net/show_bug.cgi?id=12648

ClamDScan will fail to scan any file after running into an "ExcludePath" exclusion when using `--fdpass` or `--stream` AND `--multiscan` (`-m`). The issue is because the `parallel_callback()`
callback function used by file tree walk (FTW) feature returns an error code for excluded files rather than "success". Memory for the accidentally-excluded paths for a given directory also appears to be leaked. This commit resolves this accidental-abort issue and the memory leak.

There was an additional single file path memory leak when using `--fdpass` caused by bad error handling in `cli_ftw()`. This was fixed by removing the confusing ternaries, and using separate pointers for each filename copy.

ClamDScan with ExcludePath regex may fail to exclude absolute paths when performing relative scans because the exclude-check function may match using provided relative path (E.g. `/some/path/../another/path`) rather than an absolute path (E.g. `/some/path/another/path`). This issue is resolved by getting the real path at the start of the scan, eliminating `.` and `..` relative pathing from all filepaths.

TODO 1: In addition to being recursive (bad for stack safety), the File Tree Walk (FTW) implementation is a spaghetti code and should be refactored.

TODO 2: ExcludePath will print out "Excluded" for each path that is excluded when using `--fdpass` or `--stream`, and for each path directly scanned that is directly excluded. But in a recursive regular-scan, the "Excluded" message for the those paths is missing.

### Test: ClamDScan ExcludePath; Valgrind

Adds a basic test to validate that ExcludePath correctly excludes a subdirectory but does not exclude subsequent files. As with the other ClamD/Scan tests, it will test in each mode: regular, stream, and fdpass (if available).

Unlike the other tests, this one tests ClamDScan with Valgrind instead of ClamD.

Refactored the clamd_test.py file to reduce duplicate code, and support enabling and disabling valgrind when running ClamDScan and ClamD.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -31,6 +31,9 @@ jobs:
         with:
           args: install wixtoolset
 
+      - name: Install pytest for easier to read test results
+        run: python3 -m pip install pytest
+
       - uses: lukka/get-cmake@latest
 
       # Restore from cache the previously built ports. If cache-miss, download, build vcpkg ports.
@@ -91,6 +94,9 @@ jobs:
       - name: Install Dependencies
         run: brew install bzip2 check curl-openssl json-c libxml2 ncurses openssl@1.1 pcre2 zlib
 
+      - name: Install pytest for easier to read test results
+        run: python3 -m pip install pytest
+
       - uses: lukka/get-cmake@latest
 
       - name: Create Build Directory
@@ -138,6 +144,9 @@ jobs:
 
       - name: Install Dependencies
         run: sudo apt-get install -y check libbz2-dev libcurl4-openssl-dev libjson-c-dev libmilter-dev libncurses5-dev libpcre3-dev libssl-dev libxml2-dev zlib1g-dev
+
+      - name: Install pytest for easier to read test results
+        run: python3 -m pip install pytest
 
       - uses: lukka/get-cmake@latest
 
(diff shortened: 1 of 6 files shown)


ragusaa (approved):

micahsnyder:
Just a quick fix for the doxygen doc comments. Nearly introduced more incorrect `[out]` annotations! 😅

```
❯ git diff
diff --git a/libclamav/others_common.c b/libclamav/others_common.c
index e253200f4..3ed2a814d 100644
--- a/libclamav/others_common.c
+++ b/libclamav/others_common.c
@@ -581,9 +581,9 @@ static int get_filetype(const char *fname, int flags, int need_stat,
  *
  * @param fname         The file path
  * @param flags         CLI_FTW_* bitflag field
- * @param statbuf       [out] the stat metadata for the file.
- * @param stated        [out] 1 if statbuf contains stat info, 0 if not. -1 if there was a stat error.
- * @param ft            [out] will indicate if the file was skipped based on the file type.
+ * @param[out] statbuf  the stat metadata for the file.
+ * @param[out] stated   1 if statbuf contains stat info, 0 if not. -1 if there was a stat error.
+ * @param[out] ft       will indicate if the file was skipped based on the file type.
  * @param callback      the callback (E.g. function that may scan the file)
  * @param data          callback data
  * @return cl_error_t
```

PR #182, opened 2021-06-30
Changed default config dir to <prefix>/etc/clamav

micahsnyder:
This resolves https://bugzilla.clamav.net/show_bug.cgi?id=12718

The default config directory on Linux/Unix systems will now be
   ${CMAKE_INSTALL_PREFIX}/etc/clamav
where it used to be:
   ${CMAKE_INSTALL_PREFIX}/etc

Note that most distributions will customize it to be:
   /var/lib/clamav

### DIFF
--- a/CMakeOptions.cmake
+++ b/CMakeOptions.cmake
@@ -9,7 +9,7 @@ if(WIN32)
         "Database directory.")
 else()
     set(APP_CONFIG_DIRECTORY
-        "etc" CACHE STRING
+        "etc/clamav" CACHE STRING
         "App Config directory.")
     set(DATABASE_DIRECTORY
         "share/clamav" CACHE STRING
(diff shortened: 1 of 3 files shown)


ragusaa (approved):

micahsnyder:
This PR will requires changes to the jenkins test-framework (see PR) to use `{TEST_PATH}/etc/clamav` instead of `{TEST_PATH/etc}`. I ran into a some failures with that change on macOS because the command to reinstall the test-framework wasn't working properly.  After changing it to force-reinstall, it seems to be working nicely. Since every PR after this will fail in Jenkins with the test-framework change without a rebase, I'm going to save it for last before the RC.

micahsnyder:
I really wanted to do this but after extensive thought about this and discussing it with the mailing list, it feels like it will add extra confusion for many users.  If you want to add a subdirectory in your clamav build for the config files, it's easy enough.  Have a look at https://docs.clamav.net/manual/Installing/Installing-from-source-Unix.html#a-linux-distribution-style-build to see how.

PR #187, opened 2021-07-03
Clam 1248 extract images from xls documents

micahsnyder:
To test this PR, find or create an XLS spreadsheet with images inserted into the cells.  Run with `--gen-json` and you should see the image detected. Run with `--gen-json --leave-temps` and maybe `--tempdir=/some/path`.  Inspect the temp files and you should find the image files extracted in there.  The `metadata.json` file with --leave-temps enabled will have a path in it to make it easier to find the files.  The files should be the same as when added to the document, or when extracted using oledump from https://github.com/InQuest/DidierStevensSuite/blob/BIFF-Image-Dump-Switch/plugin_biff.py#L570-L592 as described in https://inquest.net/blog/2019/01/29/Carving-Sneaky-XLM-Files E.g. usage: `./oledump.py -p plugin_biff /path/to/file --pluginoptions "-e"`

---

OLE2 / XLS document image extraction

Added a feature to extract images from OLE2 BIFF streams.
This work was derived from InQuests blog post about extracting XLM and images from XLS files:
https://inquest.net/blog/2019/01/29/Carving-Sneaky-XLM-Files

Assorted ole2 parser code cleanup and massive error handling cleanup.

Also fixed the following:

- The XLS parser may fail to process all BIFF records if some of the records contain unexpected data or is otherwise malformed. Because the record size is already known, we can skip over the "malformed" record and continue with the rest.

- Fixed an issue where the ole2 header size was improperly calculated, failing to account for the new "has_xlm" boolean added for context.

- If XLM (and now images) are found when parsing an ole2 files the following other embedded content may not be processed:
  - document summary metadata
  - embedded ole10 files
  - ole2 temp subdirectories (i.e. recursion)

  The logic to process the above ole2 extracted temp files was present in the function which processes extracted VBA. When we added support for extracting XLM macros, processing these other data was lost. Really, the above need to be processed if any temp files were saved.

  I fixed this by restructuring the features to extract any type of temp file into separate functions per type of temp file. I then wrappped those in an ole2 temp dir scanning function. OLE2 temp directory scanning is recursive if there are subdirectories.


ragusaa:
I am noticing that filenames are CL_TYPE_ANY.<something> when the FileType is known to be CL_TYPE_PNG.  Is this a mistake or intentional?

ragusaa (approved):

micahsnyder:
> 
> 
> I am noticing that filenames are CL_TYPE_ANY. when the FileType is known to be CL_TYPE_PNG. Is this a mistake or intentional?

So this is kind of a mistake.  I switched it back to CL_TYPE_ANY for the embedded file scan so that I wasn't forcing the type detection in case extraction went poorly or the file was somehow not actually a PNG/JPEG/TIFF file.  I figure will leave the type detection up to clamav.  But as for the file name, it couldn't hurt to set something more descriptive based on what the document claims it is.  I'll find a fix.

micahsnyder:
@ragusaa I just force pushed a change to fix the filenames so they now say `PNG.<shorthash>` or `TIFF.<shorthash>` or whatever instead of `CL_TYPE_ANY.<shorthash>`. I also included a commit to fixup the doxygen params I messed up, plus some `[in/out]` -> `[in,out]` params throughout the codebase that I missed when fixing `[out]` param annotations for the progress-bar PR.

Lemme know if you come up with any great test files that we can include in a ClamScan feature-test or two to run with `ctest`. Other than maybe adding a test or two, this PR should be ready to merge.

micahsnyder:
The lgtm complaints are for an html/js scan-test file, not clamav code.

micahsnyder:
For anyone who is curious, my new commits tidied up the test files, unit_tests directory.  I wanted to add a test for this feature and wasn't sure where to put the files.  I may have gotten a little carried away.  

I also replaced the script which assembles our test files from split files with a script that decrypts (or encrypts) test files with a basic XOR.  We've found that some AV's alert on the split test files because the look like malformed PE files (which I guess they are).  XOR'ing with a long-ish key should prevent AV's from alerting on them or on our source tarball.  

The series of force commits after that were because I A) didn't notice that the daily.pdb test file was `.gitignore`'d and so was missing from the commit, and B) I did all my local testing with Ninja, which apparently creates build subdirectories _before_ the xor_testfiles.py script runs. This isn't true for Make, so the script was failing to create the test files.  It took me way longer than it should've to realize what was going wrong. 😓 

I also rebased the PR and found we'd added 1 other test that had to be adjusted for the new unit_test directory layout.

micahsnyder:
The internal jenkins piplines ran through, and look good. Going to merge this.

PR #194, opened 2021-07-07
libclamav, clamscan: load/unload callbacks & progress meters

micahsnyder:
Add progress callbacks to libclamav for:
- database load
- engine compile
- engine free

Add a progress bar to clamscan for load & compile.
These are disabled if you run with --debug.

Added code so you can test the engine-free callback by building with
ENABLE_ENGINE_FREE_PROGRESSBAR defined.

The compile & free progress callbacks pre-calculate the number of
tasks to complete to estimate the progress. Some tasks may take longer
than others so the progress speed my appear to vary a little.

### DIFF
--- a/clamscan/manager.c
+++ b/clamscan/manager.c
@@ -47,7 +47,7 @@
 #include <sys/types.h>
 #include <signal.h>
 #include <errno.h>
-#include <target.h>
+#include <math.h>
 
 // libclamav
 #include "clamav.h"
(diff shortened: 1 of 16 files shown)


libclamav/matcher-ac.c
```diff
@@ -2915,7 +2915,7 @@ cl_error_t cli_ac_addsig(struct cli_matcher *root, const char *virname, const ch
         return ret;
     }
 
-    if (new->offdata[0] != CLI_OFF_ANY && new->offdata[0] != CLI_OFF_ABSOLUTE && new->offdata[0] != CLI_OFF_MACRO) {
+    if (new->offdata[0] != CLI_OFF_ANY &&new->offdata[0] != CLI_OFF_ABSOLUTE &&new->offdata[0] != CLI_OFF_MACRO) {
```
ragusaa, line 2918: Is this spacing a result of clang-format?  I think this makes it harder to read.

micahsnyder: Huh, I hadn't noticed that.  I agree it makes it harder to read, and I have no idea why clang-format would remove those. Will investigate.

micahsnyder: It was definitely clang-format weirdness.   I haven't seen that one before.  I added parenthesis and new lines and now i think it's even easier to read and clang-format doesn't mess with it.

libclamav/clamav.h
```diff
@@ -601,6 +601,63 @@ enum cl_msg {
...
+ */
+extern void cl_engine_set_clcb_sigload_progress(struct cl_engine *engine, clcb_progress callback, void *context);
+
+/**
+ * @brief Set a progress callback function to be called incrementally during an
+ * engine compile.
+ *
+ * Disclaimer: the number of items for this is a rough estimate of the items that
+ * tend to take longest to compile and doesn't represent an accurate number of
+ * things compiled.
```
craig65535, line 635: So both `total_items` and `now_completed` will be a count of the slower items? (And is this true for _free_progress below as well?)

libclamav/clamav.h
```diff
@@ -940,7 +997,23 @@ extern cl_error_t cl_scanfile_callback(const char *filename, const char **virnam
 /* ----------------------------------------------------------------------------
  * Database handling.
  */
+
+/**
+ * @brief Load the signature databases found at the path.
+ *
+ * @param path      May be a file or directory.
+ * @param engine    The engine to load the signatures into
+ * @param signo     [out] The number of signatures loaded
```
craig65535, line 1006: This is a nitpick, but I think Doxygen uses `@param[out]`

clamscan/manager.c
```diff
@@ -631,33 +631,325 @@ static int scanstdin(const struct cl_engine *engine, struct cl_scan_options *opt
...
+    if (seconds >= 3600) {
+        fprintf(stdout, "%2lldh %02lldm", (long long)seconds / 3600, ((long long)seconds % 3600) / 60);
+    } else if (seconds >= 60) {
+        fprintf(stdout, "%2lldm %02llds", (long long)seconds / 60, (long long)seconds % 60);
+    } else {
+        fprintf(stdout, "%3llds", (long long)seconds);
+    }
+}
+
+static void print_no_sigs(size_t sigs, int bPad)
```
craig65535, line 663: `print_num_sigs`? Nitpick, but it reads like `no`

clamscan/manager.c
```diff
@@ -631,33 +631,325 @@ static int scanstdin(const struct cl_engine *engine, struct cl_scan_options *opt
...
+static void engine_free_callback(size_t total_items, size_t now_completed, void *context)
+{
+    time_t curtime = 0;
+
+    struct engine_free_progress *engineFreeProgress = (struct engine_free_progress *)context;
+
+    uint32_t i                = 0;
+    uint32_t totalNumDots     = 10;
+    uint32_t numDots          = 0;
+    double fractiondownloaded = 0.0;
```
craig65535, line 864: `fractiondownunloaded`? (Maybe I misunderstand the `down` part of this though)

libclamav/readdb.c
```diff
@@ -4520,6 +4541,126 @@ cli_insertdbtoll(struct db_ll_entry **head, struct db_ll_entry *entry)
...
+    }
+
+    sig_count++;
+    while (0 == feof(fp)) {
+        /* Get the next character */
+        current_character = fgetc(fp);
+
+        if (!in_sig) {
+            /* Not inside of a signature, yet */
+            if (!isspace(current_character) && // Ignore new lines and other forms of white space before a signature
```
craig65535, line 4571: "new lines" == "newlines"?

libclamav/readdb.c
```diff
@@ -4520,6 +4541,126 @@ cli_insertdbtoll(struct db_ll_entry **head, struct db_ll_entry *entry)
...
+ * Ignores lines starting with # comment character
+ *
+ * @param filepath
+ * @return size_t
+ */
+static size_t count_line_based_signatures(const char *filepath)
+{
+    FILE *fp              = NULL;
+    int current_character = 0;
+    int sig_count         = 0;
```
craig65535, line 4556: Should be a `size_t`

libclamav/readdb.c
```diff
@@ -4597,10 +4740,13 @@ static cl_error_t cli_loaddbdir(const char *dirname, struct cl_engine *engine, u
             if (0 == access(dbfile, R_OK)) {
                 daily_cld = cl_cvdhead(dbfile);
                 if (!daily_cld) {
-                    cli_errmsg("cli_loaddbdir(): error parsing header of %s\n", dbfile);
+                    cli_errmsg("cli_loaddbdir: error parsing header of %s\n", dbfile);
                     ret = CL_EMALFDB;
                     goto done;
                 }
+
+                /* Successfully opened the daily CLD file and ready the header info. */
```
craig65535, line 4748: "ready" = "read"?

ragusaa (approved):
LGTM

libclamav/readdb.c
micahsnyder: Haha yes I meant `\n` / `\r\n`

clamscan/manager.c
micahsnyder: That's a copy-paste typo.  I copied the implementation from the freshclam progress bar, which had used a curl callback to track download and upload progress (not that we used the upload one).  I should rename this variable.

clamscan/manager.c
micahsnyder: Yeah decent nitpick.  Will switch to `num`

libclamav/clamav.h
micahsnyder: I always forget 😅  I probably make that mistake again just last night working on the OLE2 image extraction PR.  Will revisit.

libclamav/clamav.h
micahsnyder: Yes, the slower tasks, or just 1 task for a batch of frees. 

I actually haven't worked it into the bytecode signature compile feature though because that code is more complicated and would require adding parameters to a few functions. So if using LLVM then the last task is "compile all the bytecode sigs" and will take a hot second before the overall progress is complete.  

For what it's worth, using LLVM for the runtime in 0.104 isn't likely -- since upgrading to support a modern version of libLLVM missed the cut and will have to go into 0.105.  I would like to try to find a way to have the compile progress track each bytecode signature that we compile someday though.

libclamav/clamav.h
```diff
@@ -601,6 +601,63 @@ enum cl_msg {
     CL_MSG_ERROR        = 128 /* LibClamAV ERROR: */
 };
 
+/**
+ * @brief Progress callback for sig-load, engine-compile, and engine-free.
+ *
+ * Disclaimer: total_items may increase with each call as new items to be
+ * processed are found. As such, total_items == now_completed does not mean it
+ * guaranteed to be complete.
```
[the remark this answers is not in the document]
micahsnyder: I should remove this disclaimer. I found away around this concern to pre-calculate all the compile tasks and all the free tasks. I also can add 1 task at the beginning and remove 1 task at the end if necessary to ensure that `total_items == now_completed` means we really are done... so yeah this disclaimer is now a lie, and needs to be removed.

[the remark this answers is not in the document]
micahsnyder: @craig65535 Are you okay with the progress callback not returning anything?  Would you want to be able to return something to cancel the task (or allow it to continue)?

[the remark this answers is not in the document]
craig65535: @micahsnyder As discussed, I think it would be good to future-proof this and have the callback return a `cl_error_t` (which would be ignored for now).

craig65535 (approved):

libclamav/clamav.h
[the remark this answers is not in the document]
micahsnyder: Ok just added a commit that adds a reserved return for the callback.

craig65535 (approved):

PR #202, opened 2021-07-15
Clam 1494 process memory scanning

kang-grace:
Adds process memory scanning in clamscan and clamdscan via the --memory option.
Can also kill and unload infected modules via --kill and --unload.


common/optparser.c
```diff
@@ -238,6 +238,12 @@ const struct clam_option __clam_options[] = {
     {NULL, "tgz", 0, CLOPT_TYPE_STRING, NULL, -1, "foo", 0, OPT_CLAMSCAN | OPT_DEPRECATED, "", ""},
     {NULL, "deb", 0, CLOPT_TYPE_STRING, NULL, -1, "foo", 0, OPT_CLAMSCAN | OPT_DEPRECATED, "", ""},
 
+#ifdef _WIN32
+    {NULL, "memory", 0, CLOPT_TYPE_BOOL, MATCH_BOOL, 0, NULL, 0, OPT_CLAMSCAN | OPT_CLAMDSCAN, "", ""},
+    {NULL, "kill", 'k', CLOPT_TYPE_BOOL, MATCH_BOOL, 0, NULL, 0, OPT_CLAMSCAN | OPT_CLAMDSCAN, "", ""},
+    {NULL, "unload", 'u', CLOPT_TYPE_BOOL, MATCH_BOOL, 0, NULL, 0, OPT_CLAMSCAN | OPT_CLAMDSCAN, "", ""},
```
micahsnyder, line 244: it looks like `'u'` is already being used by:
```c
    {NULL, "unpack", 'u', CLOPT_TYPE_STRING, NULL, -1, NULL, 0, OPT_SIGTOOL, "", ""},
```
I also have mixed feelings about reserving precious 1-letter shorthands for OS-specific features.  I think we should change this and the one above (kill) to `0` to remove the shorthand. notation.  Please also alter the `--help` string in `clamscan.c` to match:
```c
    mprintf("    --kill                               Kill/Unload infected loaded modules\n");
    mprintf("    --unload                             Unload infected modules from processes\n");
```

common/scanmem.c
```diff
@@ -0,0 +1,1005 @@
...
+#include "output.h"
+#include "exescanner.h"
+#include "scanmem.h"
+#include "clamdcom.h"
+
+typedef int (*proc_callback)(PROCESSENTRY32 ProcStruct, MODULEENTRY32 me32, void *data, struct mem_info *info);
+int sock;
+
+/* clamd helpers */
+static int chkpath(const char *path, struct mem_info *minfo)
```
micahsnyder, line 41: I'm curious if we can avoid having duplicate code by moving merging this and the one found in `clamdscan/proto.c`.  You will have to add at least one extra parameters to `chkpath()` to pass in `clamdopts`, and then change every use of `chkpath()` to pass that in.  I suppose the merged function will have to go somewhere under `common` but I'm not sure where. I suppose `misc.c`/`h` may be best - unless you have any better ideas. 🤷‍♀️

common/scanmem.c
```diff
@@ -0,0 +1,1005 @@
...
+            }
+            opt = opt->nextarg;
+        }
+    }
+    return 0;
+}
+
+/* Issues an INSTREAM command to clamd and streams the given file
+ * Returns >0 on success, 0 soft fail, -1 hard fail */
+static int send_stream(int sockd, const char *filename, struct mem_info *info)
```
micahsnyder, line 69: I just realized this function suffers the same copypaste issue and is nearly identical to a `send_stream()` function in `proto.c`.

Below, `connect_clamd()` also suffers the same problem.  It is mostly duplicate of [the poorly named] `dconnect()` function from `proto.c`. 

Do you think it makes sense to merge these all and put them inside of `clamdcom.c`/`.h`?

micahsnyder:
Looks like there is a build failure on Alpine:
```
/usr/lib/gcc/x86_64-alpine-linux-musl/10.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: ../common/libcommon.a(clamdcom.c.o):/tmp/testPath-G8EWUCNUM/clamav-devel/common/clamdcom.c:50: multiple definition of `nixsock'; CMakeFiles/clamdscan.dir/client.c.o:/tmp/testPath-G8EWUCNUM/clamav-devel/clamdscan/client.c:77: first defined here

collect2: error: ld returned 1 exit status
make[2]: *** [clamdscan/CMakeFiles/clamdscan.dir/build.make:145: clamdscan/clamdscan] Error 1
make[1]: *** [CMakeFiles/Makefile2:2005: clamdscan/CMakeFiles/clamdscan.dir/all] Error 2
make: *** [Makefile:182: all] Error 2
```

I'm not really sure why only Alpine is complaining about it, but there is an issue with it being defined more than once.

What should fix it is to define `struct sockaddr_un nixsock;` only in `clamdcom.c`, with an `extern`'ed declaration in `clamdcom.h`.  Then remove the declarations for `nixsock` from the other files, and make sure `clamdscan/client.c` includes `clamdcom.h`.

micahsnyder:
There were also some tests on fedora failing, but I don't expect that it's related. I suspect those will pass the next time it runs.

micahsnyder:
If I'm being really fussy, a couple of the new files are also missing new lines at the end.  

But otherwise everything seems pretty good. 👍  I tested the server program + logical signature with clamd+clamdscan and clamscan, using Administrative permissions and once with clamscan without. 

If you have time,after working on the user documentation for this, it'd be really cool if we could add a test for this in the regular tests suite (`test` repo in our gitlab) using that server program with the logical signature.

micahsnyder (changes requested):

PR #210, opened 2021-07-19
Clam 1333 update unrar 6.0.7 and max dictsize

micahsnyder:
This PR replaces https://github.com/Cisco-Talos/clamav/pull/171


libclamunrar/dll.cpp
```diff
@@ -230,7 +230,7 @@ int PASCAL RARReadHeaderEx(HANDLE hArcData,struct RARHeaderDataEx *D)
       // open callback for RAR5 archives and if password is invalid.
       if (Data->Arc.FailedHeaderDecryption)
         return ERAR_BAD_PASSWORD;
-
+
```
ragusaa, line 233: Seeing multiple places with spaces at the end of lines, did clang-format leave that in?

micahsnyder: libclamunrar isn't our code, so we don't reformat it, and my text editor hasn't had a chance to their extra whitespace.

ragusaa (approved):

PR #212, opened 2021-07-19
Callback pointer check added

strehle:
Crash in programs without callbacks
Found this regression after updating,
Caused by https://github.com/Cisco-Talos/clamav/commit/861153a656bcb5266952630f0a2aaed228883404


micahsnyder:
Thanks for the fix @strehle !

micahsnyder:
I just made a copy of this PR for 0.104 so we can include it in the upcoming feature version, release candidate.

PR #221, opened 2021-07-23
Clam 1524 fix coverity issues

ragusaa:
Addresses Coverity issues 348943, 349629, 351830, 352408, 352409, 352410, 352411, and 352427.

### DIFF
--- a/clamdscan/proto.c
+++ b/clamdscan/proto.c
@@ -240,7 +240,6 @@ static int chkpath(const char *path)
 {
     int status = 0;
     const struct optstruct *opt;
-    char *real_path = NULL;
 
     if (!path) {
         status = 1;
@@ -260,9 +259,6 @@ static int chkpath(const char *path)
     }
 
 done:
-    if (NULL != real_path) {
-        free(real_path);
-    }
     return status;
 }
 
--- a/libclamav/matcher-byte-comp.c
+++ b/libclamav/matcher-byte-comp.c
@@ -469,7 +469,7 @@ cl_error_t cli_bcomp_scanbuf(const unsigned char *buffer, size_t buffer_length, 
 
     uint32_t evalcnt = 0;
     uint64_t evalids = 0;
-    char *subsigid   = NULL;
+    char subsigid[3];
 
     if (!(root) || !(root->bcomp_metas) || !(root->bcomp_metatable) || !(mdata) || !(mdata->offmatrix) || !(ctx)) {
         return CL_SUCCESS;
@@ -484,16 +484,10 @@ cl_error_t cli_bcomp_scanbuf(const unsigned char *buffer, size_t buffer_length, 
         /* check to see if we are being run in sigtool or not */
         if (bcomp->lsigid[0]) {
 
-            subsigid = cli_calloc(3, sizeof(char));
             snprintf(subsigid, 3, "%hu", bcomp->ref_subsigid);
 
             /* verify the ref_subsigid */
             val = cli_ac_chklsig(subsigid, subsigid + strlen(subsigid), mdata->lsigcnt[bcomp->lsigid[1]], &evalcnt, &evalids, 0);
-
-            if (subsigid) {
-                free(subsigid);
-                subsigid = NULL;
-            }
 
             if (val != 1) {
                 bcm_dbgmsg("cli_bcomp_scanbuf: could not verify a match for lsig reference subsigid (%s)\n", subsigid);
--- a/libclamav/matcher-byte-comp.c
+++ b/libclamav/matcher-byte-comp.c
@@ -893,7 +893,6 @@ unsigned char *cli_bcomp_normalize_buffer(const unsigned char *buffer, uint32_t 
     uint32_t pad              = 0;
     uint32_t i                = 0;
     uint16_t opt_val          = 0;
-    uint16_t hex              = 0;
     unsigned char *tmp_buffer = NULL;
 
     if (!buffer) {
@@ -971,14 +970,10 @@ unsigned char *cli_bcomp_normalize_buffer(const unsigned char *buffer, uint32_t 
                 if (((int32_t)norm_len - (int32_t)i) - 2 >= 0) {
                     /* 0000BA -> B0000A */
                     if (isxdigit(hex_buffer[norm_len - i - 2]) || toupper(hex_buffer[norm_len - i - 2]) == 'X') {
-                        if (isxdigit(hex_buffer[norm_len - i - 2])) {
-                            hex = 1;
-                        }
                         tmp_buffer[i] = hex_buffer[norm_len - i - 2];
                     } else {
                         /* non-hex detected, our current buffer is invalid so zero it out and continue */
                         memset(tmp_buffer, '0', norm_len + 1);
-                        hex = 0;
                         /* nibbles after this are non-good, so skip them */
                         continue;
                     }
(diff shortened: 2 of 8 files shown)


libclamav/ole2_extract.c
```diff
@@ -1295,7 +1295,7 @@ static cl_error_t handler_enum(ole2_header_t *hdr, property_t *prop, const char
                     if (prop->size < (int64_t)hdr->sbat_cutoff) {
                         if (!ole2_get_sbat_data_block(hdr, hwp_check, prop->start_block)) {
                             status = CL_EREAD;
-                            break;
+                            goto done;
```
micahsnyder, line 1298: I don't think we want to do this. I think the correct solution is to remove `status = CL_EREAD;` instead.

libclamav/ole2_extract.c
```diff
@@ -1306,7 +1306,7 @@ static cl_error_t handler_enum(ole2_header_t *hdr, property_t *prop, const char
                     } else {
                         if (!ole2_read_block(hdr, hwp_check, 1 << hdr->log2_big_block_size, prop->start_block)) {
                             status = CL_EREAD;
-                            break;
+                            goto done;
```
micahsnyder, line 1309: I don't think we want to do this. I think the correct solution is to remove status = CL_EREAD; instead.

libclamav/ole2_extract.c
```diff
@@ -1319,7 +1319,7 @@ static cl_error_t handler_enum(ole2_header_t *hdr, property_t *prop, const char
                         hwp_new = cli_calloc(1, sizeof(hwp5_header_t));
                         if (!(hwp_new)) {
                             status = CL_EMEM;
-                            break;
+                            goto done;
```
micahsnyder, line 1322: This one is good though, don't want to keep going if we're out of memory.

But this could introduce a leak, where `hwp_check` isn't free'd.  We should initialize it to NULL, set it to NULL after we free it on line `1334`, and then free it after `done:` if it is not NULL.

libclamav/matcher-byte-comp.c
```diff
@@ -469,7 +469,7 @@ cl_error_t cli_bcomp_scanbuf(const unsigned char *buffer, size_t buffer_length,
 
     uint32_t evalcnt = 0;
     uint64_t evalids = 0;
-    char *subsigid   = NULL;
+    char subsigid[3];
```
micahsnyder, line 472: Ha, seems like such a silly malloc. Good call making it a stack variable.

clamdscan/proto.c
```diff
@@ -240,7 +240,6 @@ static int chkpath(const char *path)
 {
     int status = 0;
     const struct optstruct *opt;
-    char *real_path = NULL;
```
micahsnyder, line 243: Oops, can't believe I left this in there.

libclamav/others_common.c
```diff
@@ -1159,13 +1159,13 @@ char *cli_newfilepath(const char *dir, const char *fname)
     const char *mdir;
     size_t len;
 
+    mdir = dir ? dir : cli_gettmpdir();
+
     if (NULL == fname) {
         cli_dbgmsg("cli_newfilepath('%s'): fname argument must not be NULL\n", mdir);
```
micahsnyder, line 1165: Honestly I don't really see the point in printing `mdir` in the debug log here. You could also resolve this by  changing this line to:
```c
cli_dbgmsg("cli_newfilepath: fname argument must not be NULL\n");
```

Related, the only call tho this function is just below on line 1188, and has a bad log message as well, since it's clearly not the only reason this function may fail:
```c
    *name = cli_newfilepath(dir, fname);
    if (!*name) {
        cli_dbgmsg("cli_newfilepathfd('%s'): out of memory\n", dir);
        return CL_EMEM;
    }
```
A better message might be something like:
```c
        cli_dbgmsg("cli_newfilepathfd: failed to create path string for new file '%s' in directory '%s'\n", fname, dir);
```
I also wonder if these error messages shouldn't use `cli_errmsg` instead of `cli_dbgmsg`.

libclamav/sis.c
```diff
@@ -151,14 +151,12 @@ enum {
             smax       = tmp;                                                \
             if (((size_t)-1) == tmp) {                                       \
                 cli_dbgmsg("SIS: Read failed during GETD2\n");               \
-                FREE(alangs);                                                \
-                FREE(ptrs);                                                  \
-                return CL_CLEAN;                                             \
+                status = CL_CLEAN;                                           \
+                goto done;                                                   \
```
micahsnyder, line 155: Nice. 👍

micahsnyder (changes requested):

libclamav/others.h
```diff
@@ -1032,4 +1032,99 @@ cl_error_t cli_get_filepath_from_filedesc(int desc, char **filepath);
...
+#define DECLARE_POINTER(type, name) type name = NULL
+#endif
+
+#ifndef STRDUP
+#define STRDUP(buf, var, ...)                  \
+    do {                                       \
+        var = strdup(buf);                     \
+        if (NULL == var) {                     \
+            do {                               \
+                __VA_ARGS__;                   \
```
micahsnyder, line 1048: Wow this VA_ARGS thing to print error messages and set return status codes is pretty wild. I don't think I've seen that before.  😲  I think I like it.

libclamav/ole2_extract.c
```diff
@@ -1331,7 +1317,6 @@ static cl_error_t handler_enum(ole2_header_t *hdr, property_t *prop, const char
                     }
                 } while (0);
 
-                free(hwp_check);
```
micahsnyder, line 1334: Oh good call, we don't have to free it up here because it's not within the loop. 👍

libclamav/ole2_extract.c
```diff
@@ -1685,15 +1665,11 @@ static cl_error_t handler_otf(ole2_header_t *hdr, property_t *prop, const char *
     ret = ret == CL_VIRUS ? CL_VIRUS : CL_SUCCESS;
 
 done:
-    if (NULL != name) {
-        free(name);
-    }
+    FREE(name);
     if (-1 != ofd) {
         close(ofd);
```
micahsnyder, line 1670: Makes you wonder if we should add a `CLOSE_FD()` and a `CLOSE_HANDLE()` (for windows).  Maybe some day, but not today.

libclamav/ole2_extract.c
```diff
@@ -1703,7 +1679,7 @@ static cl_error_t handler_otf(ole2_header_t *hdr, property_t *prop, const char *
                 ret = CL_EUNLINK;
             }
         }
-        free(tempfile);
+        FREE(tempfile);
```
micahsnyder, line 1682: I think this will generate a warning because we've already checked if its not NULL and we've dereferenced it.  I would keep this as a regular `free()`.

micahsnyder (changes requested):

libclamav/ole2_extract.c
```diff
@@ -181,20 +181,21 @@ ole2_list_size(ole2_list_t *list)
 
 int ole2_list_push(ole2_list_t *list, uint32_t val)
 {
-    //check the cli - malloc ?
-    ole2_list_node_t *new_node;
+    DECLARE_POINTER( ole2_list_node_t *, new_node);
```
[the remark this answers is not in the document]
ragusaa: It doesn't have to be a macro, so I am fine removing DECLARE_POINTER, but I do like having all pointers/variables initialized.

libclamav/ole2_extract.c
ragusaa: I was thinking about that as well.  I could add them for this file pretty easily as part of this PR.  Was thinking it could be a good cleanup to start putting these things in as cleanup when I am already changing a file.

libclamav/ole2_extract.c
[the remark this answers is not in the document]
micahsnyder: Yeah me too. An initialized variable is a (more) variable.

micahsnyder (approved):

Issue #184, opened 2021-06-30
clamonacc fatal error


goshansp:
# issue
`clamonacc` dev/104 crashes with below ERROR when a folder like i.e. clamav.git-repo is copied and we aproach of about ~10000 files opened. this issue occurs on all kernel versions. the default ulimit=1024 needs to be raised for this to occur.

```
ERROR: Clamonacc: clamonacc has experienced a fatal error, if you continue to see this error, please run clamonacc with --verbose and report the issue and crash report to the developers
```

# reproduction steps
```
# raise root ulimit
[vagrant@rhel8 ~]$ sudo vim /etc/security/limits.conf # add following two lines 
root    soft    nofile  100000
root    hard    nofile  100000

# verify root ulimit
[vagrant@centos7 git]$ sudo su -
[root@centos7 ~]# ulimit -n
100000


# starting clamonacc
[vagrant@rhel8 ~]$ sudo git/clamav-devel/build/clamonacc/clamonacc -F --config-file=/etc/clamd.d/clamd.conf --stream --verbose

# wrecking havoc:
[vagrant@rhel8 ~]$ git clone https://github.com/Cisco-Talos/clamav.git
[vagrant@rhel8 ~]$ cp clamav-devel target_git_folder -r
[vagrant@rhel8 ~]$ cat /proc/sys/fs/file-nr
# repeat above folder copy until it `signal 11` around 11k
```

# conf
```
TCPSocket 3310
TCPAddr 127.0.0.1
TemporaryDirectory /tmp/clamav
OnAccessExcludeUname clamscan
OnAccessIncludePath /home
OnAccessIncludePath /usr
OnAccessIncludePath /etc
DatabaseDirectory /var/lib/clamav
```

# furher observations
* red hat 8 + Fedora 34 (ulimit 100k / max_user_watches = 500k). this issue can be reproduced
* cannot reproduce when clamav 103.2/103.3 installed from rpm on centos 8 or fedora 34.
* $ git clone https://github.com/Cisco-Talos/clamav.git does not cause crash (tried: 4x no crash)
* after a crash, proper termination (ctrl-c) sometimes impossible, maybe related to https://gitlab.com/goshansp/ansible-role-clamav/-/issues/5
* possibly related to `no such file or directory` https://github.com/Cisco-Talos/clamav/issues/186 ?
* after the crash sometimes one can ctrl-c out, sometimes until segfault ... sometimes a kill -9 from other terminal needed.

# log 
```
[...]

PR #227, opened 2021-07-27
gh-184 Fix segfault and socket fd leak

m-sola:
This fixes a fatal issue that would occur when unable to queue events due to
clamonacc improperly using all available fds.

It also fixes the core fd socket leak issue at the heart of the segfault by
properly cleaning up after a failed curl connection.

### DIFF
--- a/clamonacc/client/client.c
+++ b/clamonacc/client/client.c
@@ -139,6 +139,7 @@ int onas_check_remote(struct onas_context **ctx, cl_error_t *err)
             } else {
                 logg("!ClamClient: Could not connect to clamd, %s\n", curl_easy_strerror(curlcode));
             }
+            curl_easy_cleanup(curl);
             *err = CL_EARG;
             return ret;
         }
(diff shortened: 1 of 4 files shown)


micahsnyder:
Can you please edit the commit message so it starts with `ClamOnAcc: `?
And for the ticket, `gh-184` works but I'd recommend just a link at the end of the commit message so it's easier to refer to when not browsing on github.com.  Something like:
`Fixes https://github.com/Cisco-Talos/clamav/issues/184`

clamonacc/client/protocol.c
```diff
@@ -241,6 +241,7 @@ int onas_dsresult(CURL *curl, int scantype, uint64_t maxstream, const char *file
     int sockd                                                        = -1;
     int (*recv_func)(struct onas_rcvln *, char **, char **, int64_t) = NULL;
 
+
```
micahsnyder, line 244: This file needs to be `clang-format`'ed.

goshansp:
wow, this amazing fix performs perfectly stable in my scenarios - looking forward to the merge! thanks

micahsnyder:
Looks good to me too!

micahsnyder:
Will merge into main, then cherry-pick to rel/0.104.  I'll create a separate PR for dev/0.103.4 that includes a comment in the News about the fix.

PR #230, opened 2021-07-28
CLAM-1531: Fix clamdscan directory scan on Windows

kang-grace:
Cause: _wopen on Windows doesn't work on directories and gives a Permission Denied error.
Old approach used _wopen to get a file descriptor and gets the
realpath from that. New approach redefines realpath() to _fullpath
on Windows, removing the use of _wopen in clamdscan on directories.

_fullpath: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fullpath-wfullpath?view=msvc-160

### DIFF
--- a/libclamav/others.h
+++ b/libclamav/others.h
@@ -993,6 +993,19 @@ cl_error_t cli_ftw(char *base, int flags, int maxdepth, cli_ftw_cb callback, str
 
 const char *cli_strerror(int errnum, char *buf, size_t len);
 
+#ifdef _WIN32
+/**
+ * @brief   Attempt to get a filename from an open file handle.
+ * 
+ * Windows only.
+ *
+ * @param hFile          File handle
+ * @param[out] filepath  Will be set to file path if found, or NULL.
+ * @return cl_error_t    CL_SUCCESS if found, else an error code.
+ */
+cl_error_t cli_get_filepath_from_handle(HANDLE hFile, char **filepath);
+#endif
+
 /**
  * @brief   Attempt to get a filename from an open file descriptor.
  *
--- a/libclamav/others_common.c
+++ b/libclamav/others_common.c
@@ -1461,7 +1461,7 @@ cl_error_t cli_realpath(const char *file_name, char **real_filename)
     char *real_file_path = NULL;
     cl_error_t status    = CL_EARG;
 #ifdef _WIN32
-    int desc = -1;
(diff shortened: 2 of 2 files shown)


micahsnyder (changes requested):
This fixes the error, but unfortunately `_fullpath()` only converts relative paths to absolute paths and it doesn't evaluate symlinks. 

I spent some time on this today trying to figure out a different solution and found that we could instead replace `safe_open()` with a call to `CreateFileA()`, and then convert the handle to a file descriptor before passing it to `cli_get_filepath_from_filedesc()`.  Like this:

```c


    hFile = CreateFileA(file_name,                  // file to open
                        GENERIC_READ,               // open for reading
                        FILE_SHARE_READ,            // share for reading
                        NULL,                       // default security
                        OPEN_EXISTING,              // existing file only
                        FILE_FLAG_BACKUP_SEMANTICS, // may be a directory
                        NULL);                      // no attr. template
    if (hFile == INVALID_HANDLE_VALUE) {
        cli_warnmsg("Can't open file %s: %d\n", file_name, GetLastError());
        status = CL_EOPEN;
        goto done;
    }

    if (-1 == (desc = _open_osfhandle((intptr_t)hFile, O_RDONLY))) {
        cli_warnmsg("Can't convert handle to file descriptor.\n");
        status = CL_EOPEN;
        goto done;
    }
    hFile = INVALID_HANDLE_VALUE; // _open_osfhandle() gives up ownership of the handle to desc.

    status = cli_get_filepath_from_filedesc(desc, &real_file_path);
```

This works pretty well, though it will still print errors on some devices like RAM disks (see the commit message https://github.com/Cisco-Talos/clamav/commit/3a7a20c for details).  That's not _so_ bad though.  It's a pretty uncommon situation.

But it's a little bit silly because `cli_get_filepath_from_filedesc()` will simply convert the file descriptor back to a handle. 

I think ideally we could refactor `cli_get_filepath_from_filedesc()` into two functions, adding one called `cli_get_filepath_from_handle()`.  We could call that one directly from `cli_realpath()` and then not have to do the switch to an fd & then back to a handle. 

@kang-grace would you up for giving that a try?

micahsnyder (changes requested):

micahsnyder:
Perfect! Also, it's working really nicely.

micahsnyder (approved):

micahsnyder:
I also cherry-picked the fix into the rel/0.104 branch

PR #231, opened 2021-07-28
CMake: Fix build on systems lacking inttypes format string macros

micahsnyder:
Define _SF64_PREFIX and _SF32_PREFIX on systems that do not have these
macros: PRIu64, PRIx64, PRIi64, PRIu32, PRIi32, PRIx32

This logic is the same as in the previous build system, here:
https://github.com/Cisco-Talos/clamav/blob/rel/0.102/m4/reorganization/types.m4#L83

Patch courtesy of Mark Fortescue.


micahsnyder:
I cherry-picked these two commits into the `rel/0.104` branch so they make it into the next RC and the stable release.

PR #234, opened 2021-07-29
Fix mspack library name

ragusaa:
Modified to change libmspack to libclammspack to avoid name collisions
on installation.


micahsnyder (approved):

micahsnyder:
I also cherry-picked this to rel/0.104.

PR #236, opened 2021-07-30
Fix clamd_valgrind tests when built with clang-8

ragusaa:
Since strcpy only writes a null-byte to the first terminating byte,
valgrind is throwing errors about uninitialized reads for strncpy's
that could potentially read beyond the null-byte.  Initializing
the whole array to 0 resolves this.


micahsnyder (approved):

micahsnyder:
I also cherry-picked this to rel/0.104

PR #238, opened 2021-07-30
Tests: tune valgrind suppression rule

ragusaa:
Handle the case where thrmgr_dispatch_internal was called from somewhere
other than thrmgr_group_dispatch, triggering the valgrind supression
rule.


micahsnyder:
Also cherry-picked to rel/0.104 to be in the next RC and 0.104 release

PR #242, opened 2021-08-08
docs: Fix a few typos

timgates42:
There are small typos in:
- libclamav/others_common.c
- libclamav/pe.c
- libclamav/unzip.c
- libclamav/yara_arena.c
- libclamunrar/qopen.cpp

Fixes:
- Should read `descriptor` rather than `desriptor`.
- Should read `record` rather than `reocrd`.
- Should read `overarching` rather than `overaching`.
- Should read `initialization` rather than `intialization`.
- Should read `additionally` rather than `additionaly`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md

### DIFF
--- a/libclamav/yara_arena.c
+++ b/libclamav/yara_arena.c
@@ -591,7 +591,7 @@ int yr_arena_allocate_memory(
 // yr_arena_allocate_struct
 //
 // Allocates a structure within the arena. This function is similar to
-// yr_arena_allocate_memory but additionaly receives a variable-length
+// yr_arena_allocate_memory but additionally receives a variable-length
 // list of offsets within the structure where pointers reside. This allows
 // the arena to keep track of pointers that must be adjusted when memory
 // is relocated. This is an example on how to invoke this function:
--- a/libclamav/yara_arena.c
+++ b/libclamav/yara_arena.c
@@ -591,7 +591,7 @@ int yr_arena_allocate_memory(
 // yr_arena_allocate_struct
 //
 // Allocates a structure within the arena. This function is similar to
-// yr_arena_allocate_memory but additionally receives a variable-length
+// yr_arena_allocate_memory but additionaly receives a variable-length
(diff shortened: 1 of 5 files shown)


libclamunrar/qopen.cpp
```diff
@@ -63,7 +63,7 @@ void QuickOpen::Load(uint64 BlockPos)
 {
   if (!Loaded)
   {
-    // If loading for the first time, perform additional intialization.
+    // If loading for the first time, perform additional initialization.
```
micahsnyder, line 66: This is 3rd party code, so as tempting as it is to fix, we should leave it as-is.  If you want, you can submit this upstream to the unrar project https://www.rarlab.com/

libclamav/yara_arena.c
```diff
@@ -591,7 +591,7 @@ int yr_arena_allocate_memory(
 // yr_arena_allocate_struct
 //
 // Allocates a structure within the arena. This function is similar to
-// yr_arena_allocate_memory but additionaly receives a variable-length
+// yr_arena_allocate_memory but additionally receives a variable-length
```
micahsnyder, line 594: This is 3rd party code, so as tempting as it is to fix, we should leave it as-is.
It looks like they've fix it since then in their repo: https://github.com/VirusTotal/yara/blob/master/libyara/arena.c#L343
We plan to update our copy, but haven't found the time in recent years ☹

micahsnyder (changes requested):

libclamunrar/qopen.cpp
timgates42: thx restored original

libclamav/yara_arena.c
timgates42: Thanks restored the original

micahsnyder:
Thanks @timgates42 !

Issue #247, opened 2021-08-13
Broken Link to online documentation in README.md


A-Pisani:
The Issue
----------------
In the README.md file the link to step-by-step instructions for source build are broken:
- [Unix/Linux/Mac Instructions](Installing-from-source/Installing-from-source-Unix.md)
- [Windows Instructions](Installing-from-source/Installing-from-source-Windows.md)

They lead to a 404 GitHub page.

micahsnyder:
Whoops!  Thanks for the tip.  Must've been a copy-paste from the clamav-documentation source, which of course wouldn't work 🤦‍♀️

PR #250, opened 2021-08-14
0.104 install docs cleanup and flevel bump

micahsnyder:
Clean up the INSTALL.md and make it clear that folks should see the online documentation (or offline copy in the release tarball) when looking for instructions to install the dependencies and compile clamav for their operating system.

Increase the FLEVEL from 130 to 140.  Reminder, the FLEVEL for 0.103 was 120, so this gives ample room for additional 0.103 patch versions now that 0.103 will be our first Long Term Support (LTS) release.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,179 +1,372 @@
-# Installation Instructions
+# Installing ClamAV
 
-As of ClamAV 0.104, CMake is required to build ClamAV.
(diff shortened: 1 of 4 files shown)


micahsnyder:
Also fixes https://github.com/Cisco-Talos/clamav/issues/247

PR #261, opened 2021-08-21
Yara: Fix crash with partially loaded yara files

micahsnyder:
Yara rule files may contain multiple signatures. If one of the
signatures fails to load because of a parse error in the yara rule
condition, the rest of the rules still load. This is fine, but it seems
that something isn't properly cleaned up, so there end up being runtime
crashes when running the correctly loaded rules as a result.

Specifically, the crash occurs because of an assert() that expects the
operation stack to be empty and it is not. A simple fix is to print an
error or debug message instead of crashing. It's not the right fix, but
it at least prevents crash.

Resolves: https://bugzilla.clamav.net/show_bug.cgi?id=12077

Also fixed a bunch of warnings in the yara module caused by comparing
different integer types.

### DIFF
--- a/libclamav/yara_exec.c
+++ b/libclamav/yara_exec.c
@@ -144,13 +144,16 @@ int yr_execute_code(
 
   YR_RULE* rule;
   YR_STRING* string;
+#if REAL_YARA
   YR_MATCH* match;
+#endif
   YR_OBJECT* object;
   YR_OBJECT_FUNCTION* function;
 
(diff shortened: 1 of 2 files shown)


micahsnyder:
Also cherry-picked to:
- 0.103.4: 974bdff7ca3dd9a5835e18851da4b12ad26e78d5
- 0.104.1: 111faeb7725be77d11b0fea787bcaa6eef055526

PR #264, opened 2021-08-22
CMake: restore clamunrar SO version, match clamav

micahsnyder:
The libclamunrar (and libclamunrar_iface) SO versions tracked
libclamav's SO version in the old Autotools build system.

We accidentally rolled it backwards, setting it to be similar to UnRAR's
project version. Since the official UnRAR project doesn't have a Unix SO
version that we "should" match, and to prevent the theoretical
possibility of having a collision if an old and new clamav were
installed on the same box, we should make libclamunrar's version track
libclamav as it was before (and before 0.104 is released with CMake
being the stable, and only, build system).


micahsnyder:
@bleve ^ Thanks for pointing out this issue.

PR #279, opened 2021-09-01
CMake: support macOS code signing during build

micahsnyder:
To build with code signing, the macOS build must have:
```bash
  -G Xcode \
  -D CLAMAV_SIGN_FILE=ON \
  -D CODE_SIGN_IDENTITY="...your codesign ID..." \
  -D DEVELOPMENT_TEAM_ID="...your team ID..." \
```

You can find the codesign ID using:
```bash
/usr/bin/env xcrun security find-identity -v -p codesigning
```

The team ID should also be listed in the identity description.

You can verify that codesigning worked like this:
```bash
❯ codesign --verify --verbose ./clambc/RelWithDebInfo/clambc
./clambc/RelWithDebInfo/clambc: valid on disk
./clambc/RelWithDebInfo/clambc: satisfies its Designated Requirement
```

This commend for an unsigned file would report: `code object is not signed at all` and exit with with status code `1`.


micahsnyder:
Also cherry-picked to the `dev/0.104.1` branch

PR #288, opened 2021-09-08
Adding Discard Old builds + Artifcats

rsundriyal:
This will keep builds history and artifacts rotating.


micahsnyder:
For the commit message, I would change it to "jenkins: discard old builds + artifacts" so that at a glance people know it has to do with jenkins, and to fix the typo.

PR #316, opened 2021-10-05
cmake: Fix Windows .msi installer

val-ms:
At present the .msi installer is only installing documentation component
files and the vcredist files but fails to install clamav libraries,
programs, and dependencies.

It appears that explicitly installing the NEWS & README files under the
documentation component before calling "include(CPack)" was causing the
MSI installer to think it needed to install the documentation component
but nothing else.

This commit removes the component name, since we don't want to use
components in the Windows MSI installer anyways. This appears to resolve
the issue so that the MSI installer installs all the desired files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -319,8 +319,7 @@ if (WIN32)
         FILES
             ${CMAKE_CURRENT_SOURCE_DIR}/NEWS.md
             ${CMAKE_CURRENT_SOURCE_DIR}/README.md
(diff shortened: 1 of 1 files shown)


ragusaa (approved):

micahsnyder:
Also cherry-picked to `dev/0.104.1`: https://github.com/Cisco-Talos/clamav/commit/08ca25bef085f1b0fa74de3946fc0c5b0b390682

PR #322, opened 2021-10-07
Clam 1545 freshclam 403 cooldown 24 hours

micahsnyder:
* freshclam: 24-hour cooldown for 403 (forbidden)

  The ClamAV CDN may send a 403 (forbidden) response for database downloads to networks that are explicitly blocked for one reason or another, or for download requests from out-of-date freshclam clients or clients other than freshclam and cvdupdate.
  
  The volume of data serving 403 responses to clients that are blocked and retry in a tight loop is considerable.
  
  This commit seeks to remedy that for future freshclam versions by extending freshclam's self-regulated cool-down for 429 (retry later) responses to include 403 (forbidden) responses.
  
  The cooldown period if a 403 is received will be 24 hours.

* test: extend freshclam 403 test to test cool-down

  Extends the existing freshclam 403 test to check if a repeated freshclam 403 response outputs the cool-down message instead of the original 403 response message.

### DIFF
--- a/libfreshclam/libfreshclam.c
+++ b/libfreshclam/libfreshclam.c
@@ -665,6 +665,15 @@ fc_error_t fc_update_database(
                     break;
                 }
                 case FC_EFORBIDDEN: {
+                    char retry_after_string[26];
+                    struct tm *tm_info;
+                    tm_info = localtime(&g_freshclamDat->retry_after);
+                    if (NULL == tm_info) {
+                        logg("!Failed to query the local time for the retry-after date!\n");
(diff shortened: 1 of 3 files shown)


micahsnyder:
cherrypicked to:
- dev/0.104.1: fd3eddf91beedd386cdd71782093748ad531d337 & 028a52cb7d30903bb71d3089c0cde9e673e37d87
- dev/0.103.4: a13962bee9a5dba77e8a7bb405b802f5a1bca9c8

PR #331, opened 2021-10-16
Use all the link line from ncurses pkg-config

lu-zero:
Otherwise it would fail at link time if ncurses has a stand alone tinfo library.

### DIFF
--- a/cmake/FindCURSES.cmake
+++ b/cmake/FindCURSES.cmake
@@ -58,11 +58,7 @@ if(NCURSES_NOT_FOUND EQUAL -1)
     set(HAVE_LIBNCURSES 1)
     set(CURSES_INCLUDE "<ncurses.h>")
 
-    find_library(CURSES_LIBRARY
-      NAMES ncurses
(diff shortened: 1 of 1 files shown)


micahsnyder:
@lu-zero can you provide me with a distro/version system description where the build would fail without this fix so I can test it?

lu-zero:
The standard Gentoo x86_64 is one.

micahsnyder:
I only just realized there are docker images. `gentoo/stage3:latest`, for example.  Maybe I can test this after all. 
They are MASSIVE though!

micahsnyder:
I also cherry-picked this back to `dev/0.104.1` https://github.com/Cisco-Talos/clamav/commit/5678b555e1294ce983eb130adba902767dadbe44

sthen:
This breaks systems which don't have pkg-config files for curses, like OpenBSD

micahsnyder:
@sthen  will this help for now until we can make it work with and without pkg-config? https://github.com/Cisco-Talos/clamav/issues/358#issuecomment-961484921

PR #334, opened 2021-10-21
Clam 1576 clam 1240 clam 1584 memory leaks

micahsnyder:
* Fix leak in email parser when using gen-json option

  The `mbox.c`::`messageGetFilename()` function returns a COPY of the filename, which must be `free()`'d.
  Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31775

  Also fixed and issue where the email parser may fail to finish parsing some multipart emails when `--gen-json` is enabled if `messageGetJObj()` or `cli_json_addowner()` fail. These should be non-fatal failures. The rest of the (broken) email should be parsed.

  The email parser (`mbox.c` & `message.c`) also has a lot of assertions instead using if()'s for error handling.

  Because of the complexity of the email parser, it's unclear for many of the assertions if they could be triggered based on user input like scanning a malformed email. So to be safe, I've replaced all of the assertions in this parser with error handling to fail gracefully.

* Fix memory leak in zip parser error handling

  The zip parser may leak a string for a zip record filename if the
  the record in the central directory is the last record.

  This fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=31760

  It also appears that there may be a related record filename leak if an error occurred when indexing the files in the central directory header, but I don't have any test file for this but it was an obvious fix.

### DIFF
--- a/libclamav/mbox.c
+++ b/libclamav/mbox.c
@@ -38,6 +38,7 @@
 #include <errno.h>
 #include <assert.h>
 #include <string.h>
+#include <stdbool.h>
 #ifdef HAVE_STRINGS_H
 #include <strings.h>
 #endif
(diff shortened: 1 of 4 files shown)


micahsnyder:
Fixed minor merge conflict in mbox.c.

micahsnyder:
Force-pushed fix to remove extra `records_count++;` that was left in by mistake and caused a test failure.  Everything looks good now, all validated with the fuzz regression tests as well.

micahsnyder:
cherrypicked to:
- dev/0.104.1: 7ed8c5f4234b9d7b7d945356a57d7c2f908452ba & b03713ca463d84d3c480fac88bc99a64f00d0da9
- dev/0.103.4: 96490d9d8a2f72cf7c637608f5e1e88f607eb2ba & 7028ee6854894f5b5d408e933435cc6cdc509331

PR #338, opened 2021-10-24
ClamD: Fix clamdscan error with broken symlink on macOS

micahsnyder:
The `realpath()` function on macOS will fail when scanning a symlink
that doesn't exist because it tries to determine the real path of the
thing the symlink points to and fails when that thing doesn't exist.
This behavior is different than on Linux or FreeBSD Unix.

I'm fixing this by opening the symlink with `O_SYMLINK` then getting
realpath of the link using `cli_get_filepath_from_filedesc()`, much
like we do on Windows.

Side note: I did try using macOS's _DARWIN_BETTER_REALPATH variant to
see if that resolved the issue, but it didn't seem to.

This resolves https://bugzilla.clamav.net/show_bug.cgi?id=12792

This commit also removes the problematic "access denied" clamd test from
the test suite. This commit broke that test on macOS because the error
message when clamdscan fails to `open()` the file to check the
realpath() is different than the error message when clamd tries to scan
it, but has access denied.

(It's like "Permission denied" instead of "Access denied")

It's not worth #ifdef'ing around macOS to get a test pass because this
test is already causing problems in 32-bit Docker environments where
access isn't actually denied to the user (!). Honestly, it's not worth
keeping the test that simply verifies ClamD's error message when denied
matches expectations.

I also switched to use the C_DARWIN macro instead of __APPLE__ because
I kno C_DARWIN is defined in clamav-config.h on macOS, and I found that
__APPLE__ wasn't defined when I tested.

### DIFF
--- a/libclamav/others_common.c
+++ b/libclamav/others_common.c
@@ -1402,7 +1402,8 @@ cl_error_t cli_get_filepath_from_filedesc(int desc, char **filepath)
         goto done;
     }
 
-#elif __APPLE__
+#elif C_DARWIN
+
     char fname[PATH_MAX];
     memset(&fname, 0, PATH_MAX);
 
@@ -1462,6 +1463,8 @@ cl_error_t cli_realpath(const char *file_name, char **real_filename)
     cl_error_t status    = CL_EARG;
 #ifdef _WIN32
     HANDLE hFile = INVALID_HANDLE_VALUE;
+#elif C_DARWIN
+    int fd = -1;
 #endif
 
     cli_dbgmsg("Checking realpath of %s\n", file_name);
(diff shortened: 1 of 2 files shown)


ragusaa (approved):

micahsnyder:
Rebased to watch it go through the test pipelines.

micahsnyder:
Test pipelines on Jenkins looked good.

micahsnyder:
Cherry-picked to dev/0.104.1 as well.  
NOT cherry-picked back to 0.103 because the code in this area changed enough in 0.104 that the merge conflict isn't trivial.  

So this bug fix will only be present in 0.104.1 and 0.105.

PR #339, opened 2021-10-26
Fix recursion limit alert & other issues introduced in scan recursion fix (#295)

val-ms:
- Fix recursion limit alert

  The previous commit broke alerting when exceeding the recursion limit because recursion tracking is so effective that by limiting the final layer of recursion to a scan of the fmap, we prevented it from ever hitting the recursion limit.

  This commit removes that restriction where it only does an fmap scan (aka "raw scan") of files that are at their limit so that we can actually hit the recursion limit and alert as intended.

  Also tidied up the cache_clean check so it checks the `fmap->dont_cache_flag` at the right point (before caching) instead of before setting the "CLEAN" verdict. Note: The `cache_clean` variable appears to be used to record the clean status so the `ret` variable can be re-used without losing the verdict. This is of course only required because the verdict is stored in the error enum. *cough*

  Also fixed a couple typos.

- Fix issues spotted by Coverity from fmap recursion fix

  CID 361074: fmap.c: Possible invalid dereference if status != success and the new map was not yet allocated.

  CID 361077: others.c: Structurally dead code revealed a bug in the cli_recursion_stack_get_size() function.

  CID 361080, 361078, 361083: sigtool.c: Inverted check for if engine needs to be free'd, could leak the engine structure.

  CID 361075: sigtool.c: Missed a `return -1` that should've been `goto done;` and would leak the new_map buffer.

  CID 361079: sigtool/vba.c: Checking if we should free the new_map on failure only if ctx also needs to be free'd, which would leak the new_map if ctx was not allocated yet.

### DIFF
--- a/libclamav/fmap.c
+++ b/libclamav/fmap.c
@@ -125,7 +125,7 @@ fmap_t *fmap_check_empty(int fd, off_t offset, size_t len, int *empty, const cha
         *empty = 1;
         return NULL;
     }
-    if (!CLI_ISCONTAINED(0, st.st_size, offset, len)) {
+    if (!CLI_ISCONTAINED_0_TO(st.st_size, offset, len)) {
         cli_warnmsg("fmap: attempted oof mapping\n");
         return NULL;
     }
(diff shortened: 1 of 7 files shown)


micahsnyder:
Cherry-picked these to `dev/0.104.1`

Issue #353, opened 2021-11-03
clamav 0.104.1 Could NOT find CURSES - using ncursesw (wide-character)


Devstellar:
Clamav 0.104.1 fails to detect CURSES when ncursesw is installed (wide-character).
However Clamav 0.104.0 detects CURSES with same ncursesw installed.

This issue is somehow related to #292 

First of all, in ncurses:
```
 --enable-widec

    This switch causes wide-character libraries (e.g., libncursesw.so.6.2) to be built instead of normal ones (e.g., libncurses.so.6.2). These wide-character libraries are usable in both multibyte and traditional 8-bit locales, while normal libraries work properly only in 8-bit locales. Wide-character and normal libraries are source-compatible, but not binary-compatible.
```
Also:
```
# cat /usr/lib64/libncurses.so
INPUT(-lncursesw)
```
So, basically systems has ncurses.so that redirects to ncursesw as ncursesw can handle "normal" curses calls.

0.104.0 detects library without problem (clamdtop working). Logs:
-- Found CURSES: /usr/lib64/libncurses.so

0.104.1 gives error. Logs:
CMake Error at /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  Could NOT find CURSES (missing: CURSES_LIBRARY)
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindCURSES.cmake:65 (find_package_handle_standard_args)
  CMakeLists.txt:490 (find_package)

-- Configuring incomplete, errors occurred!
See also "/compile/clamav-0.104.1/build/CMakeFiles/CMakeOutput.log".
See also "/compile/clamav-0.104.1/build/CMakeFiles/CMakeError.log".

micahsnyder:
This works for me.  I also confirmed `clamdtop` works if you remove `ncurses` and just use `ncursesw` on a box with both.

micahsnyder:
This issue should be resolved by:
https://github.com/Cisco-Talos/clamav/pull/355
https://github.com/Cisco-Talos/clamav/pull/364

The fixes have also been backported for the 0.104.3 patch release here: 
https://github.com/Cisco-Talos/clamav/pull/480

Issue #292, opened 2021-09-12
clamav 0.104.0 libcursesw support


dilyanpalauzov:
I do not need the curses apps.  clamav 0.103.3/configure worked without curses, the Cmake build fails, if it cannot find libcurses.  For the record, I have instead libncursesw .

* Permit building clamd, freshclam, milter, and others, when curses are missing.

dilyanpalauzov:
So, I want to compile clamd and clamscan, without clamdtop, but the build system does not allow me)

micahsnyder:
I've been thinking about how to respond to this. Yes clamdtop is not optional when configuring the build in 0.104.  Making it optional like it was with the previous build system would be mostly trivial. But reducing build configuration complexity was one of my goals when starting over with CMake, which is why the option doesn't exist. I don't want to have to test additional build permutations to serve a very small minority that could get by simply by installing all of the project dependencies. This is why I'm hesitant to bring it back.  I'm inclined to say "sorry, you need ncurses or pdcurses." Mixed feelings. 🙁

But you raise the point that you have ncursesw.  I hadn't heard of ncursesw before.  I did a search and found this helpful page describing the library and common pitfalls when adding support for it http://www.roguebasin.com/index.php?title=Ncursesw
So I guess ncursesw is a fork of ncurses to support UTF-8 that "is surprisingly difficult to get working with C programs". 
I do have a desire to improve ClamAV's UTF-8 support.  So ... now I am wondering if we should try to support ncursesw in the future?

For now I guess I need to know: does ncurses play nicely with ncursesw?  Can you install ncurses to get over this hurdle and we can deal with ncursesw some other day?

dilyanpalauzov:
I run Linux from Scratch, I have setup ncursesw and I do not want to change it, as this can cause problems with booting/logging in, since bash/readline also uses ncursesw.

ncurses and ncursesw are compiled from the same source code from https://invisible-island.net/ncurses/ncurses.html . I see now only ftp links for downloading tagged relleases. I compile it with

> ./configure --enable-widec --without-cxx-binding --enable-pc-files --with-shared --without-debug --without-normal --with-termlib --with-ticlib --enable-sp-funcs --enable-ext-colors --enable-ext-mouse --enable-ext-putwin --enable-sigwinch

I guess, --without-normal avoids creating libncurses while --enable-widec creates libncursesw .

adb014:
Isn't just as simple as ?

```
[...]

PR #355, opened 2021-11-04
CMake: Fix FindCurses.cmake try packate ncursesw if ncurses not found

Devstellar:
Change to pkg_search_module, first try to find ncurses pkg, if not found, then try with ncursesw pkg

### DIFF
--- a/cmake/FindCURSES.cmake
+++ b/cmake/FindCURSES.cmake
@@ -43,7 +43,7 @@ The following cache variables may also be set:
 
 find_package(PkgConfig QUIET)
 # First try for NCurses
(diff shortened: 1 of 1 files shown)


Devstellar:
This should fix #353 and maybe #292

micahsnyder:
This works for me. 

I found that most `cursesw-dev` packages install `libcurses.so` as well as the `libcursesw.so`.

So I also confirmed that modifying this PR to use `ncursesw`-only on a box with both `libncurses.so` and `libncursesw.so` would link with the `w` variant and that `clamdtop` works.

PR #357, opened 2021-11-04
remove most -Wsign-compare warnings

bigio:
remove most warnings spotted by gcc -Wsign-compare


libclamav/yara_arena.c
```diff
@@ -168,7 +168,7 @@ int _yr_arena_make_relocatable(
   YR_RELOC* reloc;
   YR_ARENA_PAGE* page;
 
-  size_t offset;
+  ssize_t offset;
```
micahsnyder, line 171: This also doesn't feel right to me. Offsets usually shouldn't be signed.  I also noted that you missed one `offset = va_arg(offsets, size_t);`.  

Our old version of the libyara headers has the offset as an `int32_t`, which of course feels wrong... so I looked in the newer version of the full yara source and it look they changed it at some point to a `uint32_t`: https://github.com/VirusTotal/yara/blob/master/libyara/include/yara/arena.h#L48-L71

I dug further and found that they changed their `offset` variable in their `arena.c` to a `size_t`: https://github.com/VirusTotal/yara/commit/09027e5350c5f6ce8b011318717508fb0ef2eff2

I think the correct fix here would be to keep it as `size_t` and instead change the while condition to `while (offset != EOL)`.

libclamav/yara_arena.h
```diff
@@ -80,7 +80,7 @@ void* yr_arena_base_address(
 void* yr_arena_next_address(
     YR_ARENA* arena,
     void* address,
-    int offset);
+    size_t offset);
```
micahsnyder, line 83: Looks like the upstream libyara project made this and other related changes. Seems okay to me. https://github.com/VirusTotal/yara/commit/f502bdf4903545cfc75dfe15bed650138da53a81

I don't mind making this change, but I would note this will all get replaced when we finally get around to upgrading our yara-derived sources. Related discussion in: https://github.com/Cisco-Talos/clamav/issues/283

micahsnyder (changes requested):

micahsnyder:
@bigio the commit history for this PR is messed up.  It looks like maybe you pulled/merged the `main` branch into your development branch, instead of rebasing.  I won't be able to merge or even review it in this state.  Please fix it or create a replacement that is fixed.

bigio:
fixed on pr #452

PR #363, opened 2021-11-05
Add lgtm.yml file for newer CMake support

micahsnyder:
lgtm (CodeQL) is still using and older version of Ubuntu where the CMake
version isn't new enough to build Clam. Adding lgtm.yml so we can
install a newer version with Pip to get lgtm integration working again.

### DIFF
--- a/lgtm.yml
+++ b/lgtm.yml
@@ -1,0 +1,16 @@
+path_classifiers:
+  test:
+    - "unit_tests/*"
+queries:
+  - exclude: cpp/use-of-goto
+extraction:
+  cpp:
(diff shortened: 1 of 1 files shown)


micahsnyder:
I give up. I wasn't able to get it to use the `cmake` installed by `pip`. Not when export'ing it into the PATH, not when explicitly calling it. No matter what it ends up doing default "if [-f CMakeLists.txt ]" routine, followed by using the system-installed CMake. WIll just have to wait for lgtm to upgrade from ubuntu 16 -> 18+

PR #361, opened 2021-11-05
Bump version to 0.104.2 and FLEVEL to 142

micahsnyder:

PR #364, opened 2021-11-05
CMake: curses detection with-or-without pkg-config

micahsnyder:
FreeBSD and OpenBSD package maintainers identified failures when
pkg-config .pc files not present for curses/ncurses.

Patch courtesy of Stuart Henderson.

### DIFF
--- a/cmake/FindCURSES.cmake
+++ b/cmake/FindCURSES.cmake
@@ -58,8 +58,14 @@ if(NCURSES_NOT_FOUND EQUAL -1)
     set(HAVE_LIBNCURSES 1)
     set(CURSES_INCLUDE "<ncurses.h>")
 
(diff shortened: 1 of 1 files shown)


micahsnyder:
We should cherry-pick this to dev/0.104.2 after https://github.com/Cisco-Talos/clamav/pull/361 has merged

micahsnyder:
Related: https://github.com/Cisco-Talos/clamav/pull/355
Maybe this one should also try for cursesw?

micahsnyder:
Rebased & added a commit to support ncursesw when pkg-config files aren't present.

micahsnyder:
This is largely someone else's work, and I added just a single word.  We'll consider me the reviewer. I'm happy it.  Merging.

PR #369, opened 2021-11-11
Fix #368

### LINKED ISSUE #368: Docker container missed to expose port and health check
Describe the bug
----------------
This is using multi-stage docker build, while the port and health check info incorrectly placed in immediate stage of the build instead of in the final build.
https://github.com/Cisco-Talos/clamav/blob/b5e4d953738ad8c804f297653369ba2b458d126e/Dockerfile#L11-L14

How to reproduce the problem
----------------------------
Running docker image, do not see any health check occurred.

Recommended Fix
-----------
Place those statement to the final stage of the build.

sammyhk:
Fix missed to expose port and health check in docker container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,6 @@
 # hadolint ignore=DL3007  latest is the latest stable for alpine
 FROM index.docker.io/library/alpine:latest AS builder
 
-LABEL maintainer="ClamAV bugs <clamav-bugs@external.cisco.com>"
-
-EXPOSE 3310
-EXPOSE 7357
-
-HEALTHCHECK CMD "clamdcheck.sh"
-
 WORKDIR /src
 
 COPY . /src/
@@ -89,6 +82,11 @@ RUN apk add --no-cache \
 
 FROM index.docker.io/library/alpine:latest
 
+LABEL maintainer="ClamAV bugs <clamav-bugs@external.cisco.com>"
+
+EXPOSE 3310
+EXPOSE 7357
+
 RUN apk add --no-cache \
         fts \
         json-c \
@@ -110,5 +108,7 @@ RUN apk add --no-cache \
 COPY --from=builder "/clamav" "/"
 COPY "./dockerfiles/clamdcheck.sh" "/usr/local/bin/"
 COPY "./dockerfiles/docker-entrypoint.sh" "/init"
+
+HEALTHCHECK CMD "clamdcheck.sh"
 
 ENTRYPOINT [ "/init" ]


oliv3r:
@sammyhk, thank you for this fix :)

Could you please move those changes just under the second FROM line? This helps in layer optimization. By keeping things above the run, docker can use the caches up until the HEALTHCHECK. While not super critical in terms of caching layers here, it is a better design pattern within Dockerfiles. You can see similar changes in my merge request !380

Other then that, lgtm @micahsnyder and can be merged after those changes :)

sammyhk:
Moved LABEL, EXPOSE, HEALTHCHECK and ENTRYPOINT statement immediate after FROM statement for utilizing docker layer optimization

Dockerfile
```diff
@@ -89,6 +82,15 @@ RUN apk add --no-cache \
...
 FROM index.docker.io/library/alpine:latest
 
+LABEL maintainer="ClamAV bugs <clamav-bugs@external.cisco.com>"
+
+EXPOSE 3310
+EXPOSE 7357
+
+HEALTHCHECK CMD "clamdcheck.sh"
+
+ENTRYPOINT [ "/init" ]
```
oliv3r, line 92: Not sure why you also moved the entry point; as the file does not exist at this place; so logically this is a bit murkey. Yes the statement is optimized in docker cache layers; but it potentially actually breaks things, as its undefined here. Moving the COPY up to here is also troublesome, because we don't have a container yet, and you don't want to risk anything getting over-written etc.

oliv3r: Btw, @micahsnyder do you also want a 'clean' commit history, with just a single commit? I guess this change is trivial enough that a 'squash commits' merge would fix any issue, otherwise, I'd recommend @sammyhk to cleanup the commit history as well. But that's up to @micahsnyder :)

sammyhk: I agree moving up the ENTRYPOINT statement is a bit murky as it is not yet exists at that point. Actually I'm thinking the same thing as the HEALTHCHECK statement as clamdcheck.sh is not yet exists. So I wondered if just move EXPOSE statement is enough or not

oliv3r: You are correct, the healthcheck should probably stay there too :) well caught; I'll update that in my own Dockerfiles where applicable too :) I think I made that mistake elsewhere :)

sammyhk: Clean them up, only move LABEL and EXPOSE statement after 2nd FROM statement. Moved HEALTHCHECK statement after the COPY statements. Cleaned the history as single commit.

oliv3r:
@micahsnyder I don't know if there's an approval button I can push; but to me this MR looks great; if we can merge it; i can rebase mine :)

oliv3r (approved):

micahsnyder:
> @micahsnyder I don't know if there's an approval button I can push; but to me this MR looks great; if we can merge it; i can rebase mine :)

Sorry about the delay. :( Yes definitely will merge now.

micahsnyder:
Thanks for the fix @sammyhk

PR #374, opened 2021-11-17
CLAM-1598 cargo workspace cleanup

shutton:
Tidy the Rust code integration and test integration. This structures `libclamav_rust` as a workspace member crate under a top-level `Cargo.toml`, and enables execution of the crate's unit tests.


micahsnyder:
I would squash these commits into 1, since the first commit adds the libclamav_rust unit tests and the second commit makes it so those tests actually run. 

I'm seeing a build failure on Windows for the unit tests.  It appears to me that the unit test program isn't linked with libclamav.  I suspect this will also fail with the cdiff PR for any `cli_*` API's used by the `cdiff` module once updated.  I'll play with it a bit now to see if I can figure how how fix it.

micahsnyder:
I just squashed down most of the commits, adding detail to the commit messages. I dropped the first commit because we switched back to adding the rust tests within the `unit_tests/CMakeLists.txt` file, for access to the library paths and environment variables and as such don't need to support adding tests early on. 

Here's the diff before/after the force-push:
```c
❯ git diff shutton/CLAM-1598-cargo-workspace-cleanup
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 99485c0e2..564d496fe 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -962,12 +962,6 @@ else()
         COMPONENT libraries)
 endif()

-# include CTest to enable testing before we add any subdirectories which may
-# add additional tests (like with `add_rust_test()`)
-if(ENABLE_TESTS)
-    include(CTest)
-endif()
-
 #
 # ClamAV Build targets!
 #
@@ -993,8 +987,8 @@ if(WIN32)
     add_subdirectory( win32/compat )
 endif()

-add_subdirectory( libclamav )
 add_subdirectory( libclamav_rust )
+add_subdirectory( libclamav )

 if(NOT ENABLE_LIBCLAMAV_ONLY)
     add_subdirectory( common )
@@ -1041,6 +1035,7 @@ if(ENABLE_EXAMPLES)
     add_subdirectory( examples )
 endif()

+include(CTest)
 add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND})
 if(ENABLE_TESTS)
     add_subdirectory( unit_tests )
```

PR #378, opened 2021-11-20
CLAM-1599 integrate cbindgen via build.rs

shutton:
Eliminate the need to have cbindgen installed.  Instead, make it a build-dependency and call from build.rs.

No changes to the config -- still uses `cbindgen.toml`


libclamav_rust/.gitignore
```diff
@@ -0,0 +1 @@
+clamav_rust.h
```
micahsnyder, line 1: Can we have it build the binding in the `build/libclamav_rust` directory instead, so the build system doesn't contaminate the source directory?

libclamav_rust/build.rs
```diff
@@ -26,9 +26,25 @@ const LIB_LINK_WINDOWS: [&str; 4] = ["wsock32", "ws2_32", "Shell32", "User32"];
...
 fn main() {
+    detect_clamav_build();
+    execute_cbindgen();
+}
+
+fn execute_cbindgen() {
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    cbindgen::generate(crate_dir)
+        .expect("Unable to generate bindings")
+        .write_to_file(C_HEADER_OUTPUT);
```
micahsnyder, line 40: There may be an environment variable we could use to set the path for where it writes the file.

shutton: I tweaked `execute_cbindgen()` to check for `ARGS_WORKING_DIRECTORY`, but it still drops into the `libclamav_rust` directory.  Wherever it lands, it needs to be within the `-I` path `cc` is using.  Suggestions?

shutton: Should be resolved now. Looks like the top of the build directory is already in the C include path.

micahsnyder:
I'm seeing a build failure in the Jenkins CI with this branch and am unsure why:
```
error: failed to calculate checksum of: /usr/share/jenkins/workspace/test-pipelines_build-0.105/clamav-devel/libclamav_rust/.cargo/vendor/syn/tests/.gitignore
Caused by:
  failed to open file `/usr/share/jenkins/workspace/test-pipelines_build-0.105/clamav-devel/libclamav_rust/.cargo/vendor/syn/tests/.gitignore`
Caused by:
  No such file or directory (os error 2)
make[2]: *** [libclamav_rust/x86_64-unknown-linux-gnu/debug/libclamav_rust.a] Error 101
```

In Jenkins, our first step creates a tarball with `VENDOR_DEPENDENCIES=ON`, like this:
```sh
                cmake .. -D VENDOR_DEPENDENCIES=ON
                cpack --config CPackSourceConfig.cmake 
```
See https://github.com/Cisco-Talos/clamav/blob/main/Jenkinsfile#L73
After subsequent stages then build using the generated source tarball. 

For whatever reason, in this PR `cargo` is expecting to find `libclamav_rust/.cargo/vendor/syn/tests/.gitignore` but it isn't  included in the tarball.

shutton:
The missing `.gitignore` issue has been identified as an issue with unintentional removal of those files by the Jenkins workflow.

A better solution would likely be to update `cargo` to recursively remove `.gitignore` files while vendoring. However, it only removes those found in the root of the crate.

PR #379, opened 2021-11-22
Tests: Hoist repeated code outside if statement

krnick:
Refactor the repeated code in unit tests in `unit_tests/freshclam_test.py`


micahsnyder:
Thanks @krnick, this is a little tidier!

PR #380, opened 2021-11-22
Dev/docker updates

oliv3r:
A couple of small updates to the docker contianer.

* Delay healthcheck to reduce false positives at start
* Add default timezone to the container (user-overriddable)

Contributes to mailu/mailu#2059

With these changes, mailu could use our container, which includes the base virus database, which could significantly reduce the load on clamav's database servers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,6 +109,6 @@ COPY --from=builder "/clamav" "/"
 COPY "./dockerfiles/clamdcheck.sh" "/usr/local/bin/"
 COPY "./dockerfiles/docker-entrypoint.sh" "/init"
 
-HEALTHCHECK CMD "clamdcheck.sh"
+HEALTHCHECK --start-period=6m CMD "clamdcheck.sh"
 
 ENTRYPOINT [ "/init" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,6 +87,8 @@ LABEL maintainer="ClamAV bugs <clamav-bugs@external.cisco.com>"
 EXPOSE 3310
 EXPOSE 7357
 
+ENV TZ Etc/UTC
+
 RUN apk add --no-cache \
         fts \
         json-c \
@@ -99,6 +101,7 @@ RUN apk add --no-cache \
         ncurses-libs \
         pcre2 \
         tini \
+        tzdata \
         zlib \
     && \
     addgroup -S "clamav" && \
--- a/README.Docker.md
+++ b/README.Docker.md
@@ -85,6 +85,10 @@ Do not use `--pull always` with the larger ClamAV images.
 
 > _Tip_: It's common to see `-it` instead of `--interactive --tty`.
 
+In some situations it may be desirable to set (any of) the containers to a
+specific timezone. The `--env` parameter can be used to set the `TZ` variable
+to change the default of `Etc/UTC`.
+
 ### Running ClamD using a Locally Built Image
 
 You can run a container using an image built locally
--- a/README.Docker.md
+++ b/README.Docker.md
@@ -363,6 +363,10 @@ occasion send a `ping` to `clamd` on the default port and wait for the pong
 from `clamd`. If `clamd` fails to respond, Docker will treat this as an error.
 The healthcheck results can be viewed with `docker inspect`.
 
+When the container starts up, the health-check also starts up. As loading the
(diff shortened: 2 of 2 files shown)


oliv3r:
@micahsnyder I'm not sure why clang-format is checking uncommited source code ... is it running `clang-format-diff`?

Something like this should work (the git fetch is needed (on gitlab) to ensure that all commits are available up until master ($CI_MERGE_REQUEST_TARGET_BRANCH_NAME) so that git diff can be run and piped into `clang-format-diff`.
```sh
git fetch --depth=1 origin "$CI_MERGE_REQUEST_TARGET_BRANCH_NAME" 
git diff --unified=0 --no-color "origin/${CI_MERGE_REQUEST_TARGET_BRANCH_NAME}..HEAD" | clang-format-diff -p1
```

micahsnyder:
@oliv3r  don't worry about the clang-format failure.  clang-format has been flip-flopping on how it wants to format large ternary statements and that's caused some flaky failures. We're using a github action that tests specific specific directories with some exclusions. I don't really feel like fussing with re-doing it to only test changed files. 

I'll clean it up so we can get back to passing.

micahsnyder:
@oliv3r 2 of your 3 commits mostly-duplicates an existing PR that I haven't yet reviewed: 
https://github.com/Cisco-Talos/clamav/pull/369

I'm glad to see that you've requested the same change.  I'd like to merge theirs first, to give them credit for their work.  Would you be ok to update this, adding the ` --start-period=6m` and `ENV TZ Etc/UTC` after I merge their work?

oliv3r:
@micahsnyder I see and absolutely agree. While I hadn't seen his solution; he was 'first' and should most definitely get credit :)

I gave it a review and recommended a change. I can do the change in this branch later if the other repo doesn't move forward, so you can merge it as is too; but would add a second needless commit.

oliv3r:
@micahsnyder updated and rebased ;)

micahsnyder:
Looks great at a glance. I'm on PTO at present and won't be able to properly review it until I'm back next week. 

It will be great to add the documentation updates to the Docker section of the clamav-documentation repository as well.

oliv3r:
@micahsnyder Cisco-Talos/clamav-documentation#16

But do you really want to keep duplicate documentation? Should we not make this a link to the doc repo at some point? or some other easy automated way, rather then having to push around patches and applying them twice? :) Food for thought ;)

micahsnyder:
> @micahsnyder [Cisco-Talos/clamav-documentation#16](https://github.com/Cisco-Talos/clamav-documentation/pull/16)
> 
> But do you really want to keep duplicate documentation? Should we not make this a link to the doc repo at some point? or some other easy automated way, rather then having to push around patches and applying them twice? :) Food for thought ;)

I don't really want duplicate documentation.  I prefer it in the `clamav-documentation` repo.  It just didn't make sense to put it there at the time we merged your initial version of the Docker support.

micahsnyder:
Should I expect to see a "Clamd is up" message from the health check every 6 minutes when doing `docker run <hash>` for a newly built image?  clamd and freshclam seem to be running nicely, I'm just having a little trouble seeing the health check run.

micahsnyder:
> Should I expect to see a "Clamd is up" message from the health check every 6 minutes when doing `docker run <hash>` for a newly built image? clamd and freshclam seem to be running nicely, I'm just having a little trouble seeing the health check run.

@oliv3r  friendly ping -- I'm holding off on this ticket because I am unsure how to test it and if it is working as intended.

oliv3r:
> Should I expect to see a "Clamd is up" message from the health check every 6 minutes when doing `docker run <hash>` for a newly built image? clamd and freshclam seem to be running nicely, I'm just having a little trouble seeing the health check run.

Hey @micahsnyder
Sorry for the late reply. No, the behavior should be exactly as before, except that docker/the-healthcheck won't be run for the first 6 minutes. This to allow clamd to startup and load its database (which can be slow) without triggering health failures.

micahsnyder:
Okay finally circling back around to this. Thanks for your patience.  It looks good to me  `docker inspect <name>` is showing the regular health checks.  The 6 minute delay seems to work well Will merge and backport for the next 0.104 patch version.

oliv3r:
@micahsnyder no problem; you follow my pace on these kind of things :p Thanks!

Issue #245, opened 2021-08-10
clamonacc hanging, causing high system load


frank-fegert:
Describe the bug
----------------

We're using ClamAV (clamd with clamonacc) on RHEL 7 and are experiencing similar issues as described in
https://github.com/Cisco-Talos/clamav/issues/184
The clamonacc seems to hang indefinately causing high system load (>>40 with an 8 CPU system) and will crash immediately if a strace or gdb is attached to the process.
The system this happens most oftenly is a Jenkins build server with a very large amount of files and directories dynamically being generated during builds. The paths of the files and directories seem to contain special ASCII characters (e.g. whitespaces and "@").

We've seen this issue:
- since updating from v0.101.x
- with the stock EPEL packages (https://src.fedoraproject.org/rpms/clamav/blob/epel7/f/clamav.spec) using TCP based communication (due to old curl version v7.29.0 on RHEL) between clamd and clamonacc
- with custom build packages of v0.103.4 and cherry picked patches, **dynamically** linked against the old curl v7.29.0 using TCP based communication (due to old curl version v7.29.0 on RHEL) between clamd and clamonacc
- with custom build packages of v0.103.4 and cherry picked patches **dynamically** linked against an updated curl v7.77.0
- with custom build packages of v0.104.0 (based on https://github.com/Cisco-Talos/clamav/commit/d0d0a8307c07f54eee815bf98db956d1072c2b64), **statically** build against an updated curl v7.77.0 (as described in https://github.com/Cisco-Talos/clamav/issues/199)

How to reproduce the problem
----------------------------

An exact reproducer is currently not known. The issue **seems** to apper while creating a large number of files and directories with names containing special ASCII characters (e.g. whitespaces and "@") under the paths in OnAccessIncludePath.

```
Checking configuration files in /etc

Config file: clamd.d/scan.conf
------------------------------
LogFile = "/var/log/clamd/clamd.scan.log"
LogTime = "yes"
LogSyslog = "yes"
LogVerbose = "yes"
LocalSocket = "/var/run/clamd.scan/clamd.sock"
StreamMaxLength = "536870912"
MaxThreads = "50"
MaxQueue = "200"
MaxFileSize = "536870912"
OnAccessIncludePath = "<Jenkins Path>"
OnAccessExcludeUname = "clamupdate"
OnAccessPrevention = "yes"
OnAccessCurlTimeout = "20000"
OnAccessMaxThreads = "10"
OnAccessRetryAttempts = "1"

[...]

PR #382, opened 2021-11-23
clamonacc: Harmonize signal handling on the different clamonacc threads.

frank-fegert:
Explicit setting of pthread_sigmask in the individual threads
should not be necessary at all(?), since it is already done in
the main process/thread and should be inherited from there. But
since the sigfillset/sigdelset lines are already there, at least
they should be consistent with regard to the undefined behaviour
caused by ignoring SIGFPE, SIGILL, SIGSEGV, or SIGBUS signals.


frank-fegert:
#245

m-sola:
Mentioned this on another commit, but this cleanup doesn't seem like it will impact cleanup code at all, so I have no issue with it.

frank-fegert:

PR #384, opened 2021-11-23
clamonacc: Added debug log output to bundled c-thread-pool.

frank-fegert:
This is not strictly necessary, but i found it rather helpful to know
what is going on during the debugging process and while trying to under-
stand how clamonacc actually works.

### DIFF
--- a/clamonacc/c-thread-pool/thpool.c
+++ b/clamonacc/c-thread-pool/thpool.c
@@ -18,7 +18,9 @@
 #include <time.h>
 #if defined(__linux__)
 #include <sys/prctl.h>
+#include <sys/syscall.h>
 #endif
+#include "output.h"
 
 #include "thpool.h"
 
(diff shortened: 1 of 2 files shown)


frank-fegert:
#245

m-sola:
Another great QoL fix for devs, but historically we prefer not to make modifications to external projects as sometimes makes upkeep trickier. I'll leave final say on this one to @micahsnyder, but I suspect the preferred option here is to make a change to upstream which creates some sort of subscriber-type logging in the c-thread-pool project.

frank-fegert:
@micahsnyder What would be your opinion on this? I'm open to keep this as a patch for our local build or even ask the upstream c-thread-pool project to include this. Just let me know how to proceed on this. Thanks!

micahsnyder:
Thank you for your patience.  I feel terrible for how long I've made you wait on this.  

I don't mind putting in the fix here, but you should definitely ask the upstream project to make the same changes.  If it was a spelling fix or just a warning fix then I'd take a difference stance but this looks fairly important. 

We would of course appreciate knowing what the upstream project does if you can report back.

PR #386, opened 2021-11-23
clamonacc: Added workaround for NULL-pointer usage in clamonacc-ddd thread.

frank-fegert:
The clamonacc-ddd thread (clamonacc/inotif/inotif.c) has currently no
explicit handling for the special inotify kernel events IN_UNMOUNT,
IN_Q_OVERFLOW and IN_IGNORED. It treats the inotify_event structs in
the same way as regular inotify events, although for those special
events the "len" and "name" fields a value of zero respectively NULL.
This can lead to a SIGSEGV of the clamonacc-ddd thread when calling
"strlen(name)" on a NULL value.
This commit adds just a quick'n'dirty workaround for the SIGSEGV and
some logging output. It DOES NOT fix the more overall issue, that a
IN_Q_OVERFLOW inotify event also leads to an out-of-sync situation
between the monitored filesystem/directory and the internal data
structures "ddd_ht" and "wdlt" of clamonacc.

### DIFF
--- a/clamonacc/inotif/inotif.c
+++ b/clamonacc/inotif/inotif.c
@@ -631,7 +631,10 @@ void *onas_ddd_th(void *arg)
 
                 event = (const struct inotify_event *)p;
                 wd    = event->wd;
-                path  = wdlt[wd];
+                if (wd >= 0)
+                    path  = wdlt[wd];
+                else
+                    path  = NULL;
                 child = event->name;
 
                 if (path == NULL) {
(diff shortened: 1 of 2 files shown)


frank-fegert:
#245

frank-fegert:
Added a workaround for another SIGSEGV source in the clamonacc-ddd thread, which was observed during further analysis.

frank-fegert:
@m-sola Thank you very much for the review! :smile: I ran this through `clang-format-11` which only came up with minor whitespace changes in https://github.com/Cisco-Talos/clamav/pull/386/commits/92beae2fc4aa624e580e9a71699f4519642e496d.

micahsnyder:
> @frank-fegert looks like just one merge-conflict. Can I trouble you to rebase it with the upstream `main` branch and resolve it, and then force-push the update?

Nevermind, I just realized that merging the previous PR's without rebasing and testing broke the build because of an unrelated PR that changed the `logg()` function API. I'll take care of it.

I've created https://github.com/Cisco-Talos/clamav/pull/483 to fix that, and replace this PR.

PR #388, opened 2021-11-24
CMake: Rust RelWithDebInfo and MinSizeRel builds

micahsnyder:
RelWithDebInfo is our preferred build type. It has optimizations and
should run faster, but includes debugging symbols for better profiling,
stack traces, etc.

The Rust MinSizeRel support is just Release mode for now. There are
optimizations we can do to shrink it further, but for now at least it
won't actually be Debug (aka slow).


micahsnyder:
@shutton Updated this PR after conflicts.  Ready for review.

ragusaa (approved):

shutton (approved):

PR #396, opened 2021-12-06
Fix length of strftime

AlexanderS:
The max bytes supplied to strftime should be the length of result
string, including the terminating null byte.

Without the extra byte for the terminating null byte, the output is one
byte too long and results in undefined behavior:

> If the length of the result string (including the terminating null
> byte) would exceed max bytes, then strftime() returns 0, and the
> contents of the array are undefined.


micahsnyder (approved):
This change make sense to me and outside of some test infrastructure wonkiness on our side, all the builds look good.

micahsnyder:
Thanks for the fix @AlexanderS

PR #400, opened 2021-12-14
CLAM-1280 cdiff rewrite

shutton:
Redesign cdiff handling in Rust

PR #405, opened 2021-12-20
Generate Rust-bindings for some internal C-headers

micahsnyder:
Use bindgen to generate Rust-bindings for some libclamav internal
functions and structures in an new "sys.rs" module.

"sys.rs" will be generated at build-time and unfortunately must be
dropped in the source/libclamav_rust/src directory rather than under the
build directory. As far as I know, Cargo/Rust provide no way to set an
include path to the build directory to separately place generated files.
TODO: Verify if this is true and move sys.rs to the build directory if
possible.

Using the new bindings with the logging module.
I was unable to get bindgen to create bindings for `cli_dbgmsg()`,
possibly because it is an `inline` function.
Setting `generate_inline_functions(true)` in build.rs hasn't solved it.
TODO: Figure out how to get `bindgen` to generate this for us.


micahsnyder:
The changes in `build.rs` will piggyback on a change in https://github.com/Cisco-Talos/clamav/pull/400 that sets the environment variable `CARGO_CMD=build`.  That's not available in this branch (yet), so to use this you have to do the build with something like:
```sh
# configure
cmake .. -D CMAKE_INSTALL_PREFIX=install

#make
CARGO_CMD=build make -j12       <--- manually set the variable for now, until #400 merges

#install 
make install
```

libclamav_rust/Cargo.toml
```diff
@@ -8,8 +8,11 @@ edition = "2018"
...
 [dependencies]
-log = {version = "*", features = ["std"]}
+log = { version = "*", features = ["std"] }
 
 [lib]
 crate-type = ["staticlib"]
 name = "clamav_rust"
+
+[build-dependencies]
+bindgen = "*"
```
shutton, line 18: Suggest pinning this to "0.59" in case APIs change.

libclamav/CMakeLists.txt
```diff
@@ -21,6 +21,15 @@ if(WIN32)
     include_directories(${CMAKE_SOURCE_DIR}/win32/compat)
 endif()
 
+set(EXTRA_CFLAGS)
+if(NOT CMAKE_C_COMPILER_ID MATCHES "MSVC")
+    # Generate code for functions even if they are fully inlined.
+    # Specifically needed to make sure `cli_dbgmsg` is available for Rust-bindings
+    # See: https://rust-lang.github.io/rust-bindgen/faq.html#why-isnt-bindgen-generating-bindings-to-inline-functions
+    extract_valid_c_flags(EXTRA_CFLAGS -fkeep-inline-functions)
```
shutton, line 29: This might be a bit of a sledgehammer approach for just a few functions.  The functions that are currently defined as inline that you want to export probably just need to have their `static` keyword removed, moved inside the implementation file (if they were defined in the header, as seems to be common), and then just leave a matching declaration in the header.  This is the approach when I did my "make fmap opaque" PoC.

libclamav_rust/build.rs
```diff
@@ -48,6 +53,47 @@ fn main() {
...
+            .clang_arg("-I../libclammspack")
+            .clang_arg(cc_include_path_build)
+            // Generate bindings for *at least* the following.
+            .header("../libclamav/matcher.h")
+            .allowlist_type("cli_matcher")
+            .header("../libclamav/matcher-ac.h")
+            .allowlist_type("cli_ac_data")
+            .allowlist_type("cli_ac_result")
+            .header("../libclamav/others.h")
+            .allowlist_function("cli_ctx")
```
shutton, line 84: Can we define these `const`, as we did with the other things we inspect (e.g., `LIB*`), and then just repeatedly update `bindings`?

libclamav_rust/src/logging.rs
```diff
@@ -25,11 +25,13 @@ use std::os::raw::c_char;
...
 use log::{set_max_level, Level, LevelFilter, Metadata, Record};
 
+use crate::sys;
+
 extern "C" {
-    fn cli_warnmsg(str: *const c_char, ...) -> ();
+    /// bindgen doesn't discover cli_dbgmsg, I think because it is `inline`
+    /// Setting `generate_inline_functions(true)` in build.rs hasn't solved it.
+    /// TODO: Figure out how to get `bindgen` to generate this for us.
     fn cli_dbgmsg(str: *const c_char, ...) -> ();
```
shutton, line 34: The `-> ()` is implied.  I think my `clippy` was flagging that.

.gitignore
```diff
@@ -234,3 +234,7 @@ Cargo.lock
 
 # These are backup files generated by rustfmt
 **/*.rs.bk
+
+# The name of our generated ClamAV-Rust bindings file.
+libclamav_rust/src/sys.rs
```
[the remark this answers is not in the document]
micahsnyder: I was thinking the same thing. I was tempted to name it `cli`, to mirror the `cli_` prefix we use for internal API that aren't stable for public use... but then I thought maybe we want to use `cli` in the future as a module that provides safe wrappers around `sys` bindings.  

I copied the `sys` name from someone else's work.  I agree that it does sort of imply that it comes from the "system" but since it also represents "the unsafe bindings" vs `rs` which would be used for "safe idiomatic wrapper around sys-bindings" I figure `sys` is good enough.  I at least like that it's shord :D

libclamav/CMakeLists.txt
micahsnyder: I don't think the sledgehammer approach of creating functions and symbols for `inline` functions and `static` functions has a performance penalty, except when called using the exported symbol.  The internal code should continue to inline it. 

In this case I want to make sure the `cli_dbgmsg` function has a function generated and symbol exported: https://github.com/Cisco-Talos/clamav/blob/main/libclamav/others.h#L853-L857
Note that it's already not "static", but that isn't enough to make sure the function is generated/exported.  

Another thing I could do is make a non-inlined wrapper for `cli_dbgmsg` (eg: `cli_dbgmsg_no_inline`) and use that, so I don't need to use `-fkeep-inline-functions`.  Maybe that's the way to go.

libclamav_rust/build.rs
micahsnyder: That's an idea. 👍 I'll give it a try.

libclamav/CMakeLists.txt
shutton: There shouldn't be a performance penalty, but it will bloat the executable size a bit, since the compiler will be forced to generate standalone implementations when none may have been required earlier since they were all inlined.

Again, not a big deal -- just results in a larger executable and symbol table than might be required.

libclamav_rust/build.rs
micahsnyder: Updated.  Better?

libclamav/CMakeLists.txt
micahsnyder: I removed it and added a single copy of the `cli_dbgmsg` function (not a wrapper but a full copy, because passing vaargs from one variable param function to another sucks, and I didn't feel like writing the C code.

micahsnyder:
Just rebased the PR to keep up with `main`. Though this PR is still blocked by https://github.com/Cisco-Talos/clamav/pull/400.

@shutton One thing to note about this PR is that it places the `sys.rs` file under the source directory.  Do you know if it's possible to put Rust sources outside of the source `./libclamav_rust/src` directory and put it under the equivalent build directory instead, like we do with generated C sources? I don't like having generated files mixed in with original sources if I can help it.

libclamav_rust/build.rs
```diff
@@ -30,13 +32,46 @@ const LIB_ENV_LINK: [&str; 12] = [
...
+const BINDGEN_TYPES: &[&str] = &["cli_matcher", "cli_ac_data", "cli_ac_result"];
+
+// Find the required functions and types in these headers:
+const BINDGEN_HEADERS: &[&str] = &[
+    "../libclamav/matcher.h",
+    "../libclamav/matcher-ac.h",
+    "../libclamav/others.h",
+    "../libclamav/dsig.h",
+];
+// Find the required headers in these directories:
```
shutton, line 66: Need blank line above for consistency.

shutton (approved):

micahsnyder:
@shutton With the change to generate only in MAINTAINER_MODE and to omit the struct layout tests, this now passes our test pipelines. Any further concerns before I merge it?

shutton:
> @shutton With the change to generate only in MAINTAINER_MODE and to omit the struct layout tests, this now passes our test pipelines. Any further concerns before I merge it?

Nope.

PR #406, opened 2021-12-22
Fix ability to disable filesize limit with libclamav C API

micahsnyder:
You should be able to disable the maxfilesize limit by setting it to
zero. When "disabled", ClamAV should defer to inherent limitations, which
at this time is UINT_MAX - 2 bytes.

This works okay for ClamScan and ClamD because our option parser
converts max-filesize=0 to 4294967295 (4GB). But it is presently broken
for other applications using the libclamav C API, like this:
```c
    cl_engine_set_num(engine, CL_ENGINE_MAX_FILESIZE, 0);
```

The limit checks added for cl_scanmap_callback and cl_scanfile_callback
in 0.103.4 and 0.104.1 broke this ability because we forgot to check if
the `maxfilesize > 0` before enforcing it.

This commit adds that guard so you can disable by setting to `0`.

While working on this, I also found that the `max_size` variables in our
libmspack scanner code are using an `off_t` type, which is a SIGNED integer
that may be 32bit width even on some 64bit platforms, or may be a 64bit
width. AND the default `max_size` when `maxfilesize == 0` was being set to
UINT_MAX (0xffffffff), aka `-1` when `off_t` is 32bits.

This commit addresses this related issue by:
    - changing the `max_size` to use `uint64_t`, like our other limits.
    - verifying that `maxfilesize > 0` before using it.
    - checking that using `UINT32_MAX` as a backup will not exceed the
    max-scansize in the same way that we do with the maxfilesize.

### DIFF
--- a/libclamav/libmspack.c
+++ b/libclamav/libmspack.c
@@ -34,7 +34,7 @@ struct mspack_name {
 
 struct mspack_system_ex {
     struct mspack_system ops;
-    off_t max_size;
+    uint64_t max_size;
 };
 
 struct mspack_handle {
@@ -45,7 +45,7 @@ struct mspack_handle {
     off_t offset;
 
     FILE *f;
-    off_t max_size;
+    uint64_t max_size;
 };
 
 static struct mspack_file *mspack_fmap_open(struct mspack_system *self,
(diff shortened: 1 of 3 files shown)


micahsnyder:
Just updated the commit message a bit to explain what we found

ragusaa (approved):

micahsnyder:
Fixed typo in commit message

PR #419, opened 2022-01-10
Defer error handling from cacheset_add()

shutton:
Also remove unused code from prior implementation (`USE_LRUHASHCACHE`).

### DIFF
--- a/libclamav/cache.c
+++ b/libclamav/cache.c
@@ -33,13 +33,6 @@
 #include "cache.h"
 #include "fmap.h"
 
-// #define USE_LRUHASHCACHE /* The replacement policy algorithm to use */
-#define USE_SPLAY
-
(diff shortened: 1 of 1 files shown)


micahsnyder (approved):
I have no complaints.  Glad to get rid of dead code.  It'll always be in the commit history if we need to look it up.  

Some of the messages were warnings while others were errors.  I'm fine with just using errors in this case, or just warnings.  It doesn't seem to matter which. 

Before merging it, I would change the commit message to have something intelligible to average users skimming the commit history.

shutton:
> Before merging it, I would change the commit message to have something intelligible to average users skimming the commit history.

How about, "Defer error reporting in `cacheset_add()` to prevent running callbacks while mutex is held."

PR #424, opened 2022-01-13
Docker: Enable db update script to work with older versions

micahsnyder:
The database update script has a line to delete freshclam.dat after the
update, but this fails when attempting to update older images that have
mirrors.dat or no dat file at all. This issue is compounded by a bug
where the -t (--tags) option doesn't work so it tries to update all
images every time, and then of course fails on the older ones.

This commit has the script try removing freshclam.dat or mirrors.dat and
allows it to succeed even if neither exist.


micahsnyder:
Note to self: we should backport this in the next 0.104 patch release as well.

micahsnyder:
Cherry-picked to `dev/0.104.3`

PR #425, opened 2022-01-14
unit tests: Fix ck_assert_msg() call

micahsnyder:
The first argument to ck_assert_msg() should be a logical condition.


micahsnyder:
Recreated https://github.com/Cisco-Talos/clamav/pull/367

ragusaa (approved):

PR #428, opened 2022-01-19
fuzz: Fix fuzz target race condition flakiness

micahsnyder:
The fuzz targets that write a file before using the file are using a
buffered file writer that results in a race condition clamav when
reading the file during the scan/load test.

This commit flushes the FILE stream before running the test.


shutton:
`fclose()` is documented to perform a flush of buffered writes prior to closing the handle, so that's a little suspicious.

What I do notice is no checking of the return values from any of the `f*()` calls, which should really be done before trusting that they worked.

micahsnyder:
> `fclose()` is documented to perform a flush of buffered writes prior to closing the handle, so that's a little suspicious.
> 
> What I do notice is no checking of the return values from any of the `f*()` calls, which should really be done before trusting that they work

This fix doesn't seem right after all.  My tired-brain saw the problem go away and I maybe mis-attributed it to the flush. Will cancel the PR and keep playing with it.

Issue #432, opened 2022-01-19
Race condition in fuzz tests


mmmds:
Fuzz tests `clamav_dbload_fuzzer.cpp` and `clamav_scanfile_fuzzer.cpp` create a fixed-name input file. If the tests are running in parallel mode, many threads are accessing the same file. Filenames should be unique per process to make parallel fuzzing effective.

https://github.com/Cisco-Talos/clamav/blob/0037f5825b0b17a789c7eb29c9cb9a2d39c452bc/fuzz/clamav_dbload_fuzzer.cpp#L120

https://github.com/Cisco-Talos/clamav/blob/0037f5825b0b17a789c7eb29c9cb9a2d39c452bc/fuzz/clamav_scanfile_fuzzer.cpp#L122

Issue #431, opened 2022-01-19
Unnecessary keyword in cmake foreach loop


mmmds:
`foreach` loops in clamav/fuzz/CMakeLists.txt unnecessarily use `in` keyword. In fact, it is not treated as a keyword but as an element. It leads to creating `fuzz/clamav_scanmap_in_fuzzer`, `fuzz/clamav_scanfile_in_fuzzer`, `fuzz/clamav_dbload_in_fuzzer`files that do not serve any purpose. `in` keyword can be removed from the loops.

https://github.com/Cisco-Talos/clamav/blob/0037f5825b0b17a789c7eb29c9cb9a2d39c452bc/fuzz/CMakeLists.txt#L46
https://github.com/Cisco-Talos/clamav/blob/0037f5825b0b17a789c7eb29c9cb9a2d39c452bc/fuzz/CMakeLists.txt#L62

micahsnyder:
Huh! I never noticed that.  Nice observation.  I'm working on a fix for #432 and will include this fix there.

micahsnyder:
Elsewhere in the project we have `foreach(m IN LISTS SOME_LIST_VARIABLE)`, as described in https://stackoverflow.com/a/68018445/3430496.  This is probably how I got confused when adding this one.

PR #434, opened 2022-01-21
fuzz: stability of parallel fuzz testing

micahsnyder:
The fuzz targets that write a temp file currently use the same filename
every time. One of the users identified that if the tests are running
in parallel mode, many threads are accessing the same file.
This results in unstable input to the API being tested, where the file
may be overwritten as the function is being tested.

This commit fixes it by putting a random number in the filenames for the
scanfile and dbload fuzz targets.

Resolves: https://github.com/Cisco-Talos/clamav/issues/432

Also fixed a CMake bug that built an extra fuzz target file that doesn't
serve any purpose.

Resolves: https://github.com/Cisco-Talos/clamav/issues/431

### DIFF
--- a/fuzz/clamav_dbload_fuzzer.cpp
+++ b/fuzz/clamav_dbload_fuzzer.cpp
@@ -51,59 +51,11 @@ class ClamAVState
         cl_set_clcb_msg(clamav_message_callback);
 
         cl_init(CL_INIT_DEFAULT);
-
-        dboptions =
-            CL_DB_PHISHING | CL_DB_PHISHING_URLS |
-            CL_DB_BYTECODE | CL_DB_PUA | CL_DB_ENHANCED;
-
-#if defined(CLAMAV_FUZZ_CDB)
-        tmp_db_name = "dbload_tmp_fuzz.cdb";
-#elif defined(CLAMAV_FUZZ_CFG)
-        tmp_db_name = "dbload_tmp_fuzz.cfg";
-#elif defined(CLAMAV_FUZZ_CRB)
-        tmp_db_name = "dbload_tmp_fuzz.crb";
-#elif defined(CLAMAV_FUZZ_FP)
-        tmp_db_name = "dbload_tmp_fuzz.fp";
-#elif defined(CLAMAV_FUZZ_FTM)
-        tmp_db_name = "dbload_tmp_fuzz.ftm";
-#elif defined(CLAMAV_FUZZ_HDB)
-        tmp_db_name = "dbload_tmp_fuzz.hdb";
-#elif defined(CLAMAV_FUZZ_HSB)
-        tmp_db_name = "dbload_tmp_fuzz.hsb";
-#elif defined(CLAMAV_FUZZ_IDB)
-        tmp_db_name = "dbload_tmp_fuzz.idb";
-#elif defined(CLAMAV_FUZZ_IGN)
-        tmp_db_name = "dbload_tmp_fuzz.ign";
-#elif defined(CLAMAV_FUZZ_IGN2)
-        tmp_db_name = "dbload_tmp_fuzz.ign2";
-#elif defined(CLAMAV_FUZZ_LDB)
-        tmp_db_name = "dbload_tmp_fuzz.ldb";
-#elif defined(CLAMAV_FUZZ_MDB)
-        tmp_db_name = "dbload_tmp_fuzz.mdb";
-#elif defined(CLAMAV_FUZZ_MSB)
-        tmp_db_name = "dbload_tmp_fuzz.msb";
-#elif defined(CLAMAV_FUZZ_NDB)
-        tmp_db_name = "dbload_tmp_fuzz.ndb";
-#elif defined(CLAMAV_FUZZ_PDB)
-        tmp_db_name = "dbload_tmp_fuzz.pdb";
-#elif defined(CLAMAV_FUZZ_WDB)
-        tmp_db_name = "dbload_tmp_fuzz.wdb";
-#elif defined(CLAMAV_FUZZ_YARA)
-        tmp_db_name = "dbload_tmp_fuzz.yara";
-#else
-        tmp_db_name = "dbload_tmp_fuzz";
-#endif
     }
 
     ~ClamAVState()
     {
-        if (NULL != tmp_db_name) {
-            unlink(tmp_db_name);
-        }
     }
-
-    const char* tmp_db_name;
-    unsigned int dboptions;
 };
 
 // Global with static initializer to setup an engine so we don't need to do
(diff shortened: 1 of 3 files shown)


fuzz/clamav_dbload_fuzzer.cpp
```diff
@@ -117,16 +125,17 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
     FILE* fuzzdb             = NULL;
     struct cl_engine* engine = NULL;
 
-    fuzzdb = fopen(kClamAVState.tmp_db_name, "w");
+    fuzzdb = fopen(kClamAVState.tmp_db_name.c_str(), "w");
```
shutton, line 128: Could this be `"wx"` to make sure it's impossible for two processes to write to the same file?

fuzz/clamav_scanfile_fuzzer.cpp
```diff
@@ -119,15 +127,16 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
 {
     FILE* fuzzfile = NULL;
 
-    fuzzfile = fopen(kClamAVState.tmp_file_name, "w");
+    fuzzfile = fopen(kClamAVState.tmp_file_name.c_str(), "w");
     fwrite(data, size, 1, fuzzfile);
+    fflush(fuzzfile);
```
shutton, line 132: Shouldn't be necessary.  `fclose()` is documented to do this, or the OS is broken.

shutton (changes requested):

fuzz/clamav_scanfile_fuzzer.cpp
micahsnyder: Oops, didn't mean to commit that.  Yeah it's not required.

fuzz/clamav_dbload_fuzzer.cpp
micahsnyder: Sure! I've just updated the PR with this change and in local testing it also seemed to work nicely.

mmmds:
Hi! I've looked at this PR and I'm afraid it won't solve the problem.

1. Unique filename is based on random number. Even if the number range is huge, there is still a chance for collision to happen.
2. By using "wx" flag you make sure to get rid of filename collisions but it leads to unnecessary crashes.
3. Unique filename is generated in the constructor which is executed once and then shared between many fuzzing processes.

When I run fuzzing the second execution of LLVMFuzzerTestOneInput fails because it reuses the filename and fails because of the "x" flag.
```
$ fuzz/clamav_scanfile_fuzzer
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3387112167
INFO: Loaded 1 modules   (52945 inline 8-bit counters): 52945 [0x13b6d20, 0x13c3bf1), 
INFO: Loaded 1 PC tables (52945 PCs): 52945 [0x13c3bf8,0x1492908), 
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
AddressSanitizer:DEADLYSIGNAL
=================================================================
==6053==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x7f30a58f74a5 bp 0x000000000001 sp 0x7ffd5cb75bf0 T0)
==6053==The signal is caused by a READ memory access.
==6053==Hint: address points to the zero page.
    #0 0x7f30a58f74a5 in fwrite /build/glibc-eX1tMB/glibc-2.31/libio/iofwrite.c:37:3
    #1 0x4ca70f in fwrite (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x4ca70f)
    #2 0x561dc8 in LLVMFuzzerTestOneInput /home/osboxes/clamav/build-fuzz/../fuzz/clamav_scanfile_fuzzer.cpp:131:5
    #3 0x464891 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x464891)
    #4 0x463d9a in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool, bool*) (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x463d9a)
    #5 0x466074 in fuzzer::Fuzzer::ReadAndExecuteSeedCorpora(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x466074)
    #6 0x466339 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x466339)
    #7 0x454926 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x454926)
    #8 0x47e842 in main (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x47e842)
    #9 0x7f30a58980b2 in __libc_start_main /build/glibc-eX1tMB/glibc-2.31/csu/../csu/libc-start.c:308:16
    #10 0x42905d in _start (/home/osboxes/clamav/build-fuzz/fuzz/clamav_scanfile_fuzzer+0x42905d)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /build/glibc-eX1tMB/glibc-2.31/libio/iofwrite.c:37:3 in fwrite
==6053==ABORTING
MS: 0 ; base unit: 0000000000000000000000000000000000000000
0xa,
\x0a
artifact_prefix='./'; Test unit written to ./crash-adc83b19e793491b1c6ea0fd8b46cd9f32e592fc
Base64: Cg==

```

If I remove the "x" flag and run fuzzing with two workers, both of them use the same file:
```
$ ../fuzz/clamav_scanfile_fuzzer -workers=2
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3462281827
INFO: Loaded 1 modules   (52945 inline 8-bit counters): 52945 [0x13b6d20, 0x13c3bf1), 
INFO: Loaded 1 PC tables (52945 PCs): 52945 [0x13c3bf8,0x1492908), 
INFO: -max_len is not provided; libFuzzer will not generate inputs larger than 4096 bytes
INFO: A corpus is not provided, starting from an empty corpus
#2	INITED cov: 11 ft: 12 corp: 1/1b exec/s: 0 rss: 45Mb
	NEW_FUNC[1/112]: 0x563900 in cli_logg_setup /home/osboxes/clamav/build-fuzz/../libclamav/others_common.c:87
	NEW_FUNC[2/112]: 0x563980 in cli_ctx_tls_key_alloc /home/osboxes/clamav/build-fuzz/../libclamav/others_common.c:82
#221	NEW    cov: 458 ft: 463 corp: 2/7b lim: 6 exec/s: 0 rss: 50Mb L: 6/6 MS: 3 CopyPart-ChangeByte-InsertRepeatedBytes-
#222	NEW    cov: 458 ft: 468 corp: 3/13b lim: 6 exec/s: 0 rss: 50Mb L: 6/6 MS: 1 ChangeBinInt-
#233	NEW    cov: 459 ft: 469 corp: 4/19b lim: 6 exec/s: 0 rss: 50Mb L: 6/6 MS: 1 ChangeBit-
#234	NEW    cov: 459 ft: 471 corp: 5/25b lim: 6 exec/s: 0 rss: 51Mb L: 6/6 MS: 1 CrossOver-
	NEW_FUNC[1/4]: 0x6284f0 in td_isutf8 /home/osboxes/clamav/build-fuzz/../libclamav/textdet.c:96
	NEW_FUNC[2/4]: 0x628c70 in td_isutf16 /home/osboxes/clamav/build-fuzz/../libclamav/textdet.c:150
#240	NEW    cov: 497 ft: 561 corp: 6/31b lim: 6 exec/s: 0 rss: 51Mb L: 6/6 MS: 1 ChangeBinInt-
[...]
```

```
$ ls
tmp.scanfile-e8017b2e4807f973

```


My suggestion is to resign from random names and use PID to make filenames unique. LibFuzzer workers are running as separated processes, thus using PID will guarantee that there is no collision. Also, filename generation code should be moved from the constructor to the LLVMFuzzerTestOneInput function to make sure that it is executed for each worker process.

micahsnyder:
Thanks @mmmds  by the way with the help on this.  Your suggestions make sense.  I'm side-tracked working on some other high priority things that came up but I'll make the recommended changes when I get a chance.

micahsnyder:
@mmmds just updated with the PID recommendation. What do you think?

fuzz/clamav_scanfile_fuzzer.cpp
```diff
@@ -117,21 +101,28 @@ ClamAVState kClamAVState;
...
 
     cl_scanfile(
-        kClamAVState.tmp_file_name,
+        tmp_file_name,
         &virus_name,
         &scanned,
         kClamAVState.engine,
         &kClamAVState.scanopts);
 
+    unlink(tmp_file_name);
```
shutton, line 125: Always good to check the return value from `unlink()`, just in case you need to fail because something horrible went wrong during the test (like a path being moved out from under you).

shutton (approved):

mmmds:
> @mmmds just updated with the PID recommendation. What do you think?

Looks good to me, thanks!

fuzz/clamav_scanfile_fuzzer.cpp
micahsnyder: I don't know that there's anything we can do if it fails. @mmmds any recommendations here?

PR #437, opened 2022-01-24
GH-actions: remove CIFuzz workflow

micahsnyder:
The CIFuzz tool was observed detecting an issue in code unrelated to a
pull request. While it was relatively benign in this case, it has the
potential to accidentally disclose a vulnerability publicly without
allowing a non-disclosure period for us to publish a fixed patch
version.

I'm not comfortable with this risk, so I'm removing the workflow.


micahsnyder:
For more context, despite the oss-fuzz integration, there's no indication that an issue (and especially a security issue) would be privately reported through OSS-Fuzz if the lines in the stack traces produced by the sanitizer are unrelated to the PR diff. 

https://google.github.io/oss-fuzz/getting-started/continuous-integration/

If I'm mistaken and such a safety feature exists, or were added, then we could bring this back.

micahsnyder:
BTW @DavidKorczynski I thought you should know.  I do appreciate your help setting this up, and I'm sad we're now disabling it.

DavidKorczynski:
Thanks for the notification @micahsnyder

PR #438, opened 2022-01-25
Update vendored TomsFastMath; Support external TomsFastMath

micahsnyder:
- Update vendored TomsFastMath code to 0.13.1

    Update the vendored TomsFastMath (TFM) library to v0.13.1.

    Resolves: https://bugzilla.clamav.net/show_bug.cgi?id=11992

    I removed compatibility macro's from when libTomMath was used.
    This required removing a bunch of faux-error handling because
    the fast-math equivalent functions return void, and cannot fail.

    The previous version used had named the header "bignum_fast.h"
    instead of "tfm.h" and had customizations in that header to enable
    TFM_CHECK all the time, and also TFM_NO_ASM if __GNUC__ not defined
    or if the system isn't 64bit architecture. This update uses tfm.h
    as-is, and has CMake define TFM_CHECK and TFM_NO_ASM as needed.

    I've kept bignum.h as an interface to including tfm.h so that in
    the future we can more easily add support for system-installed
    TomsFastMath instead of the vendored one, taking inspiration from
    Debian's patch to support system-TomsFastMath.

    See: https://salsa.debian.org/clamav-team/clamav/-/blob/unstable/debian/patches/add-support-for-system-tomsfastmath.patch

- CMake: Support external TomsFastMath library (libtfm)

    Add support for compiling with external TomsFastMath library provided by
    the system instead of compiling the vendored copy into libclamav.

    The vendored source is still built directly into libclamav instead of as
    a separate library the way libmspack is done.
    The rationale is that:
    A) it's more complicated to deal with possibly compiling as static or
    dynamic, and also
    B) libmspack and libunrar are compiled separately primarily because of
    licensing concerns. TomsFastMath public domain, so that isn't a concern.

    Resolves: https://bugzilla.clamav.net/show_bug.cgi?id=12562


shutton (approved):
Full disclosure -- I did not look at every single file changed under the vendored stuff.

Issue #436, opened 2022-01-24
Logs starting with ~ are very confusing if log message is a path


dracan:
Sending a `SCAN` command to scan a specific file, and getting the following error message...

```
Mon Jan 24 08:49:25 2022 -> ~/files/b337f8a2-e03e-43f8-9774-2f0cca652d37.dat: Can't open file or directory ERROR
```

The error was just because my host's virus checking software was blocking it. However, before I realised what it was - I wasted a lot of time trying to work out why the path in the error message was wrong - ie. relative to my home (~) directory.

Digging further into the ClamD code to work out what was going on - I worked out that the codebase has a pattern of prefixing some logs with "~". This is super-confusing if the log message starts with a path.

micahsnyder:
Hi @dracan this is longstanding bug in our `logg` logging mechanism when `LogTime` is enabled.  We have a ticket somewhere for it, though I was having trouble finding it. 

The log message previous is to indicate the log-level, and is supposed to be reformatted with:
- `!` --> `ERROR: `
- `^` --> `WARNING: `
- and no prefix for the ones for Info, Verbose, or Debug levels. 

My current idea for how to fix this is to change our `logg()` function from:
```c
int logg(const char *str, ...)
```
to
```c
int logg(loglevel_t, const char *str, ...)
```
where `loglevel_t` would be a new enum containing DEBUG, VERBOSE, INFO, WARN, and ERROR.  And then replace use of the prefixes like `~` with use of these enums. 

Someone (maybe me) just needs to find time to work on it. 

See: https://github.com/Cisco-Talos/clamav/blob/main/common/output.c#L312-L317
And: https://github.com/Cisco-Talos/clamav/blob/main/common/output.c#L411-L423

dracan:
Hi @micahsnyder. Thanks for the quick and detailed reply 🙂 I guess a quicker option if removing the ~ is a lot of work, is perhaps if there aren't too many places where a file path starts at the beginning of a log entry - maybe wrap the path in those particular logs in quotes? Your suggestion is obviously a much better longer-term one - but this might be a quick fix to avoid paths that look they they originate from the user's home directory.

micahsnyder:
That's a good point @dracan , we could at least put a space or something.

mko-x:
Hi @micahsnyder & @dracan 

I would like to contribute to clamav and this issue is marked as good-first-bug. 

I was digging into the code and I was thinking about a possible quick solution.

[...]

PR #444, opened 2022-01-28
Clean irritating log output

mko-x:
Fix for #436 by adopting file logging logic.


micahsnyder:
You've since made #445 which supersedes this change. Canceling this PR.

PR #445, opened 2022-01-29
436 add log level to logg

### LINKED ISSUE #436: Logs starting with ~ are very confusing if log message is a path

mko-x:
Fix #436 .

Now logg and mprintf methods use loglevel enum to indicate loglevel instead of hardcoded prefix chars.


common/output.c
```diff
@@ -300,24 +300,15 @@ void logg_close(void)
...
- *  $	  no	   mprintf     no	yes   LOG_DEBUG
- *  none  yes	   mprintf     yes	yes   LOG_INFO
+ * NAME     OLD        EXPLAIN
+ * INFO     none or ~   normal
+ * INFO_NF  #           normal, no foreground
+ * DEBUG    *           debug, verbose
+ * DEBUG_NV $           debug, non-verbose
+ * WARNING  ^           warning
+ * ERROR    !           ERROR
  */
```
micahsnyder, line 310: The "OLD" column probably isn't needed but it sure helps for this review. 

The `_NF` and `_NV` suffixes aren't intuitive to me but the "EXPLAIN" column helps.  I think "`not foreground (logfile and syslog only)`" would be a little better than just "`no foreground`" to explain the `_NF` variant. 

Please also update the copy in `output.h`. You may even consider moving this comment block to `output.h` just above the `logg()` function prototype so it's more obvious in the future that the comment exists for both the function and the enum.  I do like having a copy for the function for easy reference when i highlight `logg` in my editor.  Maybe the function-version should also explain the return value.

common/output.c
```diff
@@ -300,24 +300,15 @@ void logg_close(void)
...
-int logg(const char *str, ...)
+int logg(int loglevel, const char *str, ...)
```
micahsnyder, line 311: The loglevel type should be of type `loglevel_t`

common/output.h
```diff
@@ -39,11 +39,26 @@ int mdprintf(int desc, const char *str, ...) __attribute__((format(printf, 2, 3)
...
+ * ERROR    !           ERROR
+ */
+enum loglevel_t{
+    INFO, 
+    INFO_NF, 
+    DEBUG, 
+    DEBUG_NV, 
+    WARNING, 
+    ERROR
+};
```
micahsnyder, line 59: The CI failed on Windows because of a namespace collision with "ERROR" being defined in the Windows headers.  I'm a little surprised actually that we didn't run into that with any of the others. 

I guess we should prefix them with something recognizable, like "`LOGG_`".

It would also be nice to make this a typedef so we can reference the `loglevel_t` enum without having to use the worth `enum`.  I'm thinking:

```c
typedef enum loglevel {
    LOGG_INFO,
    LOGG_INFO_NF,
    LOGG_DEBUG,
    LOGG_DEBUG_NV,
    LOGG_WARNING,
    LOGG_ERROR
} loglevel_t;
```

Sorry about the hassle renaming everything, again.

common/output.h
```diff
@@ -55,11 +70,7 @@ extern short logg_syslog;
 int logg_facility(const char *name);
 #endif
 
-#ifdef __GNUC__
-void mprintf(const char *str, ...) __attribute__((format(printf, 1, 2)));
-#else
-void mprintf(const char *str, ...);
-#endif
+void mprintf(int loglevel, const char *str, ...);
```
micahsnyder, line 73: I hadn't considered before that we lose the `__attribute__((format(printf, 1, 2))` format string checks, but honestly it's worth it to me to fix this bug and get rid of the hacky `buff[0]` checks and un-intuitive log-level special characters.  👍

micahsnyder (changes requested):
This is awesome @mko-x. Thank you! 

I like how you split up the commits.

I see that clang-format has [some complaints](https://github.com/Cisco-Talos/clamav/runs/5030642455?check_suite_focus=true) about the formatting in `clamd_others.c`, `server-th.c` and a handful of files under `common`.  

*After you make the requested changes*, can I trouble you to run the `./clam-format` script with `clang-format` installed. Either clang-format version 10 or version 11 should work. Instead of `./clam-format` you can use `clang-format -i <filename>` for each file to fix them a la carte.

common/output.c
mko-x: I updated and moved the comment block to `output.h`, please review the comments again to determine if I got you right

common/output.h
mko-x: Well, it's not that hard job to rename nowadays with vs-code. Changed the enum to a typedef

common/output.h
mko-x: Oh well, I think we don't need to lose them. We can shift the args, then it'll work even with new loglevel (:

mko-x: See [4448c09](https://github.com/Cisco-Talos/clamav/pull/445/commits/4448c092c83d1297706e12a6c029677d921e4ff1)

mko-x:
Hi @micahsnyder !

Thank you very much for your review 🙏🏼 

I fixed anything you mentioned, I guess.

I noticed that we can keep attribute format string checks by shifting the arg indices by 1. 👯 

Did a run with clang-format, should work probably.

micahsnyder:
@mko-x  sorry for the bad news, I think you may have to redo the commits from the start.  Something in your follow-up search & replace actions broke the build entirely. 

I also observed this error when compiling clamav-milter on ubuntu that is one more place `logg()` was hiding that was missed earlier:
```
/home/runner/work/clamav/clamav/clamav-milter/netcode.c: In function ‘nc_recv’:
/home/runner/work/clamav/clamav/clamav-milter/netcode.c:273:28: error: incompatible type for argument 1 of ‘logg’
  273 |             strerror_print("!recv failed after successful select");
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                            |
      |                            char *
/home/runner/work/clamav/clamav/clamav-milter/netcode.c:58:34: note: in definition of macro ‘strerror_print’
   58 | #define strerror_print(msg) logg(msg ": %s\n", cli_strerror(errno, er, sizeof(er)))
      |                                  ^~~
In file included from /home/runner/work/clamav/clamav/clamav-milter/netcode.c:53:
/home/runner/work/clamav/clamav/common/output.h:74:21: note: expected ‘loglevel_t’ {aka ‘enum loglevel’} but argument is of type ‘char *’
   74 | int logg(loglevel_t loglevel, const char *str, ...) __attribute__((format(printf, 2, 3)));
      |          ~~~~~~~~~~~^~~~~~~~
make[2]: *** [clamav-milter/CMakeFiles/clamav-milter.dir/build.make:118: clamav-milter/CMakeFiles/clamav-milter.dir/netcode.c.o] Error 1
```

mko-x:
Hey @micahsnyder !

I don't know why the tests on Windows fail, maybe the timeout is too short for loading all. I see only the listing of envars in the logs and then the timeout. Shouldn't be related to my changes - should it?

Search and replace didn't break the build, I guess. It seems to be by the clang-format config. I fixed it to fit to work with the build system again.

I fixed the hidden logg calls in milter and added the loglevel to the calls.

mko-x:
ctest runs perfectly for me:

```bash
clamav/build on  436_AddLogLevelToLogg took 2m 0s 
➜ ctest
Test project /Users/muka/Documents/repos/clamav/build
    Start 1: libclamav
1/6 Test #1: libclamav ........................   Passed   10.29 sec
    Start 2: libclamav_rust
2/6 Test #2: libclamav_rust ...................   Passed   75.59 sec
    Start 3: clamscan
3/6 Test #3: clamscan .........................   Passed    4.92 sec
    Start 4: clamd
4/6 Test #4: clamd ............................   Passed   14.83 sec
    Start 5: freshclam
5/6 Test #5: freshclam ........................   Passed  116.58 sec
    Start 6: sigtool
6/6 Test #6: sigtool ..........................   Passed    0.82 sec

100% tests passed, 0 tests failed out of 6

Total Test time (real) = 223.07 sec
```

mko-x:
I reset the .clang-format style config file now which seems to be used by the CI jobs.

I was wondering why it was able to run the check for `libclamav`but not for other targets.

Ubuntu build is working fine now. 

No idea why the windows tests run into timeouts.

micahsnyder:
It was a long read but it looks great to me all around!

I'm sorry I procrastinated reviewing this the past week. And there is now a merge conflict with `bignum_fast.h` that needs to be resolved before I can merge it - we deleted `bignum_fast.h` in a recent commit.  By the way, since it's a bunch of commits now, I'm planning to squash merge it as one commit.  I hope you don't mind.

Can you resolve the conflict?  I'm prepared to merge it afterwards.

micahsnyder (approved):

PR #452, opened 2022-02-03
fix sign warnings

bigio:
remove most warnings spotted by gcc -Wsign-compare

### DIFF
--- a/libclamav/yara_arena.c
+++ b/libclamav/yara_arena.c
@@ -180,7 +180,7 @@ int _yr_arena_make_relocatable(
   base_offset = (uint8_t*) base - page->address;
   offset = va_arg(offsets, size_t);
 
-  while (offset != -1)
+  while (offset != EOL)
   {
     assert(base_offset + offset <= page->used - sizeof(int64_t));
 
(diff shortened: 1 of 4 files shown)


libclamav/yara_arena.c
```diff
@@ -178,9 +178,9 @@ int _yr_arena_make_relocatable(
   assert(page != NULL);
 
   base_offset = (uint8_t*) base - page->address;
-  offset = va_arg(offsets, size_t);
+  offset = va_arg(offsets, ssize_t);
```
micahsnyder, line 181: I don't think this change is needed. 

Everything else looks good to me now.

micahsnyder:
It appears this was updated incorrectly, pulling in changes from main rather than rebasing with main.

micahsnyder:
@bigio I just force-pushed a fix that corrects the commit history and squashes your two commits into one.  Provided everything goes fine through the test pipeline (and I don't see why it wouldn't), I'll merge right away.

micahsnyder:
Thanks again for the work you put into this @bigio

PR #454, opened 2022-02-05
Windows: Fix crash in Debug builds when scanning some PE files

micahsnyder:
The endianness conversion when reading the PE image import descriptor is
making the change in-place in the fmap.
On Windows, the fmap is read-only and so in Debug builds that's causing
a crash.

This uses a buffer on the stack and copies each image import descriptor
before doing the conversions and then processing each.

### DIFF
--- a/libclamav/pe.c
+++ b/libclamav/pe.c
@@ -2411,125 +2411,146 @@ static inline int hash_impfns(cli_ctx *ctx, void **hashctx, uint32_t *impsz, str
 
 static cl_error_t hash_imptbl(cli_ctx *ctx, unsigned char **digest, uint32_t *impsz, int *genhash, struct cli_exe_info *peinfo)
 {
-    struct pe_image_import_descriptor *image;
+    cl_error_t status = CL_SUCCESS;
+    cl_error_t ret;
+    struct pe_image_import_descriptor image;
+    const struct pe_image_import_descriptor *impdes;
     fmap_t *map = ctx->fmap;
     size_t left, fsize = map->len;
     uint32_t impoff, offset;
-    const char *impdes, *buffer;
+    const char *buffer;
     void *hashctx[CLI_HASH_AVAIL_TYPES];
     enum CLI_HASH_TYPE type;
-    int nimps      = 0;
-    cl_error_t ret = CL_SUCCESS;
+    int nimps = 0;
     unsigned int err;
-    int first = 1;
+    int first          = 1;
+    bool needed_impoff = false;
 
     /* If the PE doesn't have an import table then skip it. This is an
      * uncommon case but can happen. */
     if (peinfo->dirs[1].VirtualAddress == 0 || peinfo->dirs[1].Size == 0) {
         cli_dbgmsg("scan_pe: import table data dir does not exist (skipping .imp scanning)\n");
-        return CL_SUCCESS;
+        goto done;
     }
 
     // TODO Add EC32 wrappers
     impoff = cli_rawaddr(peinfo->dirs[1].VirtualAddress, peinfo->sections, peinfo->nsections, &err, fsize, peinfo->hdr_size);
     if (err || impoff + peinfo->dirs[1].Size > fsize) {
         cli_dbgmsg("scan_pe: invalid rva for import table data\n");
-        return CL_SUCCESS;
+        goto done;
     }
 
     // TODO Add EC32 wrapper
-    impdes = fmap_need_off(map, impoff, peinfo->dirs[1].Size);
+    impdes = (const struct pe_image_import_descriptor *)fmap_need_off(map, impoff, peinfo->dirs[1].Size);
     if (impdes == NULL) {
         cli_dbgmsg("scan_pe: failed to acquire fmap buffer\n");
-        return CL_EREAD;
+        status = CL_EREAD;
+        goto done;
     }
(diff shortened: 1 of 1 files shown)


libclamav/pe.c
```diff
@@ -2468,28 +2468,30 @@ static cl_error_t hash_imptbl(cli_ctx *ctx, unsigned char **digest, uint32_t *im
...
         /* DLL name acquisition */
-        offset = cli_rawaddr(image->Name, peinfo->sections, peinfo->nsections, &err, fsize, peinfo->hdr_size);
+        offset = cli_rawaddr(image.Name, peinfo->sections, peinfo->nsections, &err, fsize, peinfo->hdr_size);
         if (err || offset > fsize) {
             cli_dbgmsg("scan_pe: invalid rva for dll name\n");
             /* TODO: ignore or return? */
             /*
-              image++;
-              continue;
+            impdes += sizeof(struct pe_image_import_descriptor);
```
shutton, line 2492: It's not obvious (at first glance) what's going on here (i.e., what's the intended change to this pointer?).  It was originally defined as a `char *`, so this appears to be moving the pointer along a contiguous set of those structures.

Might be clearer if defined as `struct pe_image_import_descriptor *`, which would allow you to just do `impdes++` without specifying the size.

I also don't see the bounds check, but I'm not familiar with this structure.

libclamav/pe.c
```diff
@@ -2516,13 +2518,14 @@ static cl_error_t hash_imptbl(cli_ctx *ctx, unsigned char **digest, uint32_t *im
...
         /* DLL function handling - inline function */
-        ret = hash_impfns(ctx, hashctx, impsz, image, dllname, peinfo, &first);
+        ret = hash_impfns(ctx, hashctx, impsz, &image, dllname, peinfo, &first);
         free(dllname);
         dllname = NULL;
         if (ret != CL_SUCCESS)
             goto hash_imptbl_end;
 
-        image++;
+        impdes += sizeof(struct pe_image_import_descriptor);
```
shutton, line 2527: see above

libclamav/pe.c
```diff
@@ -2468,28 +2468,30 @@ static cl_error_t hash_imptbl(cli_ctx *ctx, unsigned char **digest, uint32_t *im
         }
     }
 
-    image = (struct pe_image_import_descriptor *)impdes;
-    while (left > sizeof(struct pe_image_import_descriptor) && image->Name != 0 && nimps < PE_MAXIMPORTS) {
+    memcpy(&image, impdes, sizeof(struct pe_image_import_descriptor));
```
shutton, line 2471: `impdes` is declared `const char *` above, but received from `fmap_need_off()`, which returns `void *`.  Recommend aligning these types (`struct pe_image_import_descriptor *`)

shutton (changes requested):

libclamav/pe.c
micahsnyder: Aligning the types for the `image` and `impdes` pointers sounds good to me.

micahsnyder:
@shutton ready for another review pass

libclamav/pe.c
```diff
@@ -2411,125 +2411,146 @@ static inline int hash_impfns(cli_ctx *ctx, void **hashctx, uint32_t *impsz, str
...
-        return CL_SUCCESS;
+        goto done;
     }
 
     // TODO Add EC32 wrappers
     impoff = cli_rawaddr(peinfo->dirs[1].VirtualAddress, peinfo->sections, peinfo->nsections, &err, fsize, peinfo->hdr_size);
     if (err || impoff + peinfo->dirs[1].Size > fsize) {
         cli_dbgmsg("scan_pe: invalid rva for import table data\n");
-        return CL_SUCCESS;
+        goto done;
```
shutton, line 2440: Does this call for an update to `status`?

shutton (changes requested):
Only one minor thing to consider.

libclamav/pe.c
```diff
@@ -2411,125 +2411,146 @@ static inline int hash_impfns(cli_ctx *ctx, void **hashctx, uint32_t *impsz, str
 
 static cl_error_t hash_imptbl(cli_ctx *ctx, unsigned char **digest, uint32_t *impsz, int *genhash, struct cli_exe_info *peinfo)
 {
-    struct pe_image_import_descriptor *image;
+    cl_error_t status = CL_SUCCESS;
+    cl_error_t ret;
+    struct pe_image_import_descriptor image;
```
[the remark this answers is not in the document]
micahsnyder: ` = {0} ` is safe to use in all the platforms we test with.  Incidentally, apparently C23 will introduce `= {}` for initializing _anything_ to zero, in the same way that memset would.  I learned that ` = {0}` doesn't initialize the padding in structs, but `  = {}` will.  🤷 

Good idea to initialize it though.

libclamav/pe.c
micahsnyder: probably!! nice catch.
Speaking of which, I used to follow a practice of setting the most common failure type for the `status` initialization and then only setting it if a different status was needed... but I've changed my ways and prefer now to be explicit every time I `goto done;` to prevent a bug if someone changes the initialization default.  Please call me out on it if I revert to my old ways.

shutton (approved):

PR #461, opened 2022-02-10
Validate the length of the crt subject before memcpying it

ragusaa:


libclamav/readdb.c
```diff
@@ -2943,6 +2943,30 @@ static int cli_loadcdb(FILE *fs, struct cl_engine *engine, unsigned int *signo,
     return CL_SUCCESS;
 }
 
+static cl_error_t setHash(const char *const token, uint8_t hashDest[SHA1_HASH_SIZE],
```
shutton, line 2946: `setHash` doesn't seem to match the naming convention in the rest of the file (i.e., camelCase vs. snake_case).  Is that intentional?

shutton: A side curiosity -- what's the standard for the meaning of a "hash" within the ClamAV codebase?  Depending on context, it either means a dictionary or a cryptographic digest (or even just a `#`).  IMHO, it's usually a little clearer to use a term like "sha1", "sha2_256", "sha3_256", *etc.* for digests.

shutton (approved):

libclamav/readdb.c
micahsnyder: True we generally do snake_case, not camelCase.  
It would also be helpful to give a descriptive function comment.

libclamav/readdb.c
```diff
@@ -2943,6 +2943,30 @@ static int cli_loadcdb(FILE *fs, struct cl_engine *engine, unsigned int *signo,
...
+    }
+
+    if (cli_hex2str_to(token, (char *)hash, strlen(token) < 0)) {
+        cli_errmsg("cli_loadcrt: line %u: Cannot convert %s to binary string\n", (unsigned int)line, varname);
+        ret = CL_EMALFDB;
+        goto end;
+    }
+    memcpy(hashDest, hash, SHA1_HASH_SIZE);
+
+end:
```
micahsnyder, line 2966: We should use the `done:` label instead of `end:`, so our new alloc-macros will work with it, and for convention's sake.

libclamav/readdb.c
```diff
@@ -3116,16 +3134,12 @@ static int cli_loadcrt(FILE *fs, struct cl_engine *engine, struct cli_dbio *dbio
...
         ca.hashtype = CLI_HASHTYPE_ANY;
         crtmgr_add(&(engine->cmgr), &ca);
-        free(subject);
-        free(pubkey);
-        subject = pubkey = NULL;
+
+        FREE(pubkey);
     }
 
 end:
```
micahsnyder, line 3141: It's a bit outside of the scope, but I would recommend changing this label to `done:` as well while cleaning up this function.

micahsnyder (changes requested):

libclamav/readdb.c
```diff
@@ -2943,6 +2943,30 @@ static int cli_loadcdb(FILE *fs, struct cl_engine *engine, unsigned int *signo,
     return CL_SUCCESS;
 }
 
+static cl_error_t setHash(const char *const token, uint8_t hashDest[SHA1_HASH_SIZE],
+                          const char *const varname, uint32_t line)
```
micahsnyder, line 2947: Are these supposed to be `const char *const` or just `const char*`?   I'm thinking the latter.

Also, github's static analysis tool is claiming an issue with the `cli_hex2str_to` call: https://lgtm.com/projects/g/Cisco-Talos/clamav/rev/pr-bc58cc28d41743a051684a375e8147c902f01749  I'm wondering if it's because of the `const char *const`.  What do you think @ragusaa?

libclamav/readdb.c
ragusaa: I agree, I changed it to set_sha1.

libclamav/readdb.c
ragusaa: The lgtm issue was due to having the ')' in the wrong place.  I am really surprised this did not produce a warning, especially with -Wall -Wextra.

As far as the const char * const, that sets the pointer itself, as well as the data, const.

micahsnyder:
rebased & fixed merge conflict. added 2 commits:
- update generated sys.rs file (issue from a previous merge, my bad)
- increased test timeout to 10min after observing the libclamav_valgrind test fail while testing this PR because it exceeded 200sec.  Nothing wrong with this PR ... just my machine was busy and 200sec wasn't enough.

micahsnyder (approved):

PR #464, opened 2022-02-10
Corrected buffer size check

ragusaa:


shutton (approved):

micahsnyder:
I couldn't repro the issue using just a local ASAN build with ENABLE_FUZZ=ON, but I was able to reproduce it using the oss-fuzz framework.  I confirmed this fix works.

My only complaint is the commit message isn't helpful.  I'll massage it with a squash&merge.

PR #466, opened 2022-02-12
Image fuzzy hash: new logical sub-signature feature

micahsnyder:
Add a new logical signature subsignature type for matching on images with image fuzzy hashes.
The hash algorithm used is a Rust implementation of the `phash()` function from [the `imagehash` Python package](https://github.com/JohannesBuchner/imagehash). That is a BSD-2 licensed clean-room approximation of the pHash DCT image fuzzy hashing algorithm. 

Fuzzy hash matching is only performed for supported image types.

Image fuzzy hash subsigantures follow this format:

    fuzzy_img#<hash>#<dist>

Note: In this initial signature implementation, the hamming distance (dist) option is ignored and only exact fuzzy hash matches will alert.

Example signature:
```
cisco.png;Engine:140-255,Target:0;0;fuzzy_img#9463944473afd82f#0
```

While `clamscan --debug` can be used to find the fuzzy hash, you can also get it from the "ImageFuzzyHash" field in the `--gen-json` metadata. Not in this PR; We should probably add an option to `sigtool` to generate it. 

Also: Fixed an issue where the signature name (`virname`) is being allocated and stored for every subsignature or even ever sub-pattern in an AC-pattern (i.e. NDB sig or LDB subsig) containing a `{n-m}` or `*` wildcard. This fix is only for LDB subsigs though. NDB signatures are still allocaing one virname per sub-pattern.

This fix was required because I needed a place to store the virname with fuzzy-hash subsignatures. Storing it in the fuzzy-hash subsig metadatathe way AC-pattern, PCRE, and BComp subsigs were doing it wouldn't work because it would cross the C-Rust FFI boundary and giving pointers to Rust allocated stuff is dicey. Not to mention native Rust strings are different than C strings. Anyways, the correct thing to do was to store the virname with the actual logical signature.

To-do, but not in this PR: Keep track of NDB signatures in the same way and store the virname for NDB sigs there instead of in AC-patterns so that we can get rid of the virname field in the AC-pattern struct.


micahsnyder:
Note, the first 3 commits are supporting commits but don't actually add this feature.  The first of which is broken out into PR https://github.com/Cisco-Talos/clamav/pull/453.  I will rebase this PR once that one merges.  I wouldn't mind splitting out the next two into a separate PR, as well, to make review easier.

libclamav/matcher.h
```diff
@@ -164,11 +166,17 @@ struct cli_matcher {
     uint32_t bcomp_metas;
     struct cli_bcomp_meta **bcomp_metatable;
 
+    /* Fuzzy Image Hash */
+    uint32_t fuzzy_hashes;
```
micahsnyder, line 170: Not  needed?

libclamav/matcher.c
```diff
@@ -191,10 +192,27 @@ static inline cl_error_t matcher_run(const struct cli_matcher *root,
         }
     }
 
+    switch (ftype) {
+        case CL_TYPE_GIF:
+        case CL_TYPE_TIFF:
+        case CL_TYPE_JPEG:
+        case CL_TYPE_PNG: {
```
micahsnyder, line 199: Should we add CL_TYPE_GRAPHICS here? BMP? JPEG-2000, etc?

micahsnyder: (else back out creating  hashes for these.)

libclamav_rust/Cargo.toml
```diff
@@ -8,10 +8,12 @@ version = "0.0.1"
 flate2 = "1"
 hex = "0.4"
 libc = "0.2"
-log = {version = "0.4", features = ["std"]}
+log = { version = "0.4", features = ["std"] }
 sha2 = "0.10"
 tempfile = "3"
 thiserror = "1"
+image = "0.23"
+img_hash = "*"
```
micahsnyder: remove!

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,333 @@
...
+            }
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+pub struct FuzzyHashMeta {
+    lsigid: u32,
+    subsigid: u32,
+    _hamming_distance: u32,
```
micahsnyder, line 59: Either remove or feature it out.  Ask Scott.

micahsnyder:
Just rebased after merging the prep-PR, and squashed the squashme-commits for the fixes that came up in our team overview meeting. Ready for review now.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+        Some(x) => x,
+        None => return Err(FuzzyHashError::Format),
+    };
+
+    let hash = match hexsig_split.next() {
+        Some(x) => x,
+        None => return Err(FuzzyHashError::Format),
+    };
+
+    let distance: u32 = match hexsig_split.next() {
```
shutton, line 218: `: u32` should be implicit here. It's the wrapped return type from `u32::from_str_radix()`.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+                return Err(FuzzyHashError::FormatHammingDistance(x.to_string()));
+            }
+        },
+        None => 0,
+    };
+
+    // TODO: Support non-zero distance
+    if distance != 0 {
+        error!(
+            "Non-zero hamming distances for image fuzzy hashes are not supported in this version."
```
shutton, line 231: Perhaps a FLevel thing?

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+#[derive(Error, Debug)]
+pub enum FuzzyHashError {
+    #[error("Invalid format")]
+    Format,
+
+    #[error("Unknown algorithm: {0}")]
+    UnknownAlgorithm(String),
+
+    #[error("Invalid hamming distance: {0}")]
+    FormatHammingDistance(String),
```
shutton, line 31: This should be the numeric type for hamming distance.  If the hamming distance is not a valid number, that would be a different error.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+    subsig_id: u32,
+) -> Result<(), FuzzyHashError> {
+    let mut hexsig_split = hexsig.split('#');
+
+    let algorithm = match hexsig_split.next() {
+        Some(x) => x,
+        None => return Err(FuzzyHashError::Format),
+    };
+
+    let hash = match hexsig_split.next() {
```
shutton, line 213: You can catch two errors at once here if you use `hex::decode_to_slice`.  See `util.rs` in the signature validator code for an example.  It's a touch more efficient since it doesn't require heap allocation.

shutton (changes requested):

libclamav/matcher-pcre.c
```diff
@@ -735,7 +725,7 @@ cl_error_t cli_pcre_scanbuf(const unsigned char *buffer, uint32_t length, const
                             ret = CL_EMEM;
                             break;
                         }
-                        newres->virname    = pm->virname;
+                        newres->virname    = "test";
```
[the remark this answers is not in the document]
micahsnyder: Yup!  It's used for testing signatures with sigtool or with the matcher unit tests for signature support.  I had to pick some static name because the actual signature name isn't available in this situation. 
In the future I'd like to change how sigtool works for testing signatures -- that is to test whole LDB signatures instead of testing each subsignature a la carte.  Then we could remove this condition entirely.  But I'm not sure how much work that would be and didn't have time to do it now.

libclamav/matcher.c
```diff
@@ -191,10 +192,28 @@ static inline cl_error_t matcher_run(const struct cli_matcher *root,
...
 
+    switch (ftype) {
+        case CL_TYPE_GIF:
+        case CL_TYPE_TIFF:
+        case CL_TYPE_JPEG:
+        case CL_TYPE_PNG:
+        case CL_TYPE_GRAPHICS: {
+            ret = fuzzy_hash_check(root, mdata, ctx);
+            if (ret != CL_CLEAN) {
+                if (ret > CL_TYPENO && acmode & AC_SCAN_VIR)
```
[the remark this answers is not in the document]
micahsnyder: Me too.  I'll clean it up.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+use transpose::transpose;
+
+use std::collections::HashMap;
+use std::convert::TryInto;
+use std::ffi::CStr;
+use std::mem::ManuallyDrop;
+use std::os::raw::c_char;
+use std::slice;
+
+use log::*;
```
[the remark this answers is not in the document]
micahsnyder: Oh right, log is external.  I think it put it below thinking it was from our logging module, by mistake.

I'm not sure if rust has a convention here.  In Python, you generally do imports in the following order, separated by a blank line:
```
<standard library imports>

<external library imports>

<internal library imports>
```
I like the convention, though I guess in this case it's backwards. Maybe I'll move the external imports down.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+    let hashmap: HashMap<FuzzyHash, Vec<FuzzyHashMeta>> = HashMap::new();
+    let hashmap = Box::new(hashmap);
+    Box::into_raw(hashmap) as sys::hashmap_ptr_t
+}
+
+/// Free the hashmap
+#[no_mangle]
+pub extern "C" fn fuzzy_hash_free_hashmap(fuzzy_hashmap: sys::hashmap_ptr_t) {
+    if !fuzzy_hashmap.is_null() {
+        let hashmap =
```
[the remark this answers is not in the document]
micahsnyder: It felt good to do it explicitly, but yeah could do that way too.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+pub extern "C" fn fuzzy_hash_init_hashmap() -> sys::hashmap_ptr_t {
+    let hashmap: HashMap<FuzzyHash, Vec<FuzzyHashMeta>> = HashMap::new();
+    let hashmap = Box::new(hashmap);
+    Box::into_raw(hashmap) as sys::hashmap_ptr_t
+}
+
+/// Free the hashmap
+#[no_mangle]
+pub extern "C" fn fuzzy_hash_free_hashmap(fuzzy_hashmap: sys::hashmap_ptr_t) {
+    if !fuzzy_hashmap.is_null() {
```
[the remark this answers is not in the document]
micahsnyder: In C if you call `free()` on a null pointer, it doesn't do anything. It's the same sort of thing. 

Despite there being no technical need for it, convention in libclamav is to do an `if(NULL != pointer) {` before a `free(pointer); }`.  So I guess I could warn about it here, as per our convention it shouldn't happen.  🤷

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+        ManuallyDrop::new(Box::from_raw(
+            p as *mut HashMap<FuzzyHash, Vec<FuzzyHashMeta>>,
+        ))
+    };
+
+    let result: sys::cl_error_t =
+        match fuzzy_hash_load_subsignature(&mut hashmap, hexsig, lsig_id, subsig_id) {
+            Ok(_) => sys::cl_error_t_CL_SUCCESS,
+            Err(error) => {
+                error!("Error when loading fuzzy hash logical subsignature:");
```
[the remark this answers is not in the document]
micahsnyder: No idea 😅 I'll fix it.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+                    }
+                    FuzzyHashError::FormatHammingDistance(dist) => {
+                        error!("Invalid hamming distance: {}", dist);
+                        sys::cl_error_t_CL_EFORMAT
+                    }
+                    FuzzyHashError::FormatHashBytes(hash) => {
+                        error!("Invalid bytes for hash: {}", hash);
+                        sys::cl_error_t_CL_EFORMAT
+                    }
+                    _ => {
```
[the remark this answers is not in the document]
micahsnyder: I really don't want to handle all possible error types here.  I think I'm fine with reporting any failure here with `CL_EFORMAT` and making use of the `Display` trait, as you suggest.

libclamav_rust/src/fuzzy_hash.rs
micahsnyder: I could mention the current FLevel in the log, but there's no way of knowing now which future FLevel will support it.  Telling someone to verify that the FLevel is set correctly implies that some other FLevel will work, and I while I'd like to implement the hamming distance feature in the future, I can't promise if/when that will happen.  I'm inclined to keep this message as-is.

libclamav_rust/src/fuzzy_hash.rs
shutton: `FormatHammingDistance(<some integer type>)` is what I meant there, if it wasn't clear.

libclamav_rust/src/fuzzy_hash.rs
micahsnyder: Discussed out-of-band. I updated the relevant code to use `hex::decode_to_slice` (below)

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -0,0 +1,544 @@
...
+#[derive(Debug, Copy, Clone)]
+pub struct FuzzyHashMeta {
+    lsigid: u32,
+    subsigid: u32,
+    _hamming_distance: u32,
+}
+
+/// Initialize the hashmap
+#[no_mangle]
+pub extern "C" fn fuzzy_hash_init_hashmap() -> sys::hashmap_ptr_t {
```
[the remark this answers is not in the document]
micahsnyder: > A general comment on the design:
> 
> You're giving Rust HashMap objects directly back to the caller. However, you _could_ be giving back a `struct` that wraps the HashMap. Then, functions like `fuzzy_hash_load_subsignature` would drop into an `impl` for that structure and take `&self`. I think it would end up being a lot cleaner, and gives you a little more flexibility later.

I think I've made all of the suggested changes except this one. We could talk more about what you mean. I'm otherwise ready for re-review.

libclamav/matcher.h
```diff
@@ -99,6 +99,8 @@ struct cli_lsig_tdb {
     uint32_t       *macro_ptids;
 #ifdef USE_MPOOL
     mpool_t        *mempool;
+#else
+    void           *mempool;
```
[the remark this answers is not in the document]
shutton: I see the `_padding_mempool` below, but I think this one got missed.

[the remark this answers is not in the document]
micahsnyder: Whoops. Fixed.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -36,7 +36,7 @@ use rustdct::DctPlanner;
 use thiserror::Error;
 use transpose::transpose;
 
-use crate::{frs_error::FRSError, frs_result, sys, validate_str_param};
+use crate::{frs_call, frs_error, frs_error::FRSError, sys, validate_str_param};
```
micahsnyder, line 39: I'm a little surprised that Rust isn't confused between `crate::frs_error` (the macro) and `crate::frs_error` the module.  I worry we should rename the macro or the module to make it less confusing.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -110,212 +118,205 @@ pub struct FuzzyHashMeta {
...
-    };
+    let hashmap = ManuallyDrop::new(Box::from_raw(fuzzy_hashmap as *mut FuzzyHashMap));
 
     debug!(
         "Checking image fuzzy hash '{}' for signature match",
         hex::encode(hash_bytes)
     );
 
-    if let Some(meta_vec) = fuzzy_hash_check(&hashmap, hash_bytes) {
+    if let Some(meta_vec) = hashmap.check(hash_bytes) {
```
micahsnyder, line 156: Oooh I like this.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -110,212 +118,205 @@ pub struct FuzzyHashMeta {
...
-        }
-        _ => {
-            error!("Unknown fuzzy hash algorithm: {}", algorithm);
-            Err(FuzzyHashError::UnknownAlgorithm(algorithm.to_string()))
-        }
-    }
+    frs_call!(
+        err = err,
+        hashmap.load_subsignature(hexsig, lsig_id, subsig_id)
+    )
```
micahsnyder, line 186: Having the macro params separate from the function params does feel much better. I like it.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -110,212 +118,205 @@ pub struct FuzzyHashMeta {
...
 }
 
 /// C interface for fuzzy_hash_calculate_image().
 /// Handles all the unsafe ffi stuff.
+///
+/// # Safety
+///
+/// `file_bytes` and `hash_out` must not be NULL
```
micahsnyder, line 194: More like "must be valid pointers" since we're doing the NULL check.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -110,212 +118,205 @@ pub struct FuzzyHashMeta {
...
+        return frs_error!(
+            err = err,
+            FuzzyHashError::InvalidParameter(format!(
                 "hash_bytes output parameter too small to hold the hash: {} < {}",
                 hash_out_len,
                 hash_bytes.len()
-            )));
-        return unsafe { frs_result!(result_in = hash_result, err = err) };
+            ))
+        );
```
micahsnyder, line 227: This also feels much better not having to construct a Result and then consume it.  Very nice.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -466,10 +467,9 @@ pub fn fuzzy_hash_calculate_image(buffer: &[u8]) -> Result<Vec<u8>, FuzzyHashErr
         // Grab the first 8 rows.
         .take(8)
         // But only take the first 8 elements (columns) from each row.
-        .map(|chunk| chunk.chunks(8).take(1))
+        .flat_map(|chunk| chunk.chunks(8).take(1))
```
micahsnyder, line 470: Neat

micahsnyder:
Force-pushed to squash ffi_utils cleanup work into one commit. 

Test pipelines are all looking good.  I"m going to merge this.

PR #468, opened 2022-02-14
Application to scan local files on remote clamd

steveking13:
No prerequisites, except for a python3 interpreter

### DIFF
--- a/clamdstream/clamdstream
+++ b/clamdstream/clamdstream
@@ -1,0 +1,158 @@
+#! /usr/bin/env python3
+
+helptext = '''Synopsis:
+    Helper function that drives clamd's remote TCP INSTREAM
+    facility by streaming the file over TCP
+
+    This allows a machine with no clamav installation to scan files using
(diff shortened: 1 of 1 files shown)


micahsnyder:
Hi @steveking13 this is a nice script you have shared.

You can use `clamdscan --stream` to do the same, provided you configure a `clamd.conf` file that tells `clamdscan` how to connect.  For this reason, and because your `clamdstream` program isn't tied into the ClamAV build/install process, It's not the sort of thing we would want to maintain or store in repository. 

If you have this as a separate github repo, it would be great to add this tool to https://docs.clamav.net/manual/Installing/Community-projects.html.  You can put in PR to our clamav-documentation source here to add it: https://github.com/Cisco-Talos/clamav-documentation/blob/main/src/manual/Installing/Community-projects.md 

You may also consider publishing your tool on Python's PyPI package server so users can simple `pip install` it. 

Cheers,
Micah

PR #469, opened 2022-02-15
Modified read condition in javascript normalizer

ragusaa:
Modified conditional statement to test for an out of bounds index before
reading from that index.


micahsnyder (approved):

micahsnyder:
Thanks Michał Dardas for independently identifying and reporting this bug, which had also been found by oss-fuzz. 
And thanks Michał for providing a suggested fix that happened to be the exact same as the one Andy implemented!

PR #473, opened 2022-02-16
Clam 1590 phash logical subsignatures cleanup prep

micahsnyder:
Some cleanup for readdb and matcher-ac that I did while trying to understand this code.  would be good to review and merge before https://github.com/Cisco-Talos/clamav/pull/466

This PR shouldn't break anything and doesn't _exactly_ fix anything, except in that it makes it that NDB and other ac-matching signatures (FTM?) don't call the same function we use for handling an LDB subsig.


shutton (approved):
This code could definitely benefit from an overhaul (e.g., minimizing interactions with `root`, breaking things up a bit, etc.), but that's out-of-scope for this feature.

PR #476, opened 2022-02-18
CMake: Fix static build with unrar disabled

micahsnyder:
Fix test build issues when unrar is disabled and the static build is enabled.

This resolves a build/test failure in the custom-0.105 test pipeline.


unit_tests/freshclam_test.py
```diff
@@ -521,14 +558,15 @@ def test_freshclam_07_cdiff_partial_minus_2(self):
...
-        self.verify_output(output.out, expected=expected_results, unexpected=unexpected_results)
+        expected_stdout = ['test.cvd updated']
+        expected_stderr = ['Incremental update failed, trying to download test.cvd']
+        unexpected_results = ['test.cld updated']
+
+        # verify stdout
+        self.verify_output(output.out, expected=expected_stdout, unexpected=unexpected_results)
+
+        # verify stderr
+        self.verify_output(output.err, expected=expected_stderr, unexpected=unexpected_results)
```
rsundriyal, line 569: Are we expecting expected_stderr to come in some cases as we are matching it. In which case do output.err check is valid and will process ? If error comes in results are we supposed to fail the test or not ? Pardon my knowledge on the test.

rsundriyal (changes requested):

unit_tests/freshclam_test.py
micahsnyder: Good question.  So in this test we do expect some output in the `stderr` stream but it's not a failure.

The `self.verify_output` function will fail if the `output.out` stream doesn't contain the `expected_stdout`.
Likewise it will fail if the `output.err` stream doesn't contain the `unexpected_results`.

For these freshclam testss, what I've done with this and the other tests is check for the exact same "unexpected results" in both stdout and stderr, because I want to make sure the test fails if freshclam writes the unexpected message to the wrong file descriptor by mistake.

rsundriyal (approved):

PR #481, opened 2022-02-24
CMake, Rust: Improve toolchain (target triple) selection

micahsnyder:
The current method of trying to determine the target triple based
the architecture, operating system, etc. is difficult to get right
and maintain. In particular, I found that some installations like the
Alpine package will use "alpine" in the name of the triple instead of
"unknown".
E.g. "aarch64-alpine-linux-musl" instead of "aarch64-unknown-linux-musl".
This makes it nearly impossible to figure out the exact target name
based on target system specifications.

I believe it will be better to use the default target 99% of the time
and require users that which to cross-compile to use a new CMake
variable `-D RUST_TARGET_TRIPLE=<target>` to choose the target.

### DIFF
--- a/CMakeOptions.cmake
+++ b/CMakeOptions.cmake
@@ -117,3 +117,8 @@ option(ENABLE_UNRAR
 option(ENABLE_SYSTEMD
     "Install systemd service files if systemd is found."
     ${ENABLE_SYSTEMD_DEFAULT})
+
+# For reference determining target platform:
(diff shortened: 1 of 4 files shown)


micahsnyder:
@shutton force-pushed to choose a better variable name and improve the INSTALL.md cross-compiling section.

micahsnyder:
Rebased to get past the failed build.  Will double-check it passes the require tests before merging.

PR #486, opened 2022-03-01
0.105 release prep: change version suffix to `-rc` and update news 

micahsnyder:


NEWS.md
```diff
@@ -9,10 +9,86 @@ ClamAV 0.105.0 includes the following improvements and changes.
 
 ### New Requirements
 
+- Starting with ClamAV v0.105, the Rust toolchain is required to compile ClamAV.
+
+  You can install the Rust toolchain for your development environment by
+  following the instructions on the [`rustup`](https://rustup.rs/) website.
+  Some binary package distributions do provide relatively up to date packages of
```
shutton, line 16: "up-to-date" (you did it below)

You could also say "current" or "updated" if you want to mix it up.

NEWS.md
```diff
@@ -9,10 +9,86 @@ ClamAV 0.105.0 includes the following improvements and changes.
...
+  most up-to-date Rust compiler at the time of installation. Keep your toolchain
+  updated for new features and bug/security fixes by periodically executing:
+  ```sh
+  rustup update
+  ```
+
+  Building ClamAV requires, at a minimum, Rust compiler version 1.56, as it
+  relies on features introduced in the Rust 2021 Edition.
+
+  ClamAV's 3rd party Rust library dependencies are vendored into the release
```
shutton, line 27: "3rd-party" or "third-party"

NEWS.md
```diff
@@ -9,10 +9,86 @@ ClamAV 0.105.0 includes the following improvements and changes.
...
+  tarball (.tar.gz) file that we publish on the Downloads page of clamav.net.
```
shutton, line 28: link `Downloads`?

NEWS.md
```diff
@@ -9,10 +9,86 @@ ClamAV 0.105.0 includes the following improvements and changes.
...
+  But, if you build from a Git clone or an unofficial tarball taken from
+  GitHub.com, you will need an active internet connection to download the Rust
```
shutton, line 30: I generally prefer "Internet"

shutton (changes requested):
One other comment.  We're not super-consistent on the terminology around image hashing. ("fuzzy" vs. "perceptual").  IMHO, the latter feels a bit less colloquial.

We're also a bit generic in the code around this (e.g., the FuzzyHash Rust module). We probably should have been more specific to call it out as *image* hashing, but that can be fixed later, and doesn't affect the documentation.

NEWS.md
micahsnyder: Hm yeah third-party is better.

INSTALL.md
```diff
@@ -669,22 +672,58 @@ ClamAV has two bytecode runtimes:
...
+ClamAV currently supports LLVM versions 8.0 through 12.0.
 
+To build with LLVM for the bytecode runtime, use these options, specifying the
+paths and versions appropriate for your LLVM installation:
 ```sh
 cmake .. \
   -D BYTECODE_RUNTIME="llvm"       \
-  -D LLVM_ROOT_DIR="/opt/llvm/3.6" \
-  -D LLVM_FIND_VERSION="3.6.0"
+  -D LLVM_ROOT_DIR="/opt/llvm/8.0" \
```
ragusaa, line 695: I never use LLVM_ROOT_DIR or LLVM_FIND_VERSION in building.  If you only have 1 llvm installed, it may not be necessary.  Do we want to move them down a little and say something like "If llvm is not detected, try these"?

micahsnyder: Good call!

ragusaa (approved):

PR #490, opened 2022-03-04
clang-format: disable AlignConsecutiveAssignments

micahsnyder:
Two reasons:
1. Aligning consecutive assignment results in bigger diffs, affecting
nearby lines.

2. Clang-format has a bug. These two options are not playing nice together:
- AlignConsecutiveAssignments: true
- AlignAfterOpenBracket: true

AlignAfterOpenBracket is not taking account the spaces inserted by
AlignConsecutiveAssignments, so you end up with stuff like this:
```c
    bleeblah = 1;
    blah     = function(arg1,
                    arg2,
                    arg3);

                //  ^--- these args 4-left from where they should be.
```

VSCode, meanwhile, somehow fixes this whitespace issue so code that is
correctly formatted by VSCode doesn't have this bug, meaning that:
1. The clang-format check in GH Actions fails.
2. We'd all have to stop using format-on-save in VSCode and accept the
    bug if we wanted those GH Actions tests to pass.

Disabling AlignConsecutiveAssignments is the best option I can think of.

---

I'm not committed to this change, and we probably want to redo its second commit after the LLVM PR merge if we are going to merge it.  I just want feedback for now.


micahsnyder:
I found the bug: https://github.com/llvm/llvm-project/issues/32893  

It's actually quite old.  Maybe we just didn't notice until we finally edited the affected files?

Issue #494, opened 2022-03-07
clamd-devel stops working if unoffical sigs are enabled in clamd.conf with: OfficialDatabaseOnly** yes/**no**.


HeinoBackhaus:
Hi there,

i'm using "ClamAV 0.105.0-devel-20220307/26474/Mon Mar  7" as a nightly compilation, together with:
################################################################################
eXtremeSHOK.com ClamAV Unofficial Signature Updater
Version: v7.2.5 (2021-03-20)
Required Configuration Version: v96
Copyright (c) Adrian Jon Kriel :: [admin@extremeshok.com](mailto:admin@extremeshok.com)
################################################################################

**Since a 02.28  i recocnize that clamd stops working if unoffical sigs are enabled in clamd.conf with: OfficialDatabaseOnly** yes/**no**.

root@mailscanner2020:# service clamd status
● clamd.service - LSB: Clamd Init.d-script
Loaded: loaded (/etc/init.d/clamd; generated)
Active: active **(exited)** since Mon 2022-02-28 14:02:02 CET; 3s ago
Docs: man:systemd-sysv-generator(8)
Process: 31098 ExecStart=/etc/init.d/clamd start (code=exited, status=0/SUCCESS)
Feb 28 14:02:01 mailscanner2020 clamd[31102]: LibClamAV Error: cli_add_content_match_pattern: Alternative match cannot contain unbounded wildcards
Feb 28 14:02:01 mailscanner2020 clamd[31102]: LibClamAV Warning: cli_loadyara: problem parsing yara file /usr/share/clamav/rfxn.yara, yara rule Backdoor_PHP_WPVCD_Deployer
Feb 28 14:02:01 mailscanner2020 clamd[31102]: LibClamAV Error: cli_add_content_match_pattern: Alternative match cannot contain unbounded wildcards
Feb 28 14:02:01 mailscanner2020 clamd[31102]: LibClamAV Warning: cli_loadyara: problem parsing yara file /usr/share/clamav/rfxn.yara, yara rule Spam_PHP_WPVCD_ContentInjection
Feb 28 14:02:01 mailscanner2020 clamd[31102]: LibClamAV Error: cli_add_content_match_pattern: Alternative match cannot contain unbounded wildcards
Feb 28 14:02:01 mailscanner2020 clamd[31102]: LibClamAV Warning: cli_loadyara: problem parsing yara file /usr/share/clamav/rfxn.yara, yara rule Suspicious_PHP_PrependedInclude
[...]

PR #499, opened 2022-03-12
Yara: Fix support for regex strings

micahsnyder:
In 0.104 and prior, the function for adding a logical subsignature
was being used for NDB sigs, FTM, sigs, *and* Yara strings.

When cleaning up the logic for handling different types of logical
sig subsignatures, we accidentally broke support for regex strings in
yara rules.

This commit adds new logic for recording if the Yara string is a regex
string, by adding a regex subsig opt. Then in a new function for adding
different types of Yara strings, we check if it's a regex string or not
before adding as either a PCRE pattern or as an AC/BM pattern.

Resolves: https://github.com/Cisco-Talos/clamav/issues/494

### DIFF
--- a/libclamav/readdb.c
+++ b/libclamav/readdb.c
@@ -581,6 +581,106 @@ done:
     return status;
 }
 
+static cl_error_t readdb_parse_yara_string(struct cli_matcher *root, const char *virname, char *hexsig, uint8_t subsig_opts,
+                                           const char *offset, uint8_t target, const uint32_t *lsigid, unsigned int options)
+{
+    cl_error_t status = CL_EPARSE;
+    cl_error_t ret;
+    char *hexcpy = NULL;
+
+    char *start = NULL, *end = NULL;
+
+    if (strchr(hexsig, '/')) {
+        /*
+         * Looks like a pcre subsignature.
+         * expected format => ^offset:trigger/regex/[cflags]$
+         */
+        const char *trigger, *pattern, *cflags;
+        int subtokens_count;
+        char *subtokens[SUB_TOKENS + 1];
+        const char *sig;
+
+        subtokens_count = cli_ldbtokenize(hexsig, ':', SUB_TOKENS + 1, (const char **)subtokens, 0);
+        if (!subtokens_count) {
+            cli_errmsg("Invalid or unsupported ldb subsignature format\n");
+            status = CL_EMALFDB;
+            goto done;
+        }
+
+        if ((subtokens_count % 2) == 0)
+            offset = subtokens[0];
+
+        sig = (subtokens_count % 2) ? subtokens[0] : subtokens[1];
+
+        /* get copied */
+        hexcpy = cli_strdup(sig);
+        if (!hexcpy) {
+            status = CL_EMEM;
+            goto done;
+        }
+
+        /* get delimiters-ed */
+        start = strchr(hexcpy, '/');
+        end   = strrchr(hexcpy, '/');
+
+        /* get pcre-ed */
+        if (start == end) {
+            cli_errmsg("PCRE subsig mismatched '/' delimiter\n");
+            status = CL_EMALFDB;
+            goto done;
+        }
+
+        /* get checked */
+        if (hexsig[0] == '/') {
+            cli_errmsg("PCRE subsig must contain logical trigger\n");
+            status = CL_EMALFDB;
+            goto done;
+        }
+
+        /* get NULL-ed */
+        *start = '\0';
+        *end   = '\0';
+
+        /* get tokens-ed */
+        trigger = hexcpy;
+        pattern = start + 1;
+        cflags  = end + 1;
+        if (*cflags == '\0') /* get compat-ed */
+            cflags = NULL;
+
+        /* normal trigger, get added */
+        ret = cli_pcre_addpatt(root, virname, trigger, pattern, cflags, offset, lsigid, options);
+        if (CL_SUCCESS != ret) {
+            cli_errmsg("Problem adding PCRE subsignature.\n");
+            status = ret;
+            goto done;
+        }
+
+    } else {
+        /*
+         * Looks like an AC/BM content match subsignature.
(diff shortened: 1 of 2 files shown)


libclamav/readdb.c
```diff
@@ -581,6 +581,106 @@ cl_error_t readdb_parse_ldb_subsignature(struct cli_matcher *root, const char *v
...
+static cl_error_t readdb_parse_yara_string(struct cli_matcher *root, const char *virname, char *hexsig, uint8_t subsig_opts,
+                                           const char *offset, uint8_t target, const uint32_t *lsigid, unsigned int options)
+{
+    cl_error_t status = CL_EPARSE;
+    cl_error_t ret;
+    char *hexcpy = NULL;
+
+    char *start = NULL, *end = NULL;
+
+    if (strchr(hexsig, '/')) {
```
shutton, line 593: Suggest reversing `if/else` to de-nest some of the logic, or move the "pcre subsignature" handler into its own function to make the logic easier to follow.

shutton (changes requested):

libclamav/readdb.c
```diff
@@ -312,7 +312,100 @@ cl_error_t cli_sigopts_handler(struct cli_matcher *root, const char *virname, co
...
+    cl_error_t status = CL_EPARSE;
+    cl_error_t ret;
+    char *hexcpy = NULL;
+    char *start  = NULL;
+    char *end    = NULL;
+
+    const char *trigger, *pattern, *cflags;
+    int subtokens_count;
+// The maximum number of `:` delimited fields in a regex subsignature.
+#define MAX_REGEX_SUB_TOKENS 4
```
shutton, line 340: Does this belong at the top of the file?  I don't usually see constants defined within functions.

I generally think of constants as "things that might change later, but not without a revision," so grouping them at the top along with all the other potentially-changing things is a practice I've been fairly consistent with (except for unit tests, where constants are often test data).

libclamav/readdb.c
```diff
@@ -312,7 +312,100 @@ cl_error_t cli_sigopts_handler(struct cli_matcher *root, const char *virname, co
...
+
+    /* get copied */
+    hexcpy = cli_strdup(sig);
+    if (!hexcpy) {
+        status = CL_EMEM;
+        goto done;
+    }
+
+    /* get delimiters-ed */
+    start = strchr(hexcpy, '/');
```
shutton, line 364: We discovered that unescaped slashes are tolerated within PCRE subsigs -- this would be why.

libclamav/readdb.c
```diff
@@ -581,22 +620,74 @@ cl_error_t readdb_parse_ldb_subsignature(struct cli_matcher *root, const char *v
...
+                                           const char *offset, uint8_t target, const uint32_t *lsigid, unsigned int options)
+{
+    cl_error_t status = CL_EPARSE;
+    cl_error_t ret;
+
+    if (strchr(hexsig, '/')) {
+        /*
+         * Looks like a pcre subsignature.
+         */
+        ret = readdb_load_regex_subsignature(root, virname, hexsig, offset, lsigid, options);
```
shutton, line 646: Both if/else clauses end with the same thing. De-dup?

shutton (changes requested):

libclamav/readdb.c
micahsnyder: I don't think so. This is to find the start and end of the regex string.  It doesn't have anything to do with `/` inside the regex string.  it just gets the first and last one to find the bounds.

libclamav/readdb.c
micahsnyder: I would be more inclined to remove the constant and hard-code it.  Using a constant makes it easier to read later.  Putting it at the top doesn't feel right to me.  I actually tried with it at the top at first, and then changed my mind and moved it inside the functions.

libclamav/readdb.c
micahsnyder: Done.

libclamav/readdb.c
micahsnyder: To elaborate, the use of a constant here to me feels like a failing of how we tokenize strings in C. I ideally it could handle any number of tokens with an iterator.  So it is an undesirable constant, as opposed to something I would want to be configurable given a choice.  If/when we rewrite this I wouldn't preserve it. It is also specific to this function, and no other part of the file.  So I don't think it belongs at the top of the file.

libclamav/readdb.c
shutton: No worries -- this relates to some [likely inaccurate] documentation on PCRE subsignature requirements.

libclamav/readdb.c
```diff
@@ -581,22 +620,70 @@ cl_error_t readdb_parse_ldb_subsignature(struct cli_matcher *root, const char *v
...
+        if (subsig_opts) {
+            ret = cli_sigopts_handler(root, virname, hexsig, subsig_opts, 0, 0, offset, target, lsigid, options);
+        } else {
+            ret = cli_add_content_match_pattern(root, virname, hexsig, 0, 0, 0, offset, target, lsigid, options);
+        }
+    }
+
+    if (CL_SUCCESS != ret) {
+        status = ret;
+        goto done;
```
shutton, line 661: The `goto done` (and `done:`) here could simply be removed (and the trailing `status = CL_SUCCESS` be put in an `else` clause).

But I know we love `done`... so leave it be if you like. But it's nice not to have `goto`'s

For that matter, you could just return directly within the if/else blocks.

shutton (approved):

libclamav/readdb.c
micahsnyder: Yeah I definitely love the goto's because without them, people extend and extend over time with more and more cruft. Error handling gets copy-pasted and then, inevitably, something gets missed and we're left to fix a leak. 
I know it looks kind of stupid in a function this small.  It's really just a future-proofing habit, I guess.

PR #503, opened 2022-03-16
Fix benign overread in HTML/js-norm parser

micahsnyder:
Fix a possible overread in `handle_de()` where we dereference `tokens`
without boundchecking. The over-read does not cause a crash.
I do not believe this issue to be a vulnerability.

Resolves: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=44849

Also clean up very sloppy bounds checking in `match_parameters()`.
I don't have any evidence of an overflow. The code just looks sus.

Eliminated a benign warning in `cli_js_process_buffer()`.

### DIFF
--- a/libclamav/jsparse/js-norm.c
+++ b/libclamav/jsparse/js-norm.c
@@ -69,10 +69,8 @@ typedef struct scanner {
     enum tokenizer_state last_state;
 } * yyscan_t;
 
-typedef int YY_BUFFER_STATE;
-
 static int yylex(YYSTYPE *lvalp, yyscan_t);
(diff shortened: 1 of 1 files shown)


micahsnyder:
I'd like to run this one through a larger regression test to make sure it doesn't inadvertently change the javascript normalization and cause false negatives.

ragusaa (approved):

micahsnyder:
I haven't merged this yet because in a large regression test I found a number of false negatives and I haven't had time to triage each to determine if there's a bug, or if we will just have to update some signatures after this lands.  Can complete this after 0.105 ships.

PR #504, opened 2022-03-16
sigtool --fuzzy-img FILES

micahsnyder:
Added a new sigtool option `--fuzzy-img` to generate the image fuzzy hash for used in the new `fuzzy_img` logical subsignatures.

You can pass it a list of files if you want, like this:
```
❯ ./install/bin/sigtool --fuzzy-img ~/Downloads/*.png
cisco.png: 9463944473afd82f
clam-gears.png: ffa896c0df02489e
inosuke.png: 85914e78772cf1c6
mamayaga.png: c74b32c337644ab5
```

This should be more convenient than having to run `clamscan --gen-json --debug <file>`

### DIFF
--- a/sigtool/sigtool.c
+++ b/sigtool/sigtool.c
@@ -699,7 +699,7 @@ static int build(const struct optstruct *opts)
     STATBUF foo;
     unsigned char buffer[FILEBUFF];
     char *tarfile, header[513], smbuff[32], builder[33], *pt, olddb[512];
-    char patch[32], broken[32], dbname[32], dbfile[32];
+    char patch[50], broken[57], dbname[32], dbfile[36];
     const char *newcvd, *localdbdir = NULL;
     struct cl_engine *engine;
     FILE *cvd, *fh;
@@ -2445,7 +2445,7 @@ static int decodehex(const char *hexsig)
 {
     char *pt, *hexcpy, *start, *n, *decoded, *wild;
     int asterisk = 0;
-    unsigned int i, j, hexlen, dlen, parts = 0, bw;
+    unsigned int i, j, hexlen, dlen, parts = 0;
     int mindist = 0, maxdist = 0, error = 0;
 
     hexlen = strlen(hexsig);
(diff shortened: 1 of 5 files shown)


sigtool/sigtool.c
```diff
@@ -326,6 +326,97 @@ static int hashsig(const struct optstruct *opts, unsigned int class, int type)
...
+    int target_fd                   = -1;
+    FFIError *fuzzy_hash_calc_error = NULL;
+    uint8_t *mem                    = NULL;
+
+    image_fuzzy_hash_t hash = {0};
+    STATBUF st;
+    ssize_t bytes_read;
+
+    if ((target_fd = open(filename, O_RDONLY)) == -1) {
+        mprintf(LOGG_ERROR, "%s: Can't open file\n", basename(filename));
```
shutton, line 342: These messages seem to lack the error message (ala `strerror()`)-- is that handled elsewhere?

sigtool/sigtool.c
```diff
@@ -326,6 +326,97 @@ static int hashsig(const struct optstruct *opts, unsigned int class, int type)
...
+        goto done;
+    }
+
+    if (NULL == (mem = malloc((size_t)st.st_size))) {
+        mprintf(LOGG_ERROR, "%s: Malloc failed, buffer size: %zu\n", basename(filename), (size_t)st.st_size);
+        goto done;
+    }
+
+    bytes_read = read(target_fd, mem, (size_t)st.st_size);
+    if (bytes_read != (ssize_t)st.st_size) {
```
shutton, line 357: Technically, there are two possible failure results from `read()`, `-1` if an error is encountered, or a smaller-than-expected read size (which should never happen unless the file descriptor is non-blocking). I would handle these separately.

sigtool/sigtool.c
```diff
@@ -2482,7 +2573,7 @@ static int decodehex(const char *hexsig)
                 free(hexcpy);
                 return -1;
             }
-            bw = write(1, decoded, dlen);
+            (void)write(1, decoded, dlen);
```
shutton, line 2576: Ignoring `write()`'s success/fail result?

shutton (changes requested):

sigtool/sigtool.c
micahsnyder: Hmm yeah I should add the `strerror(errno)` output.

shutton (approved):

PR #506, opened 2022-03-17
Fix relative bounds checking in regex_list.c

m-sola:
The check of pattern_len against FILEBUF is largely meaningless since
pattern is derived from a strchr() call against buffer (with length FILEBUFF)

This fix ensures that the relative size is checked against max buffer size
which prevents overwriting stack memory with a single null byte


micahsnyder (approved):
Seems good!  I spent too long reading and debugging to feel comfortable with the assumption that the buffer is null terminated.  But I think it's good.  Will merge.

Issue #430, opened 2022-01-19
possible invalid free in libclamav/matcher-ac.c (v0.104.1)


easylyou:
Describe the bug
----------------

Hi!
We are using improved fuzzing technology with fault injection. I'm pretty sure the bugs I report are far away from security issues. they are more close to robustness. If it's not true or you are still worry about it, please point out and I would edit the issues immediately.

*Different from traditional fuzzing, fault injection can test error-handling codes while the errors happen in some extreme conditions. (such as malloc() return NULL when out of memory). So it is hard to reproduce without our fault injection environment. I try to analysis the related codes to find the cause but not sure, needing your helps. I'm glad to provide more details if you need.*
```c
// in libclamav/matcher-ac.c
2488 cl_error_t cli_ac_addsig( ... )↩
2489 {↩
2490     struct cli_ac_patt *new;↩
         // In default, `MPOOL_CALLOC` is `mpool_calloc`
2507     if ((new = (struct cli_ac_patt *)MPOOL_CALLOC(root->mempool, 1, sizeof(struct cli_ac_patt))) == NULL)↩
2508         return CL_EMEM;↩

2631         if (!(hexnew = (char *)cli_calloc(1, hexnewsz))) {↩
          // But here, `free` the `new`. It is raw `free` in libc. So it is a bad free.
2632             free(new);↩

2635         }↩

2939 }↩
```

```
LibClamAV Warning: **************************************************
LibClamAV Warning: ***  The virus database is older than 7 days!  ***
LibClamAV Warning: ***   Please update it as soon as possible.    ***
LibClamAV Warning: **************************************************
===================FAULT INJECT===================
faultinject: error actived 1! name: calloc, file: libclamav/others_common.c, line: 216
??:0
??:0
??:0
??:0
??:0
??:0
/home/xkt/fitest_work/clamav_log/clamav-0.104.1/libclamav/readdb.c:1880
/home/xkt/fitest_work/clamav_log/clamav-0.104.1/libclamav/readdb.c:1929
??:0
/home/xkt/fitest_work/clamav_log/clamav-0.104.1/libclamav/cvd.c:343
??:0
??:0
/home/xkt/fitest_work/clamav_log/clamav-0.104.1/libclamav/readdb.c:4839
??:0
/home/xkt/fitest_work/clamav_log/clamav-0.104.1/clamscan/manager.c:1146
/home/xkt/fitest_work/clamav_log/clamav-0.104.1/clamscan/clamscan.c:171
[...]

PR #507, opened 2022-03-17
Changed free to MPOOL_FREE

ragusaa:
'new' is allocated by mpool, so should be freed by the mpool free
function.


micahsnyder (approved):

micahsnyder:
Resolves: https://github.com/Cisco-Talos/clamav/issues/430

PR #508, opened 2022-03-17
docker: make clamd listen on IPv4 and v6 ADDR_ANY

### LINKED ISSUE #505: Clamdcheck.sh is not working when other clients are
Describe the bug
----------------

We use clamav in a kubernetes cluster and experience some issues with the service not responding anymore (which maybe is related to #330?). In order to restart clamav in this situation we implemented kubernetes probes like that:

```
startupProbe:
  exec:
    command:
      - clamdcheck.sh
  periodSeconds: 10
  failureThreshold: 120
readinessProbe:
  exec:
    command:
      - clamdcheck.sh
livenessProbe:
  exec:
    command:
      - clamdcheck.sh
```

Unfortunately, the `clamdcheck.sh` script which uses netcat to ping the service doesn't seem to be working as kubernetes never clears the startup warnings. In addition, using a shell on a pod and executing the `clamdcheck.sh` script returns `ERROR: Unable to contact server` even when clamd seems to be working correctly. However, when port-forwarding to a local dev machine and talking to clamd via TCP (java-client) both ping and scan are working. What's the problem here? I don't really want to check for the clamd process running in the probes rather than pinging the service via network which is the same a client does.

How to reproduce the problem
----------------------------

Reproduction should be clear as stated above. Here's the full k8s yaml:

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  labels:
    app.kubernetes.io/name: clamav
  name: clamav
spec:
  replicas: 1
  serviceName: clamav
  selector:
    matchLabels:
      app.kubernetes.io/name: clamav
  template:
    metadata:
      labels:
        app.kubernetes.io/name: clamav
    spec:
      containers:
        - image: clamav/clamav:0.104.2_base
          imagePullPolicy: IfNotPresent
          name: clamav
          resources:
            requests:
              cpu: 1000m
              memory: 4Gi
            limits:
              cpu: 1000m
              memory: 4Gi
          startupProbe:
            exec:
              command:
                - clamdcheck.sh
            periodSeconds: 10
            failureThreshold: 120
          readinessProbe:
            exec:
              command:
                - clamdcheck.sh
          livenessProbe:
            exec:
              command:
                - clamdcheck.sh
          volumeMounts:
            - mountPath: /var/lib/clamav
              name: database
      restartPolicy: Always
  volumeClaimTemplates:
    - metadata:
        name: database
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: 4Gi
```

andremae:
Fix #505 by making clamd listen to all addresses available (IPv4 and IPv6). This is the default behavior of clamd meaning that commenting out the `TCPAddr 0.0.0.0` in `/etc/clamav/clamd.conf` should work.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ RUN apk add --no-cache \
         -e "s|.*\(PidFile\) .*|\1 /run/lock/clamd.pid|" \
         -e "s|.*\(LocalSocket\) .*|\1 /run/clamav/clamd.sock|" \
         -e "s|.*\(TCPSocket\) .*|\1 3310|" \
-        -e "s|.*\(TCPAddr\) .*|\1 0.0.0.0|" \
+        -e "s|.*\(TCPAddr\) .*|#\1 0.0.0.0|" \
         -e "s|.*\(User\) .*|\1 clamav|" \
         -e "s|^\#\(LogFile\) .*|\1 /var/log/clamav/clamd.log|" \
         -e "s|^\#\(LogTime\).*|\1 yes|" \


micahsnyder:
Looks great to me! Thanks @andremae!

micahsnyder (approved):

PR #523, opened 2022-04-03
Minor fixes: remove a couple of unused variables

rzvncj:
Removing `err` from *libclamav/png.c* and `ret` from *clamav/clamonacc/scan/onas_queue.c* gets rid of a couple of compile-time warnings. The changes should be trivial to review and merge.


micahsnyder (approved):
Looks good to me, thanks @rzvncj

Issue #525, opened 2022-04-03
SIGBUS on HP-UX 11.31/IA


twwlogin:
Describe the bug
----------------

We built 0.103.5 on HP-UX 11.31/IA and encountered a SIGBUS from clamscan when scanning an Excel spreadsheet with VBAscript.

Core was generated by `clamscan'.
Program terminated with signal 10, Bus error.
BUS_ADRALN - Invalid address alignment. Please refer to the following link that helps in handling unaligned data: http://docs.hp.com/en/7730/newhelp0610/pragmas.htm#pragma-pack-ex3
#0  0xc000000004213970:1 in cli_vba_readdir_new (ctx=0x87ffffffffffe888, 
    dir=0x87ffffffffffdeb0 "/var/tmp//20220403_230927-scantem.457869304e/ole2-tmp.c3f3d738c2/000002/000003", U=0x600000000136cd20, 
    hash=0x600000000136d57c "736007832d2167baaae763fd3a3f3cf1", which=1, 
    tempfd=0x87ffffffffffe6b0, has_macros=0x87ffffffffffe700)
   from /opt/TWWfsw/clamav0103/lib/libclamav.so.9

How to reproduce the problem
----------------------------

$ clamscan CommentPicCode.xls

Attachments
-----------

[https://www.contextures.com/CommentPicCode.zip](https://www.contextures.com/CommentPicCode.zip)

ragusaa:
This appears to be fixed by https://github.com/Cisco-Talos/clamav/pull/526, so I am going to close it.  Please re-open if you are still having problems.

PR #526, opened 2022-04-03
libclamav/vba_extract.c: Fix SIGBUS on HP-UX/IA when built as a 64-bit binary

twwlogin:
Fix for issue #525.

### DIFF
--- a/libclamav/vba_extract.c
+++ b/libclamav/vba_extract.c
@@ -462,15 +462,16 @@ cl_error_t cli_vba_readdir_new(cli_ctx *ctx, const char *dir, struct uniq *U, co
     CLI_WRITEN("REM VBA project extracted from Microsoft Office document\n\n", 58);
 
     for (data_offset = 0; data_offset < data_len;) {
-        uint16_t id;
-        uint32_t size;
+        uint16_t id, val16;
+        uint32_t size, val32;
 
         if (sizeof(uint16_t) > data_len - data_offset) {
             cli_warnmsg("vba_readdir_new: Failed to read record type from dir\n");
             ret = CL_EREAD;
             goto done;
         }
-        id = le16_to_host(*(uint16_t *)&data[data_offset]);
+        memcpy(&val16, &data[data_offset], sizeof (uint16_t));
+        id = le16_to_host(val16);
         data_offset += sizeof(uint16_t);
 
         if (sizeof(uint32_t) > data_len - data_offset) {
@@ -478,7 +479,8 @@ cl_error_t cli_vba_readdir_new(cli_ctx *ctx, const char *dir, struct uniq *U, co
             ret = CL_EREAD;
             goto done;
         }
-        size = le32_to_host(*(uint32_t *)&data[data_offset]);
+        memcpy(&val32, &data[data_offset], sizeof (uint32_t));
+        size = le32_to_host(val32);
         data_offset += sizeof(uint32_t);
 
         if (size > data_len - data_offset) {
@@ -495,7 +497,8 @@ cl_error_t cli_vba_readdir_new(cli_ctx *ctx, const char *dir, struct uniq *U, co
                     ret = CL_EREAD;
                     goto done;
                 }
-                uint32_t sys_kind = le32_to_host(*(uint32_t *)&data[data_offset]);
+                memcpy(&val32, &data[data_offset], sizeof (uint32_t));
+                uint32_t sys_kind = le32_to_host(val32);
                 data_offset += sizeof(uint32_t);
                 CLI_WRITEN("REM PROJECTSYSKIND: ", 20);
                 switch (sys_kind) {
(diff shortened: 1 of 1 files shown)


shutton:
This doesn't strike me as the best way to fix this (and might even break it).  I think the problem really started upstream, in that `cvi_vba_inflate()` returns a `char *`, but that's a cast of `blobToMem()`'s return type, which is `void *`. Had the return chain simply remained consistent with `void *`, this could have been used in `cli_vba_readdir_new()` rather than `char *`, which your compiler seems to be treating more liberally with respect to alignment.

I'm also concerned that this is breaking indexing into `data` since the expectation is that it can be indexed byte-wise, but declaring it as a larger integer will cause indexing to be at that granularity.

Bottom line, I think the better fix is to simply declare it `void *`, and then alias to (char *) so that it's correctly aligned.

twwlogin:
> Bottom line, I think the better fix is to simply declare it `void *`, and then alias to (char *) so that it's correctly aligned.

So your suggestion is something like:
--- vba_extract.c.orig  2022-04-03 13:00:41 +0000
+++ vba_extract.c       2022-04-05 21:27:27 +0000
@@ -363,7 +363,7 @@
     cl_error_t ret = CL_SUCCESS;
     char fullname[1024];
     int fd              = -1;
-    unsigned char *data = NULL;
+    void *data = NULL;
     size_t data_len;
     size_t data_offset;
     const char *stream_name = NULL;

shutton:
> > Bottom line, I think the better fix is to simply declare it `void *`, and then alias to (char *) so that it's correctly aligned.
> 
> So your suggestion is something like: --- vba_extract.c.orig 2022-04-03 13:00:41 +0000 +++ vba_extract.c 2022-04-05 21:27:27 +0000 @@ -363,7 +363,7 @@ cl_error_t ret = CL_SUCCESS; char fullname[1024]; int fd = -1;
> 
> * unsigned char *data = NULL;
> 
> * void *data = NULL;
>   size_t data_len;
>   size_t data_offset;
>   const char *stream_name = NULL;

You'd need to test it on your platform to confirm it resolves the error (since we definitely don't have any HP-UX/IA systems in our test pipelines), but I was thinking something along these lines (to minimize the changed code):
``` c
void *inflate_data = NULL
char *data = NULL;
...
    if ((inflate_data = (data_type_t *)cli_vba_inflate(fd, 0, &data_len)) == NULL) {
         cli_dbgmsg("vba_readdir_new: Failed to decompress 'dir'\n");
         ret = CL_EARG;
         goto done;
    }
    data = inflate_data;
...
```
I'm thinking that will coax a compatible alignment without additional trickery.

twwlogin:
> ```c
> void *inflate_data = NULL
> char *data = NULL;
> ...
>     if ((inflate_data = (data_type_t *)cli_vba_inflate(fd, 0, &data_len)) == NULL) {
>          cli_dbgmsg("vba_readdir_new: Failed to decompress 'dir'\n");
>          ret = CL_EARG;
>          goto done;
>     }
>     data = inflate_data;
> ...
> ```

Did you mean to leave the cast to (data_type_t *)?

twwlogin:
BTW, I compiled vba_extract.c with -Wcast-align on this platform and it gives a warning on any call to le16_to_host() and le32_to_host() where& data[data_offset] is referenced:
orig-vba_extract.c: In function 'cli_vba_readdir_new':
orig-vba_extract.c:474:14: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:474:14: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:482:16: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:482:16: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:482:16: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:482:16: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:499:37: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:499:37: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:499:37: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:499:37: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:535:33: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:535:33: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:535:33: warning: cast increases required alignment of target type [-Wcast-align]
orig-vba_extract.c:535:33: warning: cast increases required alignment of target type [-Wcast-align]
...

twwlogin:
How about something like this as a solution. Rather than:
```
size = le32_to_host(*(uint32_t *)&data[data_offset]);
```
we would use:
```
uint32_t val;
memcpy(&data[data_offset], &val, sizeof (uint32_t));
size = le32_to_host(val);
```

Less efficient though.

shutton:
Doesn't changing the type alter how this would work?
``` c
&data[data_offset]
```
You'd need to adjust all those macros to adjust the index value.

E.g., take this example code:
``` c
#include <stdio.h>
#include <stdint.h>

int main() {
    unsigned char *stuff1 = NULL;
    printf("stuff1[1] = %p\n", &stuff1[1]);
    uint32_t  *stuff2 = NULL;
    printf("stuff2[1] = %p\n", &stuff2[1]);
}
```
You get rather different results:
```
stuff1[1] = 0x1
stuff2[1] = 0x4
```

twwlogin:
We switch using memcpy() as illustrated in this commit: 273663c164450c3871b600eff09462a65462c4a3.

micahsnyder (approved):
Testing looked good. Only saw expected failures with.  Nothing new or unusual.

PR #462, opened 2022-02-10
Change data pointers to use unsigned types

ragusaa:
Changed const char pointers to uint8_t pointers when they are to be used
with data, as well as removing asserts and adding additional error
checking.

PR #530, opened 2022-04-04
Changed data pointers to use unsigned types

ragusaa:
Changed const char pointers to uint8_t pointers when they are to be used
with data, as well as removing asserts and adding additional error
checking.

### DIFF
--- a/libclamav/regex_list.c
+++ b/libclamav/regex_list.c
@@ -39,7 +39,6 @@
 
 #include <limits.h>
 #include <sys/types.h>
-#include <assert.h>
 
 #include "regex/regex.h"
 
@@ -162,13 +161,35 @@ cl_error_t regex_list_match(struct regex_matcher *matcher, char *real_url, const
     struct cli_ac_data mdata;
     struct cli_ac_result *res = NULL;
 
-    assert(matcher);
-    assert(real_url);
-    assert(display_url);
+    if (NULL == matcher) {
+        rc = CL_ENULLARG;
+        cli_errmsg("regex_list_match: matcher must be initialized\n");
+        goto done;
+    }
+
+    if (NULL == real_url) {
+        rc = CL_ENULLARG;
+        cli_errmsg("regex_list_match: real_url must be initialized\n");
+        goto done;
+    }
+
+    if (NULL == display_url) {
+        rc = CL_ENULLARG;
+        cli_errmsg("regex_list_match: display_url must be initialized\n");
+        goto done;
+    }
+
     *info = NULL;
-    if (!matcher->list_inited)
-        return CL_SUCCESS;
-    assert(matcher->list_built);
+    if (1 != matcher->list_inited) {
+        rc = CL_SUCCESS;
+        goto done;
+    }
+    if (0 == matcher->list_built) {
+        cli_errmsg("regex_list_match: matcher->list_built must be initialized\n");
+        rc = CL_ENULLARG;
+        goto done;
+    }
+
     /* skip initial '.' inserted by get_host */
     if (real_url[0] == '.') real_url++;
     if (display_url[0] == '.') display_url++;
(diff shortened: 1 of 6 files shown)


ragusaa:
This PR replaces https://github.com/Cisco-Talos/clamav/pull/462

libclamav/regex_list.c
```diff
@@ -630,21 +706,33 @@ static int add_newsuffix(struct regex_matcher *matcher, struct regex_list *info,
...
+    if (filter_add_static(&matcher->filter, (const unsigned char *)suffix, len, "regex") < 0) {
+        MPOOL_FREE(matcher->mempool, new->pattern);
+        MPOOL_FREE(matcher->mempool, new);
+        cli_errmsg("add_newsuffix: Unable to add filter\n");
+        ret = CL_ERROR;
+        goto done;
+    }
+
+done:
+    return ret;
```
micahsnyder, line 735: Can I trouble you to do something like this after the `done:`:

```c
if (CL_SUCCESS != ret) {
    if (NULL != new) {
        if (NULL != new->pattern) {
            MPOOL_FREE(matcher->mempool, new->pattern);
        }
        MPOOL_FREE(matcher->mempool, new);
    }
}
```

Then remove the copy-paste free's from the error conditions for:
```c
    if (!new->pattern) {
```
```c
    if ((ret = cli_ac_addpatt(root, new))) {
```
```c
    if (filter_add_static(&matcher->filter, (const unsigned char *)suffix, len, "regex") < 0) {
```

Other than this one thing, this PR looks great to me.  Everything checked out in testing.

micahsnyder (changes requested):
1 error handling thing to tidy up, otherwise is good

libclamav/regex_list.c
ragusaa: good idea, done

micahsnyder:
From offline conversations: this PR is presently held up because the cleanup code added is slow and is causing timeouts in the test environment.  @ragusaa is working on finding a faster way to track those pointers for cleanup.

libclamav/regex_list.c
```diff
@@ -594,24 +639,55 @@ void regex_list_done(struct regex_matcher *matcher)
...
+
+    if (NULL == suffix) {
+        cli_errmsg("add_newsuffix: suffix must be initialized\n");
+        ret = CL_ENULLARG;
+        goto done;
+    }
+
+    new = MPOOL_CALLOC(matcher->mempool, 1, sizeof(*new));
+    if (!new) {
+        cli_errmsg("add_newsuffix: unable to add \n");
```
micahsnyder, line 687: This error message is only half-formed.

micahsnyder (changes requested):
With exception to the minor changes requested above, this all looks really great.  It went through the test pipelines nicely.  I used the internal-fuzz-corpus PR to verify that it resolves CLAM-1607 and the related ones (1642, 1649, 1653) and with manual testing, confirmed that 1691 is resolved as well.

micahsnyder:
Just rebased for you to same some effort, since I did it for testing locally.

libclamav/regex_list.c
```diff
@@ -162,13 +161,35 @@ cl_error_t regex_list_match(struct regex_matcher *matcher, char *real_url, const
...
+        rc = CL_ENULLARG;
+        cli_errmsg("regex_list_match: display_url must be initialized\n");
+        goto done;
+    }
+
     *info = NULL;
-    if (!matcher->list_inited)
-        return CL_SUCCESS;
-    assert(matcher->list_built);
+    if (!matcher->list_inited) {
```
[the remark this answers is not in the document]
ragusaa: I really expected to find that the entire matcher struct was memset to 0, but I can't find anything.  In init_regex_list, matcher is memset to 0, and list_inited is immediately set to 1, so I think you are correct in saying that it is always either 1 or -1.  The variables in the same structure list_built and list_loaded both use 0 and 1, so I agree that changing the errors from -1 to 0 makes sense.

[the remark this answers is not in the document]
ragusaa: In looking at it a little more, I think choices are 0, 1, -1 because it is possible that someone could allocate a regex_matcher struct without calling init_regex_list, and that is why there are checks for being initialized.

micahsnyder (approved):
lgtm!

PR #535, opened 2022-04-07
Fix small leak when loading invalid FTM signatures

ragusaa:
Fix small leak when loading invalid FTM signatures

Resolves: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43844


micahsnyder (approved):
Super simple fix, looks great. Testing worked out well (had to disable MPOOL to reproduce it).

PR #540, opened 2022-04-11
Make idle-timeout to control running process

net1:
clamd thread won't killed or exited if client disconnect while process still scan so process won't run longer than idle-timeout limit since both ReadTimeout and CommandReadTimeout setting can't control the process if it still scanning.

### DIFF
--- a/clamd/thrmgr.c
+++ b/clamd/thrmgr.c
@@ -700,10 +700,10 @@ static void *thrmgr_worker(void *arg)
             exit(-2);
         }
         if (job_data) {
-            threadpool->handler(job_data);
-        } else if (must_exit) {
-            break;
-        }
+            if (must_exit) {
+                break;
+            }                
+        } 
     }
     if (pthread_mutex_lock(&(threadpool->pool_mutex)) != 0) {
         /* Fatal error */


clamd/thrmgr.c
```diff
@@ -700,10 +700,10 @@ static void *thrmgr_worker(void *arg)
...
         }
         if (job_data) {
-            threadpool->handler(job_data);
-        } else if (must_exit) {
-            break;
-        }
+            if (must_exit) {
+                break;
+            }                
+        }
```
micahsnyder, line 706: I'm not sure I understand the issue you're trying to fix. But this doesn't look like a correct solution.  I'm pretty sure we need to call the `handler()` function.  

And the `if (job_data)` condition exists so we only call it if the `job_data` is defined. If it weren't needed, then I think you could change it to just:
```c
        if (must_exit) {
            break;
        }
```
but since I think it is needed -- maybe the change you were looking for is this?:
```c
        if (job_data) {
            threadpool->handler(job_data);
        }
        if (must_exit) {
            break;        
        }
```

or maybe this?:
```c
        if (must_exit) {
            break;        
        }
        if (job_data) {
            threadpool->handler(job_data);
        }
```

I'm not recommending these changes.  I haven't tested anything yet.  I'm still not sure what the problem is or how to test that it is fixed.

Can you provided steps for how to reproduce the problem so I can test these possible fixes?

micahsnyder (changes requested):

clamd/thrmgr.c
net1: After submit this pull and leave system run more and check I found that it is not fix process thread when client terminate connection or timeout limit. I still can not find how to terminate or cancel scan thread when it already got ETIMEDOUT. There should have some control function to stop thread immediately but I could not found one or I am not understand it because I am not good in C language.

micahsnyder:
> After submit this pull and leave system run more and check I found that it is not fix process thread when client terminate connection or timeout limit. I still can not find how to terminate or cancel scan thread when it already got ETIMEDOUT. There should have some control function to stop thread immediately but I could not found one or I am not understand it because I am not good in C language.

Closing this PR since it doesn't solve the problem. 

Based on what you described back in April, I guess the issue is that the scan may time out from `clamdscan`'s perspective, but `clamd` continues to scan despite the fact that it doesn't have anything listening for results when it eventually finishes. And that is a waste of resources.

Perhaps someone will revisit this later. @net1 feel free to submit a github issue describing this problem if it continues to be a concern/problem.

PR #545, opened 2022-04-13
Check allocation before MPOOL_FREE

ragusaa:
Since virname is passed in, and only cloned under certain condtions,
check to see that it has actually been cloned before freeing it in any
cleanup code.


micahsnyder (approved):
@ragusaa I force-pushed to fix a formatting issue and add details to the commit message. No other changes.  Your fix looks good to me.

PR #547, opened 2022-04-14
Clam 1656 Fix leak loading malformed IDB databases

ragusaa:
Free matcher on error condition.

Reproducer artifacts are https://github.com/Cisco-Talos/clamav/suites/4980638814/artifacts/145471677

To reproduce, build with address sanitization enabled and mpool disabled
'cmake -G Ninja .. -DCMAKE_CXX_COMPILER=clang++-8 -DCMAKE_C_COMPILER=clang-8 -DCMAKE_LINKER=clang-8 -DCMAKE_INSTALL_PREFIX=~/install -DCMAKE_C_FLAGS='\''-fsanitize=address -g -fno-omit-frame-pointer'\'' -DENABLE_TESTS=OFF -DDISABLE_MPOOL=1'

Rename the reproducer test case to '.idb', and try and load the signature
'clamscan -d test.idb'

Description is here https://github.com/Cisco-Talos/clamav/runs/4860615522?check_suite_focus=true 
Click Summary->artifacts

### DIFF
--- a/libclamav/readdb.c
+++ b/libclamav/readdb.c
@@ -1286,13 +1286,13 @@ static cl_error_t cli_loaddb(FILE *fs, struct cl_engine *engine, unsigned int *s
 #define ICO_TOKENS 4
 static cl_error_t cli_loadidb(FILE *fs, struct cl_engine *engine, unsigned int *signo, unsigned int options, struct cli_dbio *dbio)
 {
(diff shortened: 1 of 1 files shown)


micahsnyder (approved):
I would've preferred swapping this to a `goto done;` solution, but this will suffice.

micahsnyder:
Rebased to see it go through the test pipelines.

Issue #418, opened 2022-01-10
Filenames with Umlaut can't be scanned


tigerplush:
Describe the bug
----------------

I have a ClamD running with an open WebSocket on Port 3310. I have written a c# program that connects with the ClamD via the Websocket and sends a scan command (`nSCAN file\n`). Because I have to be able to handle german files the system was tested with files with Umlauts in their name ( DavSEC_Testdokument_verschlüsselt.pdf ).

Relevant code snippet:
```c#
Socket sender = new Socket(_address.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
sender.Connect(_endpoint);

byte[] msg = Encoding.UTF8.GetBytes(command);
byte[] buffer = new byte[1024];
sender.Send(msg);
int receivedBytes = sender.Receive(buffer);
string returnMessage = Encoding.UTF8.GetString(buffer, 0, receivedBytes);
sender.Shutdown(SocketShutdown.Both);
sender.Close();
```

`Encoding.Default` is UTF8 with name utf-8 and code page 65001

Environment
----------------
Windows 10 Pro 21H2
Processor: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz   2.59 GHz
RAM: 32,0 GB
64bit architecture

ClamAV Version 0.104.1 
daily.cld database is up-to-date (version: 26418, sigs: 1970616, f-level: 90, builder: raynman)
main.cvd database is up-to-date (version: 62, sigs: 6647427, f-level: 90, builder: sigmgr)
bytecode.cvd database is up-to-date (version: 333, sigs: 92, f-level: 63, builder: awillia2)

How to reproduce the problem
----------------------------

Create pdf with an umlaut in it's name (e.g. `C:\file_with_ümlaut.pdf`)
Open ClamD
Connect via websocket
send command `nSCAN C:\file_with_ümlaut.pdf\n`

returns:
`C:\file_with_ümlaut.pdf File path check failure: Result too large or too small. ERROR\n`

ClamD console also shows:
`LibClamAV Warning: Can't open file C:\file_with_ümlaut.pdf: 2`
`Mon Jan 10 14:37:46 2022 -> ^File path check failure for: C:\file_with_ümlaut.pdf`

micahsnyder:
Using our `clamdscan.exe` client with an eicar file containing the umlaut, I reproduced the error message you reported.  I think this is a Windows-specific error having to do with the win32 API's we're using to verify real paths.  

[...]

PR #549, opened 2022-04-16
Clam 1782 utf8 win32 issues

micahsnyder:
- Win32: Fix utf8 filename support issue

The function used to verify the real path for the file being
scanned fails on some utf8 filenames, like:
    file_with_ümlaut.eicar.com

The scan continues despite the failure, but the error reported from
clamd and warnings from clamscan are confusing because it looks like
the scan failed even if a verdict shows up later.

The fix here is to convert the path from utf8 to utf16 and then use
the CreateFileW() API to grab the file handle for the real path check.

Resolves: https://github.com/Cisco-Talos/clamav/issues/418

- libclamav: fix utf8 logging issues on Windows

Print to the log using rust eprint() instead of fputs()

Rust's eprint() function supports full utf8, while fputs() gets
confused and prints stuff like:
    file_with_├╝mlaut.eicar.com
instead of:
    file_with_ümlaut.eicar.com

### DIFF
--- a/libclamav/libmspack.c
+++ b/libclamav/libmspack.c
@@ -20,6 +20,7 @@
 #include "fmap.h"
 #include "scanners.h"
 #include "others.h"
+#include "clamav_rust.h"
 
 enum mspack_type {
     FILETYPE_DUNNO,
@@ -290,7 +291,7 @@ static void mspack_fmap_message(struct mspack_file *file, const char *fmt, ...)
         buff[strlen(buff)]     = '\n';
         buff[strlen(buff) + 1] = '\0';
 
-        cli_eprintf(buff);
+        clrs_eprint(buff);
     }
 }
 
--- a/libclamav/libmspack.c
+++ b/libclamav/libmspack.c
@@ -290,7 +290,7 @@ static void mspack_fmap_message(struct mspack_file *file, const char *fmt, ...)
         buff[strlen(buff)]     = '\n';
         buff[strlen(buff) + 1] = '\0';
 
-        fputs(buff, stderr);
+        cli_eprintf(buff);
     }
 }
 
--- a/libclamav/others_common.c
+++ b/libclamav/others_common.c
@@ -1481,7 +1481,10 @@ cl_error_t cli_realpath(const char *file_name, char **real_filename)
     char *real_file_path = NULL;
     cl_error_t status    = CL_EARG;
 #ifdef _WIN32
-    HANDLE hFile = INVALID_HANDLE_VALUE;
+    HANDLE hFile   = INVALID_HANDLE_VALUE;
+    wchar_t *wpath = NULL;
+    WIN32_FILE_ATTRIBUTE_DATA attrs;
+
 #elif C_DARWIN
     int fd = -1;
 #endif
@@ -1495,7 +1498,13 @@ cl_error_t cli_realpath(const char *file_name, char **real_filename)
 
 #ifdef _WIN32
 
-    hFile = CreateFileA(file_name,                  // file to open
+    wpath = uncpath(file_name);
+    if (!wpath) {
+        errno = ENOMEM;
+        return -1;
+    }
+
+    hFile = CreateFileW(wpath,                      // file to open
                         GENERIC_READ,               // open for reading
                         FILE_SHARE_READ,            // share for reading
                         NULL,                       // default security
@@ -1552,6 +1561,9 @@ done:
 #ifdef _WIN32
     if (hFile != INVALID_HANDLE_VALUE) {
         CloseHandle(hFile);
+    }
+    if (NULL != wpath) {
+        free(wpath);
     }
 #elif C_DARWIN
     if (fd != -1) {
--- a/libclamav/others_common.c
+++ b/libclamav/others_common.c
@@ -1349,7 +1349,7 @@ cl_error_t cli_get_filepath_from_handle(HANDLE hFile, char **filepath)
 
     if (0 == wcsncmp(L"\\\\?\\UNC", long_evaluated_filepathW, wcslen(L"\\\\?\\UNC"))) {
         conv_result = cli_codepage_to_utf8(
-            long_evaluated_filepathW,
+            (char *)long_evaluated_filepathW,
             (wcslen(long_evaluated_filepathW)) * sizeof(WCHAR),
             CODEPAGE_UTF16_LE,
             &evaluated_filepath,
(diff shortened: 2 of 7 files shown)


micahsnyder:
While working on #418, I also took the opportunity to fix a unicode printing issue with a commit I'd been sitting on a while that swaps out the C {{fputs()}} API for Rust's {{eprint()}} function because it can handle utf8 on windows.  It feels silly to do FFI over to Rust for such a small thing, but I couldn't figure out a better way to do it on Windows in C. 

Before that commit, a debug log from the scan looks like this:
```
LibClamAV debug: Checking realpath of C:\Users\micasnyd\Downloads\file_with_├╝mlaut.eicar.com
```
After that commit, I see this:
```
LibClamAV debug: Checking realpath of C:\Users\micasnyd\Downloads\file_with_ümlaut.eicar.com
```

libclamav_rust/src/logging.rs
```diff
@@ -81,3 +84,9 @@ mod tests {
         error!("friend.");
     }
 }
+
+#[no_mangle]
+pub extern "C" fn clrs_eprint(c_buf: *const c_char) -> () {
```
shutton, line 89: This function needs some documentation, at least with respect to its intent.  It looks like you're trying to avail yourself of the Rust standard library's locking around writing to `stderr`, and pretty-printing any non-UTF8 characters as �. But, unless you're consistent throughout the code at never utilizing the C library's `fprintf(stderr, ...)`, you'll still end up intermingling output since the two implementations use separate buffers.

Also need (minimally) a NULL check on `c_buf`.

shutton (changes requested):

libclamav_rust/src/logging.rs
micahsnyder: I'll check if `fprintf(stderr` is used within scans outside of our logging functions.  If there is other use of `fprintf(stderr` and it's not a simple fix, I may just back-off the "[libclamav: fix utf8 logging issues on Windows](https://github.com/Cisco-Talos/clamav/pull/549/commits/cbf3f592b11e184bd8952489e69cd9b911c4724f)" commit.

micahsnyder: > Also need (minimally) a NULL check on c_buf.

👍 good point.

libclamav/libmspack.c
```diff
@@ -290,7 +290,7 @@ static void mspack_fmap_message(struct mspack_file *file, const char *fmt, ...)
         buff[strlen(buff)]     = '\n';
         buff[strlen(buff) + 1] = '\0';
 
-        fputs(buff, stderr);
+        cli_eprintf(buff);
```
shutton, line 293: Any function that utilizes a format string should never be passed the content as that format string (lest it contain escapes like `%s` that would cause an unintended stack read).  I think you probably meant:
``` c
cli_eprintf("%s", buff);

File: libclamav/others_common.c [hunk 3]
```diff
@@ -176,18 +177,31 @@ inline void cli_dbgmsg(const char *str, ...)
...
 
+size_t cli_eprintf(const char *str, ...)
+{
+    va_list args;
+    char buff[MSGBUFSIZ];
+    va_start(args, str);
+    vsnprintf(buff, sizeof(buff), str, args);
+    buff[sizeof(buff) - 1] = '\0';
+    va_end(args);
+    clrs_eprint(buff);
```
shutton, line 200: Any mechanism to return errors?

shutton (changes requested):

libclamav/others_common.c
micahsnyder: I have `clrs_eprint()` printing lossy, so that invalid utf-8 graphemes render inline with the rest of a message rather than failing to log it entirely.  For `stderr` logg messages I think that makes sense. 

Personally I'm okay with that.  If you pass in a NULL pointer, having it print nothing and not report an error also seems fine to me.  It's not like anyone is going to put error handling on a log function anyways.

libclamav/libmspack.c
micahsnyder: WHOOOPS!  Yes thanks for the catch.

micahsnyder: Yeah looks it should be this 
```c
        clrs_eprint(buff);
```
instead of 
```c
        cli_eprintf(buff);
```
since the buffer is already formatted.

libclamav/others_common.c
shutton: S'okay.  Had to say something.

libclamav/others_common.c
```diff
@@ -176,18 +177,31 @@ inline void cli_dbgmsg(const char *str, ...)
...
+
+    return strlen(buff);
```
shutton, line 202: vsnprintf() returns the number of characters written.  Perhaps a light optimization based on that rather than scanning the string again?

libclamav/others_common.c
```diff
@@ -176,18 +177,31 @@ inline void cli_dbgmsg(const char *str, ...)
...
     }
 }
 
+size_t cli_eprintf(const char *str, ...)
+{
+    va_list args;
+    char buff[MSGBUFSIZ];
+    va_start(args, str);
+    vsnprintf(buff, sizeof(buff), str, args);
+    buff[sizeof(buff) - 1] = '\0';
```
shutton, line 198: `vsnprintf()` already does this for you:
> The snprintf() and vsnprintf() functions will write at most size-1 of the characters printed into the output string (the size'th character then gets the terminating ‘\0’); if the return value is greater than or equal to the size argument, the string was too short and some of the printed characters were discarded.  The output is always null-terminated, unless size is 0.

shutton (changes requested):

ragusaa (approved):

libclamav/others_common.c
micahsnyder: I think you're right.  I noticed that is not the case for Microsoft's `_vsnprintf`, but as far as I know, `vsnprintf` on every platform will insert the `\0` unless `sizeof(buff) == 0`. 

I had copied it from:
```c

#define MSGCODE(buff, len, x)                             \
    va_list args;                                         \
    size_t len = sizeof(x) - 1;                           \
    char buff[MSGBUFSIZ];                                 \
    strncpy(buff, x, len);                                \
    va_start(args, str);                                  \
    vsnprintf(buff + len, sizeof(buff) - len, str, args); \
    buff[sizeof(buff) - 1] = '\0';                        \
    va_end(args)
```

I suppose we could remove the `buff[sizeof(buff) - 1] = '\0';` from that macro as well.

shutton (approved):

Issue #521, opened 2022-03-31
clamscan reports FOUND and also then OK for XLS scan


fanto666:
Describe the bug
----------------

some files are detected only when allscan is used. clamav shows name of malware, but reports no problem detected.
This was reported in mailing list:
https://lists.clamav.net/pipermail/clamav-users/2022-March/012539.html

% file intamldeosreitlu.xls
intamldeosreitlu.xls: Composite Document File V2 Document, Little Endian, Os: Windows, Version 10.0, Code page: 1251, Author: Posik, Last Saved By: HRdtjnd, Name of Creating Application: Microsoft Excel, Create Time/Date: Fri Jun  5 19:19:34 2015, Last Saved Time/Date: Mon Mar 28 08:33:09 2022, Security: 0

% clamscan intamldeosreitlu.xls
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
/home/uhlar/intamldeosreitlu.xls: OK
...
Infected files: 0

% clamscan -z intamldeosreitlu.xls
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
...
Infected files: 1

with clamdscan, problem is not detected until allscan option is used, after that it's detected:

% clamdscan /home/uhlar/intamldeosreitlu.xls
/home/uhlar/intamldeosreitlu.xls: OK
...
Infected files: 0

% clamdscan /home/uhlar/intamldeosreitlu.xls
/home/uhlar/intamldeosreitlu.xls: OK
...
Infected files: 0

% clamdscan -z /home/uhlar/intamldeosreitlu.xls
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
...
Infected files: 1

% clamdscan /home/uhlar/intamldeosreitlu.xls
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
/home/uhlar/intamldeosreitlu.xls: Doc.Downloader.Qbot03222-9942295-0 FOUND
...
Infected files: 2

with -z option I get 1 infected file, without it 2 infected...

[...]

Issue #442, opened 2022-01-27
Unofficial signature found, but infected files = 0


LSchuepbach:
Describe the bug
----------------

In some cases, a virus signature will hit, but another OK comes later, resulting in ClamAV saying the file isn't infected. Using the allmatch parameter will consider the file infected.

Am I doing anything wrong?

How to reproduce the problem
----------------------------

I've created this crude NDB signature with the goal of catching most excel 4.0 with macro:
`OLE.MACRO.XLS4:0:0,4:2E2B2C30{4-60}657863656C20342E30`

Testing it on a recent malware dropper: [Link in Virus Total](https://www.virustotal.com/gui/file/85da4a59bbf2d6320814c040dd5b564550aaa8ea4067e2329bce4a7033cde5d8)

Here's the output (note the infected file:0
```
laurent@computer:/tmp$ clamscan -d /tmp/sig.ndb /tmp/copy\ payment.xls 
/tmp/copy payment.xls: OLE.MACRO.XLS4.UNOFFICIAL FOUND
/tmp/copy payment.xls: OK

----------- SCAN SUMMARY -----------
Known viruses: 1
Engine version: 0.103.5
Scanned directories: 0
Scanned files: 1
Infected files: 0
Data scanned: 0.12 MB
Data read: 0.11 MB (ratio 1.11:1)
Time: 0.008 sec (0 m 0 s)
Start Date: 2022:01:27 09:06:05
End Date:   2022:01:27 09:06:05

```
Here's the same, with the allmatch parameter (-z), note the infected file: 1
```
laurent@computer:/tmp$ clamscan -zd /tmp/sig.ndb /tmp/copy\ payment.xls 
/tmp/copy payment.xls: OLE.MACRO.XLS4.UNOFFICIAL FOUND

----------- SCAN SUMMARY -----------
Known viruses: 1
Engine version: 0.103.5
Scanned directories: 0
Scanned files: 1
Infected files: 1
Data scanned: 0.33 MB
Data read: 0.11 MB (ratio 3.00:1)
Time: 0.012 sec (0 m 0 s)
Start Date: 2022:01:27 09:06:49
End Date:   2022:01:27 09:06:49

```

Attachments
-----------
Here's the output with debug :
```
clamscan -d /tmp/sig.ndb /tmp/copy\ payment.xls --debug 2> debug-noz.log
```
[debug-noz.log](https://github.com/Cisco-Talos/clamav/files/7948730/debug-noz.log)

[...]

PR #550, opened 2022-04-18
XLS parser: Fix issue where alert is lost

micahsnyder:
It is possible when not using allmatch mode that an alert on an XLS file
may be "lost". That is, it first reports as FOUND but then later reports
as OK.

The issue is essentially that the way allmatch and alert reporting is
done, it is easy to accidentally change the return value at one
intermediate layer and forget about the alert.

This fix doesn't clean up the design flaw, but does resolve this
specific bug.

Resolves:
- https://github.com/Cisco-Talos/clamav/issues/442
- https://github.com/Cisco-Talos/clamav/issues/521

See CLAM-1785

### DIFF
--- a/libclamav/scanners.c
+++ b/libclamav/scanners.c
@@ -362,8 +362,8 @@ static cl_error_t cli_scanrar_file(const char *filepath, int desc, cli_ctx *ctx)
             file_count += 1;
 
             /*
-            * Scan the metadata for the file in question since the content was clean, or we're running in all-match.
-            */
+             * Scan the metadata for the file in question since the content was clean, or we're running in all-match.
+             */
             status = cli_unrar_scanmetadata(&metadata, ctx, file_count);
             if ((status == CL_VIRUS) && SCAN_ALLMATCHES) {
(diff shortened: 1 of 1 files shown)


shutton (approved):
Well, that seems obvious.

micahsnyder:
Yup in most clamav scanner/parser functions, it relies on the `ret` variable to indicate `CL_VIRUS`, even if it also keeps track of the `num_viruses` separately for the "allmatch mode".  
One of my tasks for our next feature release to go through and change that so the "virus" detection is recorded by appending alerts to the scanning context instead of through return values, and then strip out all of the allmatch logic except for when appending the virus to determine if we should abort the scan by returning `CL_VIRUS` or continue the scan by returning `CL_SUCCESS`.  It will be nice!

micahsnyder:
The initial feature-test pipelines (internal in jenkins) failed on a couple on win7 and on a couple random tests.  Maybe a confict a 0.104 or 0.105 pipeline running at the same time?  I re-ran it and everything passed.

PR #551, opened 2022-04-18
XLS parser: Fix alert lost for image in XLS

micahsnyder:
The code to extract images from XLS documents isn't recording any alerts
that may be found when scanning those XLS documents. The logic to record
alerts was almost entirely done except left disconnected in one spot.

This fixes that broken link, propagating the alert all the way up.

Related to https://github.com/Cisco-Talos/clamav/pull/550, but not the same.
Will make a backport PR for 0.104.3.

### DIFF
--- a/libclamav/xlm_extract.c
+++ b/libclamav/xlm_extract.c
@@ -4633,10 +4633,11 @@ done:
 
 cl_error_t cli_extract_xlm_macros_and_images(const char *dir, cli_ctx *ctx, char *hash, uint32_t which)
 {
+    cl_error_t status = CL_SUCCESS;
+    cl_error_t ret;
     char fullname[PATH_MAX];
     int in_fd = -1, out_fd = -1;
     FILE *out_file = NULL;
-    cl_error_t ret = CL_SUCCESS;
     const char *opcode_name;
     char *tempfile = NULL;
     char *data     = NULL;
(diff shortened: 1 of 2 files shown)


libclamav/xlm_extract.c
```diff
@@ -5005,10 +5007,14 @@ cl_error_t cli_extract_xlm_macros_and_images(const char *dir, cli_ctx *ctx, char
          * A drawing group was extracted, now we need to find all the images inside.
          * If we fail to extract images, that's fine.
          */
-        (void)cli_extract_images_from_drawing_group(drawinggroup, drawinggroup_len, ctx);
+        ret = cli_extract_images_from_drawing_group(drawinggroup, drawinggroup_len, ctx);
```
ragusaa, line 5010: cli_extract_images_from_drawing_group can also return 'CL_EARG'.  Should we check for that error as well?

micahsnyder: I don't think i want to.   The can of worms there is that `cli_extract_images_from_drawing_group()`'s default return status code is `CL_EARG` and an oversight in that function has it returning the default error for a bunch of parsing-related (non-arg-related) failures. 

```c I don't want further parsing of the XLS document to be skipped because it failed to extract/scan an image from the drawing group.

ragusaa (approved):

micahsnyder:
Looking at Jenkins -- outside of the freebsd infrastructure issue we've been facing recently, and the expected failures in the custom build pipeline -- everything looks great. Thanks for the review, Andy. WIll merge.

PR #553, opened 2022-04-19
clam-1650 fixing leak when exporting pdf json metadata

m-sola:
Cleanup for the pdf->stats structure wasn't guaranteed when handling
a fatal memory error. This fix pulls stats cleanup into its own function
and handles the case where there's a fatal memory error by moving the
function call below the err: tag.

This fix also removes an extraneous call to pdf_export_json.


ragusaa:
There is a block of code in cli_pdf with an '#if 0 ... #endif' around it.  Could you remove that as part of the PR?

ragusaa:
Everything else looks good to me.  I reproduced the issue in main, and verified that it is gone with your changes, so I'll go ahead and approve it.

ragusaa (approved):

micahsnyder (approved):

micahsnyder:
Rebased to see it go through the test pipelines with the PoC in there.

PR #554, opened 2022-04-20
Clam 1693 undefined behavior

ragusaa:


micahsnyder (changes requested):

micahsnyder:
Rebased to pick up other fixes and see it go through the pipelines.

PR #556, opened 2022-04-21
libclamav: Fix typo in cl_engine_compile doc

ndrpnt:


libclamav/clamav.h
```diff
@@ -424,7 +424,7 @@ extern cl_error_t cl_engine_settings_free(struct cl_settings *settings);
 /**
  * @brief Prepare the scanning engine.
  *
- * Called this after all required databases have been loaded and settings have
+ * Called after all required databases have been loaded and settings have
```
micahsnyder, line 427: I think the typo fix for this should actually be:
```
 * Call this after all required databases have been loaded and settings have
 * been applied.
 ```

micahsnyder (approved):

PR #561, opened 2022-04-22
Zip parser: tolerate 2-byte overlap in file entries

micahsnyder:
The heuristic to alert on overlapping file entries is detecting some
non-malicious JAR files observed in critical enterprise software.
The goal with overlap detection is to alert on non-recursive zip-
bombs, so this tiny overlap isn't a concern.
We'll allow a 2-byte overlap so we don't alert on such zips.

Ref: CLAM-1763

### DIFF
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -1104,16 +1104,24 @@ cl_error_t index_the_central_directory(
             prev_record = &(zip_catalogue[index - 1]);
             curr_record = &(zip_catalogue[index]);
 
+            uint32_t prev_record_size = prev_record->local_header_size + prev_record->compressed_size;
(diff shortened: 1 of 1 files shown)


m-sola (commented):
Functionally, confirmed no more alert on provided test file.

I don't love how all of this is on one line. If it's not too much trouble could we break the logic up into something more readable?

micahsnyder:
> Functionally, confirmed no more alert on provided test file.
> 
> I don't love how all of this is on one line. If it's not too much trouble could we break the logic up into something more readable?

Ok I think I fixed it in a nice way.  Please re-review.

m-sola (approved):
Updates look good!

PR #563, opened 2022-04-25
330 - clamd becomes zombie process after some time

TairaSayo:
Issue happens when new container starts and databases are out of date.
The most likely reason is changes in database during clamd first initialization. 

This change would allow freshclam to finish updating the database, by moving its startup before clamd and introducing a delay after.

Defaulted to 60 seconds, controlled via CLAMD_STARTUP_DELAY variable.

### DIFF
--- a/dockerfiles/docker-entrypoint.sh
+++ b/dockerfiles/docker-entrypoint.sh
@@ -41,6 +41,19 @@ else
 		freshclam --foreground --stdout
 	fi
 
+	if [ "${CLAMAV_NO_FRESHCLAMD:-false}" != "true" ]; then
+		echo "Starting Freshclamd"
+		freshclam \
+		          --checks="${FRESHCLAM_CHECKS:-1}" \
+		          --daemon \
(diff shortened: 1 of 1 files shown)


micahsnyder:
@TairaSayo sorry about the delay in reviewing this.  I don't *think* there are any adverse effects to your PR.  

But I also do not think it solves the problem.  `clamd` may not run the container out-of-memory after the initial update and reload with your change -- because of the change of order of operations.  But the reload will still occur within 24 hours and so if the container does not have enough RAM for a concurrent reload, it may still cause `clamd` to become a zombie process. 

My understanding my not be correct.  If there's something I'm missing, you're welcome to propose the same or similar change over on our [new clamav-docker repository](https://github.com/Cisco-Talos/clamav-docker), as we have relocated the Docker-stuff there to make it easier to fix image issues for published versions, will [soon remove this Dockerfile.](https://github.com/Cisco-Talos/clamav/pull/764)

PR #583, opened 2022-05-13
Clam 1648 Fix benign heap over-read when loading some LDB signatures using byte-compare feature

ragusaa:
Previously 2 was hardcoded as as upper bound for loop in
cli_bcomp_freemeta.  Change to comp_count to avoid going
past the end.

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43816

### DIFF
--- a/libclamav/matcher-byte-comp.c
+++ b/libclamav/matcher-byte-comp.c
@@ -988,15 +988,14 @@ unsigned char *cli_bcomp_normalize_buffer(const unsigned char *buffer, uint32_t 
 void cli_bcomp_freemeta(struct cli_matcher *root, struct cli_bcomp_meta *bm)
 {
 
-    int i = 0;
+    uint32_t i = 0;
 
     if (!root || !bm) {
         return;
     }
 
-    /* can never have more than 2 */
     if (bm->comps) {
-        for (i = 0; i < 2; i++) {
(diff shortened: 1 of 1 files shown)


libclamav/matcher-byte-comp.c
```diff
@@ -988,15 +988,15 @@ unsigned char *cli_bcomp_normalize_buffer(const unsigned char *buffer, uint32_t
...
 {
 
-    int i = 0;
+    uint32_t i = 0;
 
     if (!root || !bm) {
         return;
     }
 
     /* can never have more than 2 */
```
micahsnyder, line 997: We should remove this comment, I guess.  But yeah looks like a simple fix. I'll do some basic testing.

micahsnyder:
Rebased this because it was failing the fuzz regression test from the latest fuzz fix, and wasn't completing the entire pipeline.

micahsnyder:
Rebased this because it was failing the fuzz regression test from the latest fuzz fix, and wasn't completing the entire pipeline.

micahsnyder:
Testing all looks good here.  I got a bit hung up on debugging a failure in the private fuzz corpus until I realized a PoC landed there before a fix, and that it didn't mean this was introducing a new bug or anything 😅.  All good. Will merge.

micahsnyder (approved):

PR #597, opened 2022-05-27
Changed freshclam warning threshold

ragusaa:
Changed freshclam warning threshold from 3 hours to 12 hours before
warning that the DNS entry is too old.


micahsnyder:
Looks good to me, and I think this change makes sense.  I want to confirm once more w/ Dave before changing it.

m-sola (approved):
Tested build for this, code looks good

PR #611, opened 2022-06-04
Clam 1643 cli_ac_addsig 2 byte over-read

micahsnyder:
- Fix possible 2-byte overread when adding sig pattern

  It is possible to create a signature pattern that tries to add a
  zero-byte matching pattern to the A-C trie. A missing check at this
  stage can end up with a 2-byte overread when indexing the (empty)
  pattern to make sure the bytes added to the A-C trie are static and
  not both zero.

  This over read issue is not a vulnerability.

  This commit fixes the issue by adding a check for the pattern length.

  Resolves: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43832

  Also added:
  - type casts and a "fall-through" comment to silence compile warnings.
  - a few additional length checks to protect against an additional 1-byte
  over read.

- Add code comments to explain AC pattern prefix process

  When adding a pattern to the AC trie, checks are done to make sure the
  bytes that go in the AC trie don't have any `?` wildcards and
  additionally that the first two bytes are not "\x00\x00".
  If they are, the position of the pattern that goes in the AC trie can be
  shifted right until a static pattern is identified that can go in the
  AC trie. Any bytes to the left of the new start of the pattern become a
  "prefix".

  During matching, once the AC trie match occurs and the bytes to the
  right of that pattern are matched, then the bytes from the prefix are
  matched.

  The reason that we don't want the bytes that go in the AC trie to start
  with "\x00\x00" is that it is such a common pattern in files that it
  would match constantly, and the scan process would spend a lot of time
  just checking through the list of patterns associated with a "\x00\x00"
  AC match, and that'd be crazy slow.
  But it is important to note that when shifting right, if there aren't
  enough nonzero, non-wildcard bytes to form a good prefix for the AC
  trie, that it is tolerable to bend the rule and let some patterns start
  with "\x00\x00". In that way, a small pattern like "0000ab" is still
  valid, and can be matched.

### DIFF
--- a/libclamav/matcher-ac.c
+++ b/libclamav/matcher-ac.c
@@ -769,6 +769,7 @@ int cli_ac_chklsig(const char *expr, const char *end, uint32_t *lsigcnt, unsigne
                     return -1;
                 }
                 pth--;
+                /* fall-through */
 
             case '>':
             case '<':
@@ -1102,9 +1103,9 @@ inline static int ac_findmatch_special(const unsigned char *buffer, uint32_t off
                     break;
                 subbp = bp;
             } else {
-                if (bp < (special->len[0] - 1))
+                if (bp < (uint32_t)(special->len[0] - 1))
                     break;
-                subbp = bp - (special->len[0] - 1);
+                subbp = bp - (uint32_t)(special->len[0] - 1);
             }
 
             match *= special->len[0];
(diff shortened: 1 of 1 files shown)


ragusaa (approved):

micahsnyder:
After rebase fuzz corpus test with the new sample passed.

PR #614, opened 2022-06-06
GH Actions: Newer version of run-vcpkg, older version of Windows

micahsnyder:
The 'windows-latest' image is now windows-2022, but the vcpkg we're
using doesn't know to look for visual studio 2022. Reverting to use
windows-2019 so vcpkg works with it.

Also switching to the run-vcpkg v7.4. Not going with the new v10 series
because that changes the inputs and I'm not ready to figure out how to
change the params to make it work.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -21,7 +21,7 @@ env:
 
 jobs:
   build-windows:
-    runs-on: windows-latest
+    runs-on: windows-2019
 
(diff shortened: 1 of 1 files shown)


micahsnyder:
For reference, this resolves the failures seen in any one of the recent PR windows builds. E.g.: https://github.com/Cisco-Talos/clamav/runs/6757891606?check_suite_focus=true

Issue #600, opened 2022-05-27
"Can't parse data ERROR" upon scanning simple png


hebelal:
Describe the bug
----------------

We are using ClamAV version 0.105.0 with default configuration.
Upon scanning certain files ClamAV throws the error “Can't parse data ERROR”.
One of those files creating this error is attached to this issue.
It’s a simple png file containing some photoshop meta-data which I suspect is the reason for the failure.

How to reproduce the problem
----------------------------

try to scan the attached file

Attachments
-----------

![error_original](https://user-images.githubusercontent.com/90389461/170700771-8381ce45-e429-40aa-b183-064b8a944050.png)

Nellox1:
We have the same problem. Had hard work to compile on Deb9 with cmake. And PNG files have problem to be spooled.
Hay news about the bug?

Thank you

micahsnyder:
Hi @Nellox1 no news, yet.  I will begin work on the fix soon.

tiiiecherle:
We have the same problem with quite a few jpg files.
Thanks for fixing in advance.

Issue #593, opened 2022-05-24
Some PDFs fail to scan with 0.105 (but work with 0.104)


alext:
Describe the bug
----------------

I'm experiencing a scanning failure for a specific PDF that only occurs with 0.105. The file scans clean when scanned with 0.104.

How to reproduce the problem
----------------------------

Running using the official docker images: `docker run -ti --rm -p 3310:3310 clamav/clamav:0.105`, and then scanning the file with clamdscan from the host: `clamdscan --config-file=./clamd.conf --stream file.pdf`.

Running against 0.105 reports:
```
/tmp/file.pdf: Can't parse data ERROR

----------- SCAN SUMMARY -----------
Infected files: 0
Total errors: 1
Time: 0.543 sec (0 m 0 s)
Start Date: 2022:05:24 09:57:30
End Date:   2022:05:24 09:57:31
```

Running against 0.104 reports:
```
/tmp/test.pdf: OK

----------- SCAN SUMMARY -----------
Infected files: 0
Time: 0.760 sec (0 m 0 s)
Start Date: 2022:05:24 10:16:09
End Date:   2022:05:24 10:16:10
```

I'm not able to provide the specific file as an attachment as it contains sensitive information, however, I've captured a debug log from the failed scan here:

https://gist.github.com/alext/a97a3ce43701ab64ab6ad35e821723f0

I'm not familiar with the ClamAV codebase, but one thing that stands out to me is the line:

```
LibClamAV debug: Failed to calculate image fuzzy hash for CL_TYPE_JPEG: Failed to load image: The decoder for Jpeg does not support the format features DNL
```

/ # clamd --version
ClamAV 0.105.0/26537/Tue May 10 08:08:04 2022

/ # clamconf -n
Checking configuration files in /etc/clamav

Config file: clamd.conf
-----------------------
LogFile = "/var/log/clamav/clamd.log"
LogTime = "yes"
PidFile = "/run/lock/clamd.pid"
LocalSocket = "/run/clamav/clamd.sock"
TCPSocket = "3310"
User = "clamav"

Config file: freshclam.conf
---------------------------
PidFile = "/run/lock/freshclam.pid"
UpdateLogFile = "/var/log/clamav/freshclam.log"
DatabaseMirror = "database.clamav.net"

Config file: clamav-milter.conf
-------------------------------
LogFile = "/var/log/clamav/milter.log"
LogTime = "yes"
PidFile = "/run/lock/clamav-milter.pid"
[...]

PR #618, opened 2022-06-23
Fix scan error when image fuzzy hash fails

micahsnyder:
Scanning a file containing an image where it fails to calculate the
image fuzzy hash may cause the entire scan to end with an error.

This commit fixes that by ignoring any image fuzzy hash failure.
If calculating the hash fails, the hash is left uninitialized (all
zeroes). To prevent doing image fuzzy hash lookups when it failed,
I added a boolean that is set to true only after creating the hash.

Fixes: https://github.com/Cisco-Talos/clamav/issues/600
Fixes: https://github.com/Cisco-Talos/clamav/issues/593

### DIFF
--- a/libclamav/scanners.c
+++ b/libclamav/scanners.c
@@ -4105,7 +4105,8 @@ static cl_error_t calculate_fuzzy_image_hash(cli_ctx *ctx, cli_file_t type)
         (void)cli_jsonstr(header, "Hash", hashstr);
     }
 
-    ctx->recursion_stack[ctx->recursion_level].image_fuzzy_hash = hash;
+    ctx->recursion_stack[ctx->recursion_level].image_fuzzy_hash            = hash;
+    ctx->recursion_stack[ctx->recursion_level].calculated_image_fuzzy_hash = true;
 
     status = CL_SUCCESS;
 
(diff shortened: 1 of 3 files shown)


libclamav/scanners.c
```diff
@@ -4652,47 +4653,76 @@ cl_error_t cli_magic_scan(cli_ctx *ctx, cli_file_t type)
              * Note: JPEG 2000 is a very different format from JPEG, JPEG/JFIF, JPEG/Exif, JPEG/SPIFF (1994, 1997)
              * JPEG 2000 is not handled by cli_scanjpeg or cli_parsejpeg.
              */
-            ret = calculate_fuzzy_image_hash(ctx, type);
+
+            // It's okay if it fails to calculate the fuzzy hash.
+            (void)calculate_fuzzy_image_hash(ctx, type);
```
m-sola, line 4658: Very minor, but should probably add a space between (void) and function name

m-sola (approved):
During initial tests of changes, it seemed like clamscan was getting caught in an infinite loop. Further inspection via gdb shows that there's an issue with cleanup, but that it's not an infinite loop.

```Reading symbols from ../bin/clamscan...
(gdb) run fuzzhashtest
Starting program: /home/mickey/clamav/bin/clamscan fuzzhashtest
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Loading:    19s, ETA:   0s [========================>]    8.62M/8.62M sigs
Compiling:   5s, ETA:   0s [========================>]       41/41 tasks

/home/mickey/clamav/test/fuzzhashtest: OK
^C
Program received signal SIGINT, Interrupt.
need_to_free_trans (root=<optimized out>, idx=<optimized out>) at ../libclamav/matcher-ac.c:670
670         for (j = 0; j < min; j++) {
(gdb) bt
#0  need_to_free_trans (root=<optimized out>, idx=<optimized out>) at ../libclamav/matcher-ac.c:670
#1  cli_ac_free (root=<optimized out>) at ../libclamav/matcher-ac.c:716
#2  0x00007ffff75ebbc1 in cl_engine_free (engine=0x61a000000c80) at ../libclamav/readdb.c:5674
#3  0x00000000004fbf61 in scanmanager (opts=0x6070000001e0) at ../clamscan/manager.c:1667
#4  0x00000000004fac3f in main (argc=<optimized out>, argv=<optimized out>) at ../clamscan/clamscan.c:171
(gdb) li
665         if (root->ac_nodes < idx) {
666             /*Should never happen, but check just to be safe.*/
667             min = root->ac_nodes;
668         }
669
670         for (j = 0; j < min; j++) {
671             if (NULL == root->ac_nodetable[j]) {
672                 continue;
673             }
674
(gdb) li
675             if (root->ac_nodetable[idx]->trans == root->ac_nodetable[j]->trans) {
676                 return 0;
677             }
678         }
679
680         return 1;
681     }
682
683     void cli_ac_free(struct cli_matcher *root)
684     {
(gdb) p min
$1 = 332427
(gdb) p j
$2 = 128777```

Let clamscan run for awhile to see the result (vs eicar) and found the same issue. Given the code changes, didn't seem related at all. A bit more testing (rolling back commits) confirmed that the issue was with recent AC trie changes.

The functional code changes in this commit are fine and do fix the error we've been seeing with 105 vs 103, so approving based on that.

libclamav/scanners.c
micahsnyder: I don't really have an opinion on this. But I just tested it and our clang-format settings refactor it to remove the space.

micahsnyder:
> During initial tests of changes, it seemed like clamscan was getting caught in an infinite loop. Further inspection via gdb shows that there's an issue with cleanup, but that it's not an infinite loop.
> ....
> Let clamscan run for awhile to see the result (vs eicar) and found the same issue. Given the code changes, didn't seem related at all. A bit more testing (rolling back commits) confirmed that the issue was with recent AC trie changes.

AHh yup the fix for CLAM-1607 (#530) introduced an issue with really slow unload times. CLAM-1995 is the ticket to fix that. It is something we need to fix soon.

PR #619, opened 2022-06-23
Silence error message when logical sig max flevel is low

micahsnyder:
libclamav will print this error message if you load a logical sig where
the max FLEVEL is less than the current flevel:

    LibClamAV Error: cli_loadldb: Failed to initialize target description block

This bug was introduced with this change:
https://github.com/Cisco-Talos/clamav/commit/fd587c741c0ca88d2a6493e9e85a2bf2453687ee#diff-1b118ae1a67627d562f94eacd4a44abe3a013db75e98405463ada73f65e65e41R2012-R2016


m-sola (approved):
Confirmed vs rel/105 that these changes no longer result in errors during dbload for sample preclass.cvd

PR #629, opened 2022-07-08
Add PDF object extraction recursion limits

micahsnyder:
Adds in object extraction recursion limits for object extraction as well
as well parsing string, array, and dictionaries during extraction.
The limit is set to 25.

### DIFF
--- a/libclamav/pdf.c
+++ b/libclamav/pdf.c
@@ -1439,6 +1439,11 @@ cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t
 
     cli_dbgmsg("pdf_extract_obj: obj %u %u\n", obj->id >> 8, obj->id & 0xff);
 
+    if (PDF_OBJECT_RECURSION_LIMIT < pdf->parse_recursion_depth) {
+        cli_dbgmsg("pdf_extract_obj: Recursion limit reached.\n");
+        return CL_SUCCESS;
+    }
+
     if (obj->objstm) {
         cli_dbgmsg("pdf_extract_obj: extracting obj found in objstm.\n");
         if (obj->objstm->streambuf == NULL) {
@@ -1582,10 +1587,13 @@ cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t
                 dict_len--;
             }
 
-            if (dict_len > 4)
+            if (dict_len > 4) {
+                pdf->parse_recursion_depth++;
                 dparams = pdf_parse_dict(pdf, obj, obj->size, (char *)pstr, NULL);
-            else
+                pdf->parse_recursion_depth--;
+            } else {
                 cli_dbgmsg("pdf_extract_obj: failed to locate DecodeParms dictionary start\n");
+            }
         }
 
         /*
@@ -3418,7 +3426,9 @@ static cl_error_t pdf_find_and_extract_objs(struct pdf_struct *pdf)
                 goto done;
             }
 
+            pdf->parse_recursion_depth++;
             status = pdf_extract_obj(pdf, obj, PDF_EXTRACT_OBJ_SCAN);
+            pdf->parse_recursion_depth--;
             switch (status) {
                 case CL_EFORMAT:
                     /* Don't halt on one bad object */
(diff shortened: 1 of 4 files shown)


libclamav/pdf.c
```diff
@@ -1440,6 +1440,11 @@ cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t
 
     cli_dbgmsg("pdf_extract_obj: obj %u %u\n", obj->id >> 8, obj->id & 0xff);
 
+    if (PDF_OBJECT_RECURSION_LIMIT < rec_level) {
```
shutton, line 1443: That's a very non-intuitive comparison.  I would've done
``` c
if (rec_level >= PDF_OBJECT_RECURSION_LIMIT)
```

libclamav/pdf.c
```diff
@@ -1429,7 +1429,7 @@ static int pdf_scan_contents(int fd, struct pdf_struct *pdf)
     return rc;
 }
 
-cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t flags)
+cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t flags, uint32_t rec_level)
```
shutton, line 1432: You could avoid this API change by just making this function a wrapper, and then call the new (static) function with the initial recursion depth.
``` c
cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t flags)
{
    return _pdf_extract_obj(pdf, obj, PRIVATE_STUFF);
}
```

You could further future-proof it by passing around a parsing context, which might be as small as this for now (but might track other resource-exhausting elements later):
``` c
typedef struct {
    int recursion_depth;
} PdfParseCxt;
```

It's small enough to pass by value. No need to allocate and pass by pointer.
``` c
cl_error_t pdf_extract_obj(struct pdf_struct *pdf, struct pdf_obj *obj, uint32_t flags, PdfParseCxt pcxt)
```

Also, I tend to think of (and see) this parameter as "depth" rather than "level".

libclamav/pdf.c
```diff
@@ -4136,7 +4141,7 @@ static void Author_cb(struct pdf_struct *pdf, struct pdf_obj *obj, struct pdfnam
         pdf->stats.author = cli_calloc(1, sizeof(struct pdf_stats_entry));
         if (!(pdf->stats.author))
             return;
-        pdf->stats.author->data = pdf_parse_string(pdf, obj, objstart, obj->size, "/Author", NULL, &(pdf->stats.author->meta));
+        pdf->stats.author->data = pdf_parse_string(pdf, obj, objstart, obj->size, "/Author", NULL, &(pdf->stats.author->meta), 0);
```
shutton, line 4144: You're effectively tracking two different types of recursion here.  Did you want to track a single parsing depth so that you know exactly how deep you are in the stack?

shutton (changes requested):

libclamav/pdf.c
micahsnyder: Yeah, the general goal is to limit parsing depth, starting with a top-level object extraction. But you're right there are basically two types of recursion:
- object extraction recursion
- object string, object dictionary, and object array recursion  (could be thought of as 3 types or generalized to one type)

The goal here is to mitigate a perceived risk in parsing where there are many layers of objects and/or dictionaries/arrays/strings. The `pdf_parse_string` function may identify objects and then parse & extract them, so presumably it could go object -> string -> object -> string, with perhaps dictionaries and arrays mixed in at any level.  

My approach is to just parse depth in general, which is why that `rec_level` (to be renamed `rec_depth`) is the same variable for both types of recursion.

libclamav/pdf.c
micahsnyder: Yeah I write a lot of [yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions).  I'm not ashamed 😅. Every now and then it'll catch a bug.

libclamav/pdf.c
micahsnyder: It's not really an API, but your point is a good one for any function that is very commonly used or for public API's.  
In this case I feel we're okay - `pdf_extract_obj()` and the `pdf_parse_*` functions in question aren't used very much and are purely internal to the pdf parser.  It's just in `pdf.h` because whoever extended the PDF parser didn't bother modularizing nicely with a `pdf_internal.h` file for things shared between the different C files. 👀 Pretty common design issue with a lot of the clamav parsers, honestly. Could definitely stand to be made into more tidy modules. 
Anyways, I see your point but I think we're okay on this one.

shutton (approved):

micahsnyder:
I squashed the fixup commits, rebased and resolved one minor merge conflict, and then added a commit to update the generated `sys.rs` file that was neglected in PR review from an unrelated PR.  I figure this one will merge very shortly (soon as I see it pass testing)... and I want to get that `sys.rs` update in.

PR #631, opened 2022-07-09
Fix BytecodeTimeout value in conf.sample

bbodenmiller:
Match code: https://github.com/Cisco-Talos/clamav/blob/cdbb4b8a4853e7431f527a613a79b04ac00acf01/common/optparser.c#L358-L359


micahsnyder:
Thanks for finding the discrepancy and fixing it in both the unix and windows config files!

micahsnyder (approved):

PR #634, opened 2022-07-15
Clam 2053 unrar 6.1.7

micahsnyder:
Upgrade from UnRAR 6.0.7 -> 6.1.7


micahsnyder:
Re: the new lgtm-com alerts - They're all in UnRAR (not our code), and all in print statements to stderr or stdout, so not critical.  I'm not going to worry about it.

PR #661, opened 2022-07-29
fix: Update bytecode timeout value in man pages

jarviliam:
Noticed that the man pages haven't been updated to include the new default bytecode timeout value 👍 .

- Update bytecode timeout values in the man pages

Signed-off-by: Liam Jarvis <jarviliam@gmail.com>


micahsnyder (approved):
Thanks @jarviliam

We somehow overlooked these with https://github.com/Cisco-Talos/clamav/pull/631
Your change looks good to me.

Issue #478, opened 2022-02-21
Container unable to run as non-root user


dasMulli:
The provided docker-built container contains user `clamav` (100) and group `clamav` (101), however when trying to run as this user/group, the image fails to start (same for current :stable and :latest tags):

```cmd
> docker run --rm -it --user 100:101 clamav/clamav:latest
Unable to find image 'clamav/clamav:latest' locally
latest: Pulling from clamav/clamav
Digest: sha256:8083453c1d23818ed281629943fe83318a77542ae07fc5113b6e70ff9dfda102
Status: Downloaded newer image for clamav/clamav:latest
install: can't create directory '/run/clamav': Permission denied
```

This is needed to run the container in Kubernetes as non-root container in order to pass stricter policies without additional need to work around them - e.g. creating PodSecurityPolicy objects to use. In the upcoming Pod Security Admissions feature, this container will be unable to be used in the Pod Security Standard "Restricted" ("Heavily restricted policy, following current Pod hardening best practices.") (see https://kubernetes.io/docs/concepts/security/pod-security-standards/).

Since I don't think ClamAV needs to do things that need root access or additional capabilities by default, being able to run as non-root user by default or having an unprivileged image would be good.

This also shows up as "HIGH" vulnerability/misconfiguration in some security scans (it's just a tool and good practices, not something to blindly make decisions on):
```sh
$ trivy repo --security-checks config https://github.com/Cisco-Talos/clamav
Enumerating objects: 52591, done.
Counting objects: 100% (52591/52591), done.
Compressing objects: 100% (25552/25552), done.
Total 52591 (delta 33169), reused 44462 (delta 26102), pack-reused 0
2022-02-21T15:48:29.073+0100    INFO    Detected config files: 1

Dockerfile (dockerfile)
=======================
Tests: 23 (SUCCESSES: 20, FAILURES: 3, EXCEPTIONS: 0)
Failures: 3 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 1, CRITICAL: 0)

+---------------------------+------------+----------------------------------+----------+------------------------------------------------+
|           TYPE            | MISCONF ID |              CHECK               | SEVERITY |                    MESSAGE                     |
+---------------------------+------------+----------------------------------+----------+------------------------------------------------+
| Dockerfile Security Check |   DS001    | ':latest' tag used               |  MEDIUM  | Specify a tag in the                           |
|                           |            |                                  |          | 'FROM' statement for image                     |
[...]

PR #666, opened 2022-08-02
ClamAV Docker Container running as non-root user

cmcdougall:
#478 

Moved the PID and socket file location to /tmp so the container can run as clamav user instead of root.

Updated docker-entrypoint.sh script to look for the PID files in /tmp.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -117,4 +117,4 @@ HEALTHCHECK --start-period=6m CMD "clamdcheck.sh"
 
 ENTRYPOINT [ "/init" ]
 
-USER clamav
+USER 100
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,8 +54,8 @@ RUN apk add --no-cache \
        "/clamav/usr/lib/pkgconfig/" \
     && \
     sed -e "s|^\(Example\)|\# \1|" \
-        -e "s|.*\(PidFile\) .*|\1 /run/lock/clamd.pid|" \
-        -e "s|.*\(LocalSocket\) .*|\1 /run/clamav/clamd.sock|" \
+        -e "s|.*\(PidFile\) .*|\1 /tmp/clamd.pid|" \
+        -e "s|.*\(LocalSocket\) .*|\1 /tmp/clamd.sock|" \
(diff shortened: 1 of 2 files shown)


micahsnyder:
This looks good at a glance. I will have to test it a bit.  And I would appreciate any other feedback/testing from others on it as well.

parmarnikunj:
when will it be available. We need it urgently.

Dockerfile
```diff
@@ -115,3 +116,5 @@ COPY "./dockerfiles/docker-entrypoint.sh" "/init"
 HEALTHCHECK --start-period=6m CMD "clamdcheck.sh"
 
 ENTRYPOINT [ "/init" ]
+
+USER clamav
```
duboisford, line 120: One of the use cases for a non-root clamav image is for deploying to kubernetes clusters that enforce `runAsNonRoot`. What I have experienced in testing this PR is the cluster I'm working with verifies if a user is non-root by looking at its UID and NOT the username. The specific error the cluster gives me is `container has runAsNonRoot and image has non-numeric user(clamav), cannot verify user is non-root`. I had to modify the above line to `USER 100` since `100` is the UID for user `clamav`. This allowed the container to spin up and work on the `runAsNonRoot` cluster I am working with. Is it possible to change this line to `USER 100` in this PR?

albundy83:
Why this merge has still not been approved ?

albundy83:
By the way, it will be nice to have this line removed, as it's already provided in `freshclam.conf`:
https://github.com/Cisco-Talos/clamav/blob/4a0382cf7a8e86e617b8f2e24f5ccb0e649f7f81/dockerfiles/docker-entrypoint.sh#L70

duboisford:
Is there any update on this pull request being merged in?

micahsnyder (approved):
I've done some testing with this and everything looks good to me.

micahsnyder:
Thank you @cmcdougall for your hard work on this.

micahsnyder:
> Thank you @cmcdougall for your hard work on this.

Sorry -- reverted due to issues presented in #478. Will continue discussion there and find a compromise.

PR #674, opened 2022-08-09
Clam 1594 fix allmatch Part 1 (Tests + append_virus logic change)

micahsnyder:
This PR adds tests to validate that allmatch-mode is working correctly. 
Some of the tests fail, because we have work to do to fix them.  Some pass.

This PR also adds the main fix to allmatch-mode to have the `cli_append_virus` function (and similar for "potentially unwanted" / "heuristic" alerts)  to only return `CL_VIRUS` if we need to stop scanning *now* (e.g. not in allmatch mode), else it will return `CL_SUCCESS`.  Alerts are stored as "indicators" in scan "evidence".  We can tally up the indicators at the end of a scan to determine the verdict, and to report the strong indicators before the PUA/Heuristic indicators, if required.

Note that I'm aiming to shift how we think of `cl_error_t` status as being separate from the scan "verdict".  `cl_errort_t` is a status code to indicate if something succeeded, or else if some event occurred (error or otherwise) that needs to be handled specially and in most cases means "end the scan".  This is different from the "verdict" which is only determined when the scan is complete. Since our C API does not distinguish between verdict and status, the two concepts are still mixed at the end of the scan when `scan_common()` returns. 

I also snuck a "bytecode debugging improvements" commit in here while investigating an allmatch failure.


ragusaa:
The changes look reasonable to me.  I do think, since this is a lot of code, that a regression run would be a good idea.

micahsnyder:
> The changes look reasonable to me. I do think, since this is a lot of code, that a regression run would be a good idea.

100%!  In its current state it doesn't pass all the tests.  This PR is just for "part 1" and will be merged into the `feature/all-match` branch.  I intend to do very extensive regression testing on the PR for that branch, when that's ready.

PR #676, opened 2022-08-11
Remove Docker registry URL

vienleidl:
The index.docker.io/library/ URL is not needed anymore. Now we can use the FROM alpine:latest and the result will be something like below:
```
latest: Pulling from library/alpine
Digest: sha256:bc41182d7ef5ffc53a40b044e725193bc10142a1243f395ee852a8d9730fc2ad
Status: Image is up to date for alpine:latest
docker.io/library/alpine:latest
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 # Copyright (C) 2021-2022 Cisco Systems, Inc. and/or its affiliates. All rights reserved.
 
 # hadolint ignore=DL3007  latest is the latest stable for alpine
-FROM index.docker.io/library/alpine:latest AS builder
+FROM alpine:latest AS builder
 
 WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


micahsnyder:
Thanks @vienleidl this seems like a reasonable change, and works for me. 
Does it fix an issue or is it just more modern?
@oliv3r do you have any concerns?

micahsnyder (approved):

micahsnyder:
@vienleidl did you have a reason why the change is needed?  Given @oliv3r's feedback and without a usecase that demonstrates a need, I will have to close this PR.

micahsnyder:
Closing for now.  We can revisit this later if a need arises.

Side note, the docker stuff is going to be moving to https://github.com/Cisco-Talos/clamav-docker soon

oliv3r:
> Side note, the docker stuff is going to be moving to https://github.com/Cisco-Talos/clamav-docker soon

why is that? Having the dockerfile here, means we can build from source on source changes. I can def. see the need for a clamav-docker however, where that repo would pull the tagged build from here, and publish it into an 'official' space. So just curious as to why you'd want to move it?

micahsnyder:
We've had a few needs to edit the Dockerfile or docker scripts without wanting to make a new clamav patch version.  At present we're blocked on this because Docker Hub removed their v1 API.  But there have been and will likely be other needs as well.

PR #678, opened 2022-08-15
ccmake: make the values of BYTECODE_RUNTIME enumeratable by ENTER

dilyanpalauzov:
When clamav is configured with `ccmake`, after this change at the line with BYTECODE_RUNTIME the user can just press the ENTER key to toggle the available values.


micahsnyder (approved):
Thanks for the PR @dilyanpalauzov this seems like a nice improvement for anyone using ccmake!

PR #684, opened 2022-08-24
Clam 1594 fix allmatch remove checks (part 2)

micahsnyder:
This is the final part of the alllmatch overhaul:
1. Removes checks for `SCAN_ALLMATCHES`
2. Makes sure to stop scanning when scan return code is `!= CL_SUCCESS`, instead of only when `CL_VIRUS` or `CL_BREAK` or whatever - to make sure we stop scanning when exceeding scan limits.
3. Some code cleanup and some swap from use of `CL_CLEAN` -> `CL_SUCCESS` to try to differentiate the purpose of `cl_error_t` from the scan verdict.  The `cl_error_t` is to indicate if something succeeded or failed, and if it failed most likely we should abort the scan with very few exceptions. Note most file parsing errors should not be fatal to the scan, so catching non-fatal errors or even just debug-logging them and moving on is the way to go.  
4. Adds some additional tests and restructure the `clamscan` tests into separate files because `clamscan_test.py` was getting way too long. Also fixed the clamscan `.ign2` test. 
5. Makes sure that `magic_scan()` only returns errors that should halt the scan, and makes sure that all calls to `magic_scan()` propagate errors reported by `magic_scan()`.  Many used to only propagate the `CL_VIRUS` return, but that would ignore critical errors like `CL_EMEM` and `CL_ETIMEOUT`.  In addition, the functions that originate `CL_VIRUS` and `CL_ETIMOUT` now also set a `ctx->abort_scan` flag which is checked in `magic_scan()` to ensure that scans terminate even if the return code was somehow lost.

For code review, It may be best to review it commit-by-commit in the https://github.com/Cisco-Talos/clamav/pull/684/commits page.  I tried to keep each commit on separate files, or at least separate issues so that none is incomplete on its own.


micahsnyder:
Hmm uh oh, merge conflicts. Not sure how this happened.  I will try to fix it. 

My force push a moment ago was because the `.idb` icon database test files didn't get added in the pe-allmatch test commit because they are apparently ignored by the .gitignore file (conflicting file extension) and it caused test failures.

micahsnyder:
Merge conflict fixed.  There wasn't a real merge conflict, just the hashes for the previous commits were different because GitHub's rebase+merge isn't a fast-forward merge when a rebase isn't required.  It still does the rebase, which alters the hash even though it accomplishes nothing. Well done, GitHub, well done.

micahsnyder:
Oops, the `test.exe` was apparently also missing because of the `.gitignore`.  Added that and did another force-push.

micahsnyder:
Another force-push, fixed the issues identified by the lgtm-com app. It was memcpy'ing to the wrong hash array (md5 instead of sha1 or sha256).

micahsnyder:
I believe that the allmatch overhaul is complete with this PR (now).  

We can probably start regression testing with this branch, although I do intend to do a rebase to squash one of the commits from this PR into one from the `feature/fix-allmatch` branch before doing another PR from `feature/fix-allmatch` over to `main`.

libclamav/asn1.c
```diff
@@ -1620,7 +1620,7 @@ static cl_error_t asn1_parse_mscat(struct cl_engine *engine, fmap_t *map, size_t
                         cli_dbgmsg("asn1_parse_mscat: Found Authenticode certificate blocked by %s\n", crt->name ? crt->name : "(unnamed CRB rule)");
                         if (NULL != ctx) {
                             ret = cli_append_virus(ctx, crt->name ? crt->name : "(unnamed CRB rule)");
-                            if ((ret == CL_VIRUS) && !SCAN_ALLMATCHES) {
+                            if (ret == CL_VIRUS) {
                                 crtmgr_free(&newcerts);
```
ragusaa, line 1624: There are a couple places in this function with this particular free.  Do you think it's worth it to move this to the finish block so that it is only done in one place?

micahsnyder: Sure can do. 

I'm afraid of doing the clean up on the RTF parser, though. 😅

micahsnyder:
Merged into the feature branch for extended testing and cleanup.  Additional review to continue in https://github.com/Cisco-Talos/clamav/pull/687

PR #686, opened 2022-08-25
Add missing MaxScanTime heuristic alert, when enabled

micahsnyder:
It appears that exceeding the max scan time may not trigger an
heuristic allert when AlertExceedsMax is enabled. This Adds that, and
sets the flag that should ensure we fully abort the scan, like we
have with MaxScanFiles.


micahsnyder:
I'm canceling this PR and mixing it into the code I'm working on.  I have other cases where I want to set/check `ctx->abort_scan` to guarantee that a scan is terminated when a limit is exceeded or an alert occurs and we're not in allmatch mode.

PR #687, opened 2022-08-26
Feature/fix allmatch

micahsnyder:
Overhaul of the allmatch system.  In allmatch mode, CL_VIRUS not returned until the very end. Removes need for allmatch checks scattered throughout the codebase.


ragusaa:
I put a comment in CLAM-2083

micahsnyder:
Just did a rebase to get things up to date, to resolve the merge conflicts and squash the commits that needed squashing.

micahsnyder:
I found that the section offset commit increased ram usage by a bit (50-ish MB?) and also increased load time by a lot (60 seconds give or take). 
So I removed the section offset fix commit and moved it to a separate branch for further investigation: https://github.com/micahsnyder/clamav-micah/tree/CLAM-2094-section-offset-acpatt

micahsnyder:
I just removed the "reduce RAM usage" commit and moved it to a separate branch so we can test and review it separate.  I found that it had some false positive issues relating to offset sigs during regression testing: https://github.com/micahsnyder/clamav-micah/tree/CLAM-2128-reduce-ac-patt-mem-usage

Since I found a different way to reduce RAM usage in this PR back to normal levels (moving the section offset bug fix to a separate branch as well), this change not as important to merge at the same time.

libclamav/7z_iface.c
```diff
@@ -189,18 +184,19 @@ int cli_7unz(cli_ctx *ctx, size_t offset)
...
 
                 cli_dbgmsg("cli_7unz: Saving to %s\n", tmp_name);
-                if (cli_writen(fd, outBuffer + offset, outSizeProcessed) != outSizeProcessed)
+                if (cli_writen(fd, outBuffer + offset, outSizeProcessed) != outSizeProcessed) {
                     found = CL_EWRITE;
-                else if ((found = cli_magic_scan_desc(fd, tmp_name, ctx, name)) == CL_VIRUS)
-                    viruses_found++;
+                }
+
+                found = cli_magic_scan_desc(fd, tmp_name, ctx, name);
```
micahsnyder, line 191: It looks like we may still want an `else {}` around this. I don't know why I removed it.

libclamav/filetypes.c
```diff
@@ -277,7 +277,8 @@ cli_file_t cli_determine_fmap_type(fmap_t *map, const struct cl_engine *engine,
     unsigned char buffer[MAGIC_BUFFER_SIZE];
     const unsigned char *buff;
     unsigned char *decoded;
-    int bread, sret;
+    int bread;
+    cli_file_t sret;
```
micahsnyder, line 281: The lines 303 -> 308 (below) should use a different variable than `sret` that IS an `int`, because `cli_memcpy` returns an int. May also be worth considering having it return a bool ... but would have to evaluate where else `cli_memcpy` is used.:
```c
        sret = cli_memcpy(buffer, buff, bread);
        if (sret) {
            cli_errmsg("cli_determine_fmap_type: fileread error!\n");
            return CL_TYPE_ERROR;
        }
        sret = 0;
```

libclamav/fmap.c
```diff
@@ -1082,49 +1086,155 @@ extern void cl_fmap_close(cl_fmap_t *map)
...
-        if (!(hashctx)) {
-            cli_errmsg("fmap_get_MD5: error initializing new md5 hash!\n");
+    switch (type) {
+        case CLI_HASH_MD5:
+            if (map->have_md5) {
+                goto complete;
+            }
+            break;
+        case CLI_HASH_SHA1:
+            if (map->have_md5) {
```
micahsnyder, line 1142: should be `map->have_sha1`

libclamav/fmap.c
```diff
@@ -1082,49 +1086,155 @@ extern void cl_fmap_close(cl_fmap_t *map)
...
+                goto complete;
+            }
+            break;
+        case CLI_HASH_SHA256:
+            if (map->have_md5) {
```
micahsnyder, line 1147: should be `map->have_sha256`

libclamav/hfsplus.c
```diff
@@ -1300,7 +1349,8 @@ static cl_error_t hfsplus_walk_catalog(cli_ctx *ctx, hfsPlusVolumeHeader *volHea
 
                             if (!ctx->engine->keeptmp) {
                                 if (cli_unlink(resourceFile)) {
-                                    ret = CL_EUNLINK;
+                                    status = CL_EUNLINK;
+                                    goto done;
                                 }
                             }
                             free(resourceFile);
```
micahsnyder, line 1356: Must set to NULL after free'ing.

micahsnyder:
Just rebased with main so I could re-build in the regression test system after a failed build.

libclamav/7z_iface.c
micahsnyder: Under further consideration, I think this is so we can at least scan partial writes... I don't know that it really matters.  I guess I'll keep this change.

micahsnyder:
Just rebased with main. That is all

ragusaa (approved):

micahsnyder:
Rebased and fixed massive merge conflict caused by archive inspection PR.

PR #690, opened 2022-08-30
Clam 2102 cl cvd unpack

micahsnyder:
- libclamav API: Add cl_cvdunpack() function

  Add `cl_cvdunpack()` function to the public API.
  
  This new API has an option to disable verification, but otherwise it
  will attempt to verify that the CVD is correctly signed.

- Freshclam, Sigtool: use public CVD unpack API

  In the interest of using the public API's as much as possible for our
  own applications (dog-fooding the API), this commit swaps sigtool and
  freshclam `cli_cvdunpack()` calls to `cl_cvdunpack()`.

- Tests: unit tests for cl_load(), cl_cvdverify(), cl_cvdunpack()

  Some basic testing is needed for the new cl_cvdunpack() API, so this
  commit adds basic unit tests for that.
  
  For reasons unknown, a number of cl_* API's have stubs for unit tests
  that weren't filled out.  The CVD load/verify ones in particular
  required access to a signed CVD.  We actually ship a very basic signed
  CVD with the databases now, so I added tests for those while I was at it.

### DIFF
--- a/libclamav/cvd.c
+++ b/libclamav/cvd.c
@@ -723,7 +723,7 @@ int cli_cvdload(FILE *fs, struct cl_engine *engine, unsigned int *signo, unsigne
     return ret;
 }
 
-int cli_cvdunpack(const char *file, const char *dir)
+static cl_error_t cli_cvdunpack(const char *file, const char *dir)
 {
     int fd, ret;
 
--- a/libclamav/cvd.c
+++ b/libclamav/cvd.c
@@ -611,11 +611,11 @@ cl_error_t cl_cvdverify(const char *file)
     return ret;
 }
 
-int cli_cvdload(FILE *fs, struct cl_engine *engine, unsigned int *signo, unsigned int options, unsigned int dbtype, const char *filename, unsigned int chkonly)
+cl_error_t cli_cvdload(FILE *fs, struct cl_engine *engine, unsigned int *signo, unsigned int options, unsigned int dbtype, const char *filename, unsigned int chkonly)
 {
     struct cl_cvd cvd, dupcvd;
     FILE *dupfs;
-    int ret;
+    cl_error_t ret;
     time_t s_time;
     int cfd;
     struct cli_dbio dbio;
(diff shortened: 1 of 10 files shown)


libfreshclam/libfreshclam_internal.c
```diff
@@ -1498,9 +1498,12 @@ static fc_error_t getcvd(
 }
 
 /**
- * @brief Change to the temp dir for storing CDIFFs for incremental database update.
+ * @brief Create a temp dir for storing CDIFFs for incremental database update.
  *
- * Will create the temp dir if it does not already exist.
+ * Will create the temp dir if it does not already exist and populate it with the
+ * unpacked CVD. Then i twill chdir to that directory.
```
TheRaynMan, line 1504: "it will chdir" vs "i twill chdir"

libclamav/cvd.c
```diff
@@ -732,3 +740,37 @@ int cli_cvdunpack(const char *file, const char *dir)
...
 }
+
+cl_error_t cl_cvdunpack(const char *file, const char *dir, bool dont_verify)
+{
+    cl_error_t status = CL_SUCCESS;
+    FILE *fs          = NULL;
+
+    fs = fopen(file, "rb");
+    if (NULL == fs) {
+        char err[128];
```
TheRaynMan, line 751: Why does this cli_strerror call use err[128] when the others do not? This is not a static function or anything.

micahsnyder: Our `cli_strerror` is a wrapper for the poorly standardized `strerror` system function, not to be confused with `cl_strerror`, which is the clamav API for converting the clam error enum to a message. 

The `err[128]` approach is something I copypasted from elsewhere in the code.  Such as: https://github.com/Cisco-Talos/clamav/blob/main/libclamav/untar.c#L318-L319

libfreshclam/libfreshclam_internal.c
micahsnyder: I'll fix it

micahsnyder: Fixed in force-push

micahsnyder:
Per our discussion today in status -- I'll create an example program that uses this.  I may be able to create an automated test for it as well.  It would be good if we tested our example programs, after all.

micahsnyder:
Rebased and also fixed a PATHSEP issue on Windows with the new unit test.

Issue #574, opened 2022-05-08
“The newest LLVM version supported is 3.6.2” and “it can use LLVM versions 8 through 12”.


dilyanpalauzov:
The file [NEWS.md](https://github.com/Cisco-Talos/clamav/blob/main/NEWS.md#notable-changes) contains:
> Updated the LLVM bytecode runtime support so that it can use LLVM versions 8 through 12 and removed support for earlier LLVM versions.

while [INSTALL.md](https://github.com/Cisco-Talos/clamav/blob/main/INSTALL.md#known-issues--to-dos) states:
> The newest LLVM version supported is 3.6.2. We ran out of time during 0.104 development to add support for newer versions of LLVM. 

The information in both files is mutually exclusive.

teoberi:
Works even with LLVM 13.0.1

micahsnyder:
Thanks for the heads up.  We totally forgot to remove the "The newest LLVM version supported is 3.6.2..." line.

wschlich:
When will LLVM 14 or 15 be supported?

teoberi:
https://github.com/Cisco-Talos/clamav/issues/580#issuecomment-1188054430

wschlich:
@teoberi I can't see any date related estimates for LLVM 14+ support in issue #580 🤷🏼‍♂️

teoberi:
I don't even think it exists yet!

wschlich:
@teoberi maybe I misinterpreted your previous comment, just mentioning another issue right after I asked for a date related estimate 🙃

PR #692, opened 2022-09-07
Implement CMake LLVM version overrides; add hard MAX and MIN versions

### LINKED ISSUE #691: 105.1: CMake LLVM overrides are broken
Hi Team,

I've been attempting to package 105.1 for Gentoo where we have LLVM 13 - 15 in the tree: gentoo/gentoo/pull/27153

Based on #574 it seems that LLVM 13 is a sane choice here, however when attempting to build CMake insists on finding LLVM 14.

Based on [your docs](https://github.com/Cisco-Talos/clamav/blob/main/INSTALL.md#llvm-optional-see-bytecode-runtime-section) I've attempted to override that using both ebuild logic and hardcoded pointers at different versions of LLVM, however even hardcoding paths CMake still insists on using LLVM 14:

```ebuild
		-DLLVM_ROOT_DIR="/usr/lib/llvm/13/"
		-DLLVM_FIND_VERSION="13.0.1"
```

Having taken a look over the source, it appears that perhaps the feature was documented before it was implemented?

https://github.com/Cisco-Talos/clamav/blob/4a0382cf7a8e86e617b8f2e24f5ccb0e649f7f81/CMakeLists.txt#L484-L517

I want to get this working for Gentoo (at least to validate the ebuild machinery) so that it works when you do support more modern versions of LLVM; I'll see about writing a patch and also including a 'max version check' to resolve this.

Kangie:
Fixes #691

- Move minimum LLVM version to variable; add max version
- Implement `-DLLVM_ROOT_DIR` and `-DLLVM_FIND_VERSION`
- If the user has set a particular version and CMake can't find it; die
- Intentionally fail if LLVM version is unsupported
- Made LLVM detection messages `STATUS` rather than `NOTICE`

`LLVM_MAX_VER` set to `13.0.1` based on #574.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,8 +491,8 @@ find_package(JSONC REQUIRED)
 # Set variable required by libclamav to use libjson-c
 set(HAVE_JSON 1)
 
-set(LLVM_MAX_VER "14.0.0")
-set(LLVM_MIN_VER "8.0.0")
+set(LLVM_MAX_VER "13")
+set(LLVM_MIN_VER "8")
 
 string (TOLOWER ${BYTECODE_RUNTIME} bytecodeRuntime)
 if(${bytecodeRuntime} STREQUAL "llvm")
@@ -519,13 +519,16 @@ if(${bytecodeRuntime} STREQUAL "llvm")
             endif()
         endif()
 
-        if (${LLVM_PACKAGE_VERSION} VERSION_LESS ${LLVM_MIN_VER})
+        math(EXPR TOO_HIGH_VERSION "${LLVM_MAX_VER} + 1" OUTPUT_FORMAT DECIMAL)
+
+        if(${LLVM_PACKAGE_VERSION} VERSION_LESS ${LLVM_MIN_VER})
             message(FATAL_ERROR "LLVM version ${LLVM_PACKAGE_VERSION} is too old")
-        elseif (${LLVM_PACKAGE_VERSION} VERSION_GREATER_EQUAL ${LLVM_MAX_VER} )
+        elseif (${LLVM_PACKAGE_VERSION} VERSION_GREATER_EQUAL ${TOO_HIGH_VERSION} )
             message(FATAL_ERROR "LLVM version ${LLVM_PACKAGE_VERSION} is too new")
         else()
             message(STATUS "LLVM version ${LLVM_PACKAGE_VERSION} found")
         endif()
+
         # Set variable required by libclamav to use llvm instead of interpreter
         set(LLVM_VERSION ${LLVM_VERSION_MAJOR}${LLVM_VERSION_MINOR})
     endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,39 +491,43 @@ find_package(JSONC REQUIRED)
 # Set variable required by libclamav to use libjson-c
 set(HAVE_JSON 1)
 
+set(LLVM_MAX_VER "14.0.0")
+set(LLVM_MIN_VER "8.0.0")
+
 string (TOLOWER ${BYTECODE_RUNTIME} bytecodeRuntime)
 if(${bytecodeRuntime} STREQUAL "llvm")
-    set (LLVM_FIND_VERSION "8.0.0")
-    find_package(LLVM REQUIRED)
+    if(DEFINED LLVM_ROOT_DIR AND DEFINED LLVM_FIND_VERSION)
+        find_package(LLVM EXACT ${LLVM_FIND_VERSION} REQUIRED HINTS ${LLVM_ROOT_DIR})
+    elseif(DEFINED LLVM_ROOT_DIR)
+        find_package(LLVM REQUIRED HINTS ${LLVM_ROOT_DIR})
+    elseif(DEFINED LLVM_FIND_VERSION)
+        find_package(LLVM EXACT ${LLVM_FIND_VERSION} REQUIRED)
+    else()
+        set (LLVM_FIND_VERSION ${LLVM_MIN_VER})
+        find_package(LLVM REQUIRED)
+    endif()
     if(LLVM_FOUND)
         if (LLVM_AVAILABLE_LIBS)
-            # Found using LLVMConfig.cmake
-            message("LLVM found using LLVMConfig.cmake")
-            set(LLVM_VERSION ${LLVM_VERSION_MAJOR}${LLVM_VERSION_MINOR})
+            message(STATUS "LLVM found using LLVMConfig.cmake")
             set(LLVM_LIBRARIES ${LLVM_AVAILABLE_LIBS})
-
-            if (${LLVM_PACKAGE_VERSION} VERSION_LESS "8.0.0")
-                message(FATAL "LLVM version ${LLVM_PACKAGE_VERSION} is too old")
-            endif()
-
         else()
-            # Found using FindLLVM.cmake
-            message("LLVM found using FindLLVM.cmake")
+            message(STATUS "LLVM found using FindLLVM.cmake")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -480,39 +480,56 @@ find_package(JSONC REQUIRED)
...
             endif()
 
-            if (${LLVM_VERSION_STRING} VERSION_LESS "8.0.0")
-                message(FATAL "LLVM version ${LLVM_VERSION_STRING} is too old")
+            if (${LLVM_VERSION_STRING} VERSION_LESS ${LLVM_MIN_VER})
+                message(FATAL_ERROR "LLVM version ${LLVM_VERSION_STRING} is too cold")
+            elseif (${LLVM_VERSION_STRING} VERSION_GREATER ${LLVM_MAX_VER} )
+                message(FATAL_ERROR "LLVM version ${LLVM_VERSION_STRING} is too hot")
+            else()
+                message(STATUS "LLVM version ${LLVM_VERSION_STRING} is just right")
```
micahsnyder, line 527: This code is getting pretty verbose.  To reduce duplication, can you please do something like
```cmake
set(LLVM_PACKAGE_VERSION ${LLVM_VERSION_STRING})
```
and then move the cold/hot/right checks outside of the if/else to check just the LLVM_PACKAGE_VERSION? 

I would also request that you change "cold"/"hot"/"is just right" to "old"/"new"/"found".  I think the three bears lines are very cute and it made me smile. But I prefer to be literal to reduce questions from users.

CMakeLists.txt
```diff
@@ -480,39 +480,56 @@ find_package(JSONC REQUIRED)
...
             endif()
         endif()
+
+        # Set variable required by libclamav to use llvm instead of interpreter
+        set(LLVM_VERSION ${LLVM_VERSION_MAJOR}${LLVM_VERSION_MINOR})
```
micahsnyder, line 532: Good call de-duplicating this one, though! 👍

CMakeLists.txt
```diff
@@ -480,39 +480,56 @@ find_package(JSONC REQUIRED)
...
                 if (${LLVM_VERSION_STRING} VERSION_LESS "10.0.0")
                     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNDEBUG")
                 endif()
             endif()
 
-            if (${LLVM_VERSION_STRING} VERSION_LESS "8.0.0")
-                message(FATAL "LLVM version ${LLVM_VERSION_STRING} is too old")
+            if (${LLVM_VERSION_STRING} VERSION_LESS ${LLVM_MIN_VER})
+                message(FATAL_ERROR "LLVM version ${LLVM_VERSION_STRING} is too cold")
+            elseif (${LLVM_VERSION_STRING} VERSION_GREATER ${LLVM_MAX_VER} )
```
micahsnyder, line 524: Instead of checking if `VERSION_GREATER 13.0.1`, let's check if `VERSION_GREATER_EQUAL 14.0.0`.  Presumably we'll suppost 13.0.2 if there is a security patch version, and we'll want any users on an older release to adopt it.

micahsnyder (changes requested):
Thanks for the PR.  Nice catch with info -> STATUS and "FATAL" -> "FATAL_ERROR". 

I haven't tested this yet but it reads well.  I just have a few changes to request.

Kangie:
Feedback addressed - I appreciate the prompt response.

Seems fine from my end. I don't have the infra available right now to test against versions of LLVM that need `FindLLVM.cmake`, but your suggestion to deduplicate that code seems fine.

Due to the use of `VERSION_GREATER_EQUAL` I've removed the conditional for CMake versioning and set that to `3.7` across the board. It's from 2016; if someone is running a system with an older version of CMake they'll have to update.

I note that `LLVM_FIND_VERSION` doesn't really do much on its own - it doesn't provide any hints to CMake about which version of LLVM it should find and doesn't override CMake's detection of LLVM versions. Instead this change just just forces an error if what you specify isn't what CMake finds (which appears to always be the highest version without `LLVM_ROOT_DIR` being set).

I feel like this is desirable behaviour, even if it just encourages users to set `LLVM_ROOT_DIR` if they're on a multi-impl system; If I specify `13.0.1` and `14.0.x` or `12.x.x` is found I'd rather that the configure process error instead of 

This shouldn't impact most users though - it's unusual to see multi-impl LLVM on distros outside of Gentoo anyway, and the ebuild for clamav there has been updated to set both `-DLLVM_FIND_VERSION` and `-DLLVM_ROOT_DIR`.

micahsnyder:
> ~Due to the use of `VERSION_GREATER_EQUAL` I've removed the conditional for CMake versioning and set that to `3.7` across the board. It's from 2016; if someone is running a system with an older version of CMake they'll have to update. Arguably this should have been done long before me; that check was already in use.~
> 
> Someone on your team will need to review your CMakelists files (or ideally migrate to something more sane, like Meson). The minimum version for CMake across the board should be at least `3.7` based on the use of `VERSION_GREATER_EQUAL` in code before I made changes. Changing this breaks (at the very least) `libclamunrar_iface`. </3 CMake

The minimum CMake versions for ClamAV are already 3.16 (windows) and 3.14 (others) https://github.com/Cisco-Talos/clamav/blob/main/CMakeLists.txt#L4-L8  Using a feature from 3.7 shouldn't be a problem. 

> I note that `LLVM_FIND_VERSION` doesn't really do much on its own - it doesn't provide any hints to CMake about which version of LLVM it should find and doesn't override CMake's detection of LLVM versions. Instead this change just just forces an error if what you specify isn't what CMake finds (which appears to always be the highest version without `LLVM_ROOT_DIR` being set).
> 
> I feel like this is desirable behaviour, even if it just encourages users to set `LLVM_ROOT_DIR` if they're on a multi-impl system; If I specify `13.0.1` and `14.0.x` or `12.x.x` is found I'd rather that the configure process error instead of compile against a version that's not the one that I explicitly specified.
> 
> This shouldn't impact most users though - it's unusual to see multi-impl LLVM on distros outside of Gentoo anyway, and the ebuild for clamav there has been updated to set both `-DLLVM_FIND_VERSION` and `-DLLVM_ROOT_DIR`.

Our `FindLLVM.cmake` originates from https://github.com/ldc-developers/ldc/blob/master/cmake/Modules/FindLLVM.cmake.  I'm happy to make improvements on our end.  Ideally we'd share them back with the `ldc` project as well.   If you'd like to work on making `LLVM_FIND_VERSION` work better, we'd be happy to review your work. 

In case it interests you, there is another/newer mechanism for finding LLVM using LLVMConfig.cmake files installed to under `/usr/lib/llvm-10/cmake/LLVMConfig.cmake` (for example).  See https://github.com/Cisco-Talos/clamav/blob/main/INSTALL.md?plain=1#L711-L737 for more details.

Kangie:
> The minimum CMake versions for ClamAV are already 3.16 (windows) and 3.14 (others)

I looked at that for far too long yesterday. Those were the strings I replaced. Why I parsed 3.16 as < 3.7 I will never know. :)

micahsnyder:
Is this ready for re-review?

Kangie:
Should be!

While I haven't touched `LLVM_FIND_VERSION` or `FindLLVM.cmake` I'm happy enough that if a particular version of LLVM is specified and it's not what CMake finds that the configure process intentionally fails; `ldc` [seems to agree](https://github.com/ldc-developers/ldc/blob/master/cmake/Modules/FindLLVM.cmake#L75) that setting `LLVM_ROOT_DIR` is appropriate in that case.

Additionally it's probably only an issue on multi-impl systems where it appears that the highest version is preferred regardless of what's specified.

I've had this patch applied to the Gentoo `clamav` package for a few weeks with no reported issues. Happy to make any changes/improvements or address additional feedback!

ragusaa (approved):

micahsnyder (approved):

PR #701, opened 2022-09-20
CMake+Rust: default build must be release with debug info

micahsnyder:
We already set the default build type to RelWithDebInfo for CMake, but we were setting it *after* adding the Rust module.  We need to do it before, or else the Rust stuff will still default to Debug, which makes for really slow scans of images that get fuzzy hashed.


shutton (approved):
LGTM

PR #712, opened 2022-10-04
Speed up freeing of signatures on shutdown

ragusaa:


micahsnyder (approved):
Code looks great. Local performance testing looks great. QA pipelines look great. 
Excellent!

PR #714, opened 2022-10-07
CMake: fix libclamav-only mode build when tests are enable

micahsnyder:
CMake's configuration step fails if you use ENABLE_LIBCLAMAV_ONLY=ON and ENABLE_TESTS=ON. This is because we try to add the libcurl module path to the test environment, for the sake of linking with the Rust unit test program, as we do with the other dependencies. However, libcurl is not required or even found in libclamav-only mode. So that fails.


ragusaa (approved):

micahsnyder:
Rebased and force-pushed.

Issue #604, opened 2022-05-31
What's the meaning of 'LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1' win7 x64 v105


ggaussling:
Hi, 

I issued this cmd line

`clamscan  --recursive c:\`

sometimes I get this warning:
 `LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1`

I don't know if this is harmless, and where does it come from.

This link here is the only appearance on the net, I was able to find:
```
>> Random Trash/John Titor und Weltlinien • Zeit;Reisen.mp4: OK

LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1
>> Random Trash/CUDA/cuda_10.0.130_win10_network.exe: OK
```
http://webcache.googleusercontent.com/search?q=cache:gQqfY4JsG6kJ:https://uraniumhexafluori.de/ClamAV/&hl=de&gl=de&strip=1&vwsrc=0
Memento: https://web.archive.org/web/20220531130349/https://uraniumhexafluori.de/ClamAV/

e.g.
```
C:\Users\Mich24\Downloads\naps2-6.1.2-setup.exe: OK
LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1
LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1
C:\Users\Mich24\Downloads\Nextcloud-3.4.4-x64.msi: OK
C:\Users\Mich24\Downloads\npp.8.3.3.Installer.x64.exe: OK
C:\Users\Mich24\Downloads\Player Setup.exe: Win.Trojan.Softpulse-519 FOUND
LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1
LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1
LibClamAV Warning: crtmgr_rsa_verify: verification failed: fp_exptmod failed with 1
C:\Users\Mich24\Downloads\ProcessExplorer.zip: OK
```

Hardware:
```
Lenovo IdeaPad N581
Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
Intel Ivy Bridge rev. 09
Intel HM76 rev. 04
4 GBytes DDR3 RAM 798.7 MHz (1:6)
[...]

PR #727, opened 2022-10-19
Fix EXE/DLL cert verification

micahsnyder:
RSA certificate verification is failing because we accidentally lowered the max size of numbers for floating point multiplication with TomsFastMath when upgrading the vendored library code.

This commit restores the default from 4096 to 8192.

Fixes https://github.com/Cisco-Talos/clamav/issues/604


shutton (approved):
Moar bits...

micahsnyder:
Backporting this one is too much pain.  The recent allmatch fixes work cleaned up so much error handling that made this work reliably.  Simply cherrypicking this fix does not cause certificate verification to correctly trust a file.  A number of bugs exist causing it to continue scanning instead of aborting.  Users will have to upgrade from 0.105 -> 1.0 for cert verification to work.

PR #730, opened 2022-10-24
Validate key length before calling encryption

ragusaa:
Add error checking to make sure we aren't calling into aes decrypt functions with larger key sizes, that would make decrypt read past the end of it's data array.

### DIFF
--- a/libclamav/ole2_extract.c
+++ b/libclamav/ole2_extract.c
@@ -2039,9 +2039,9 @@ static inline bool key_length_valid_aes_bits(const uint32_t keyLength)
 {
     switch (keyLength) {
         case SE_HEADER_EI_AES128_KEYSIZE:
-            /* fallthrough */
+            /* fall-through */
         case SE_HEADER_EI_AES192_KEYSIZE:
-            /* fallthrough */
+            /* fall-through */
         case SE_HEADER_EI_AES256_KEYSIZE:
             return true;
     }
@@ -2151,7 +2151,13 @@ static bool aes_128ecb_decrypt(const unsigned char *in, size_t length, unsigned 
     size_t i;
     bool bRet = false;
 
+    if (SE_HEADER_EI_AES128_KEYSIZE != key->key_length_bits) {
+        cli_dbgmsg("ole2: Unsupported AES key length in aes_128ecb_decrypt\n");
+        goto done;
+    }
+
     nrounds = rijndaelSetupDecrypt(rk, (const unsigned char *)key->key, key->key_length_bits);
+
     if (!nrounds) {
         cli_errmsg("ole2: Unable to initialize decryption.\n");
         goto done;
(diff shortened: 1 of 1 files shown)


libclamav/ole2_extract.c
```diff
@@ -2150,8 +2150,13 @@ static bool aes_128ecb_decrypt(const unsigned char *in, size_t length, unsigned
     int nrounds;
     size_t i;
     bool bRet = false;
+    if (SE_HEADER_EI_AES128_KEYSIZE != key->key_length_bits){
```
micahsnyder, line 2153: a new line before this would be nice.

libclamav/ole2_extract.c
```diff
@@ -2186,9 +2191,19 @@ static bool verify_key_aes(const encryption_key_t *const key, encryption_verifie
     uint8_t decrypted[AES_VERIFIER_HASH_LEN];
     uint32_t tmp = 0;
 
-    if (!aes_128ecb_decrypt(verifier->encrypted_verifier, sizeof(verifier->encrypted_verifier),
-                            decrypted, key)) {
-        goto done;
+    switch (key->key_length_bits){
```
micahsnyder, line 2194: Is this switch really required since the size check is done at the top?

libclamav/ole2_extract.c
micahsnyder: Also needs to be `clang-format`'ed

micahsnyder (changes requested):

libclamav/ole2_extract.c
```diff
@@ -2186,9 +2191,19 @@ static bool verify_key_aes(const encryption_key_t *const key, encryption_verifie
...
+        case SE_HEADER_EI_AES128_KEYSIZE:
+            if (!aes_128ecb_decrypt(verifier->encrypted_verifier, sizeof(verifier->encrypted_verifier),
+                                    decrypted, key)) {
+                goto done;
+            }
+            break;
+        case SE_HEADER_EI_AES192_KEYSIZE:
+            /*fallthrough*/
```
micahsnyder, line 2202: I think we try to format these comments like this:
`/* fall-through */`
instead of like:
`/*fallthrough*/`

In searching the code, I see the former occurs in most of our newer code, and `/*fallthrough*/` is only done in one place in the PDF parser.  I suppose gcc must accept either one for silencing the fallthrough warnings, because I haven't seen it complain about them. 🤷

ragusaa: I was just experimenting with clang-12, and it doesn't seem to accept any of them.  I'll change it to fall-through

libclamav/ole2_extract.c
ragusaa: done and done

micahsnyder (approved):

PR #732, opened 2022-10-24
Fix Coverity 401498

ragusaa:
ctx variable is previously checked for NULL in the function, and then dereferenced later without being checked.  Added checks later in the function as well.  This code was copied from elsewhere in the file, so fixed it there as well.

### DIFF
--- a/libclamav/ole2_extract.c
+++ b/libclamav/ole2_extract.c
@@ -78,7 +78,7 @@
 #pragma pack 1
 #endif
 
-//https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/05060311-bfce-4b12-874d-71fd4ce63aea
+// https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/05060311-bfce-4b12-874d-71fd4ce63aea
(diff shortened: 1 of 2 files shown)


micahsnyder (approved):

micahsnyder (changes requested):
So the real Coverity complaint isn't that the NULL checks are missing, but that a `ctx` is dereferenced before a NULL check occurs.

It's subtle, but the `SCAN_COLLECT_METADATA` macro dereferences `ctx` (and also `ctx->options`) without NULL-checks:
```c
#define SCAN_COLLECT_METADATA (ctx->options->general & CL_SCAN_GENERAL_COLLECT_METADATA)
```

Instead, we could put a NULL check at the top of the function to fail if `ctx` is NULL, and then remove the extra NULL checks.  Or even remove the NULL check entirely, since these callback functions are only ever used by `cli_ole2_extract` and it does a NULL check on `ctx` at the very top.

I'll give that a try in this branch and rerun coverity.

micahsnyder:
Rebased, fixed merge conflict, squashed commit (since coverity liked it), and force-pushed.

micahsnyder (approved):

Issue #736, opened 2022-10-26
1.0.0-rc: Freshclam fails download database files


yasuhirokimura:
Describe the bug
----------------

I'm maintainer of FreeBSD ClamAV port. Since release candidate of ClamAV 1.0.0 is released, I tried updating ClamAV port from 0.105.1 to 1.0.0-rc on my private copy of FreeBSD ports repository.

https://github.com/yasuhirokimura/freebsd-ports/tree/clamav/security/clamav

I only change the source archive file to use from clamav-0.105.1.tar.gz to clamav-1.0.0-rc.tar.gz, and don't change anything else (build options, build and/or runtime dependency, etc.). And with this simple change 1.0.0-rc is built successfuly on FreeBSD 13.1-RELEASE amd64 environment.

Next I installed 1.0.0-rc on clean installed FreeBSD 13.1-RELEASE amd64 system and tested if it works fine. But unfortunately I faced problem. If I run freshclam, it fails to download database files as following.

```
--------------------------------------
freshclam daemon 1.0.0-rc (OS: FreeBSD, ARCH: amd64, CPU: amd64)
ClamAV update process started at Wed Oct 26 13:27:19 2022
daily database available for download (remote version: 26699)
ERROR: Verification: Can't verify database integrity
Trying again in 5 secs...
daily database available for download (remote version: 26699)
ERROR: Verification: Can't verify database integrity
Trying again in 5 secs...
daily database available for download (remote version: 26699)
ERROR: Verification: Can't verify database integrity
Giving up on https://database.clamav.net...
ERROR: Update failed for database: daily
ERROR: Database update process failed: Invalid or corrupted CVD/CLD database
ERROR: Update failed.
--------------------------------------
Update process terminated
```

How to reproduce the problem
----------------------------

1. Make clean install of FreeBSD 13.1-RELEASE amd64 system.
2. Login as root
3. Install Git with `pkg install git`
4. Clone my private copy of ports repository with `cd /usr/ports; git clone -b clamav https://github.com/yasuhirokimura/freebsd-ports.git .`
5. Uninstall all packages with `pkg-static delete -af`.
6. Install ClamAV 1.0.0-rc with `cd /usr/ports/security/clamav; make BATCH=yes install`.
7. Copy `/usr/local/etc/clamd.conf.sample` and `/usr/local/etc/freshclam.conf.sample` to `/usr/local/etc/clamd.conf` and `/usr/local/etc/freshclam.conf` respectively.
[...]

PR #740, opened 2022-11-01
Clam 2167 freshclam cld incremental update

micahsnyder:
- Freshclam: fix incremental update on CLD database

  When adding the `cl_cvdunpack()` API that (optionally) verifies the
  database signature, we used it in libfreshclam in a place where it may
  also unpack CLD database archives. CLD's may not be verified, because
  the signature information is no longer valid after incremental update.

  This commit fixes the issue by only verifying the unpack if the file is
  a CVD and not a CLD.

- Test: Add test for incremental update twice in a row

  This is a regression test to ensure that freshclam can update from a CVD
  to a CLD, and then update from the CLD to a (newer) CLD.

Fixes https://github.com/Cisco-Talos/clamav/issues/736

### DIFF
--- a/libclamav_rust/src/sys.rs
+++ b/libclamav_rust/src/sys.rs
@@ -112,7 +112,10 @@ pub type clcb_pre_cache = ::std::option::Option<
         context: *mut ::std::os::raw::c_void,
     ) -> cl_error_t,
 >;
-#[doc = " @brief Pre-scan callback."]
+#[doc = " @brief File inspection callback."]
+#[doc = ""]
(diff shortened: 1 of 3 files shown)


m-sola (approved):
Code and test look good. Manually verified that freshclam runs as expected.

andrew-aitchison (approved):
[b106df](https://github.com/Cisco-Talos/clamav/pull/740/commits/b106dff9e5fb6f9828f8f060c44b515187bea408) looks good and passes my tests.

Issue #276, opened 2021-08-30
clamdscan --allmatch option is sticky: all subsequent scans will be allmatch-mode


klingerko:
Describe the bug
----------------

I have two cdb rules in a single rule database cdb file. Both rules should fire with the allmatch option enabled, but only the chronological "last" rule of the rule file fires. Switching the order of the rules will lead to a different result. I'd expect both rules to fire with the allmatch function enabled.

How to reproduce the problem
----------------------------

```
$ cat test_1.cdb 
Rule1.Zip_vbs:CL_TYPE_ZIP:*:\.[Vv][Bb][Ss]$:*:*:*:*:*:*
Rule2.Zip_vbs:CL_TYPE_ZIP:*:^[a-zA-Z0-9]{12}\.[Vv][Bb][Ss]$:*:*:*:*:*:*
$ clamscan --allmatch=yes -d test_1.cdb 18c5335e8ea82d729cb931a02839bfd3f0cf070a53f7d1a886a1ee4daab64ecf.bin 
Loading:     0s, ETA:   0s [========================>]        2/2 sigs    
Compiling:   0s, ETA:   0s [========================>]       10/10 tasks 

/home/18c5335e8ea82d729cb931a02839bfd3f0cf070a53f7d1a886a1ee4daab64ecf.bin: Rule2.Zip_vbs.UNOFFICIAL FOUND
/home/18c5335e8ea82d729cb931a02839bfd3f0cf070a53f7d1a886a1ee4daab64ecf.bin: Rule2.Zip_vbs.UNOFFICIAL FOUND

----------- SCAN SUMMARY -----------
Known viruses: 2
Engine version: 0.104.0-rc2
Scanned directories: 0
Scanned files: 1
Infected files: 1
Data scanned: 0.47 MB
Data read: 0.46 MB (ratio 1.01:1)
Time: 0.010 sec (0 m 0 s)
Start Date: 2021:08:30 15:30:37
End Date:   2021:08:30 15:30:37
$ cat test_2.cdb 
Rule2.Zip_vbs:CL_TYPE_ZIP:*:^[a-zA-Z0-9]{12}\.[Vv][Bb][Ss]$:*:*:*:*:*:*
Rule1.Zip_vbs:CL_TYPE_ZIP:*:\.[Vv][Bb][Ss]$:*:*:*:*:*:*
[...]

PR #741, opened 2022-11-02
Clam 1106 clamdscan allmatch stickiness

micahsnyder:
- Fix clamdscan --allmatch stickiness bug

  If you run clamdscan with the --allmatch option, it will cause all
  subsequent clamdscan scans to have all-match mode enabled.
  This bug is specific to clamd / clamdscan and does not affect clamscan.
  The problem was introduced when we converted the scan options from a
  single integer bitfield to a struct. The scan options set by the
  clamdscan parameters should be saved in a local copy of the scan
  options, but instead it is saving a copy of the pointer to the scan
  options struct, and so any changes to the scan options affect future
  scans.

- Test: Add clamdscan --allmatch stickiness regression test

  Test that clamdscan --allmatch does not cause future clamdscans to run
  in allmatch mode.

Fixes https://github.com/Cisco-Talos/clamav/issues/276

### DIFF
--- a/clamd/session.c
+++ b/clamd/session.c
@@ -191,12 +191,14 @@ int command(client_conn_t *conn, int *virus)
 {
     int desc                        = conn->sd;
     struct cl_engine *engine        = conn->engine;
-    struct cl_scan_options *options = conn->options;
+    struct cl_scan_options options;
     const struct optstruct *opts    = conn->opts;
     enum scan_type type             = TYPE_INIT;
     int maxdirrec;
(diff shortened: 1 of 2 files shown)


kirkham (approved):

m-sola (approved):
Comparison testing by running clamd, then clamdscan 3 times (no allmatch, allmatch, no allmatch):

In the previous version the last scan gave allmatch results.

Same test on this PR fixes that.

The code itself is straightforward. No issues there.

PR #742, opened 2022-11-03
Clam 2168 CMake: remove option to build using external TomsFastMath library

micahsnyder:
ClamAV requires non-default build options for TomsFastMath to support bigger floating point numbers. Without this change, database and Windows EXE/DLL authenticode certificate validation may fail.

See also discussion in https://github.com/Cisco-Talos/clamav/issues/736

The option to build with an external TomsFastMath library should be removed. The `ENABLE_EXTERNAL_TOMSFASTMATH` build should be ignored, with some sort of warning.

This commit removes the option. If used, the build will not fail, but CMake will print a warning that the unused option was ignored.


micahsnyder:
Updated per review comments offline to add missing PR link

PR #744, opened 2022-11-03
OLE2: Add sector size check when checking for encryption

micahsnyder:
This bug was introduced during 1.0 development and does not affect any stable release.

An OLE2 document that claims to have a sector size (big block size) over 28 will cause a signed integer overflow for our local off_t variable, when multiplied by 4. This will then result in an invalid read and segfault.

The fix is to make sure the sector size (big block size) is not greater than 12, which is the max according to microsoft's documentation.

This commit also fixes some variable types from `unsigned int` or `off_t` over to the more appropriate `size_t`, and from `int` or `uint32_t`, over to `bool`.

Fixes: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=52639

### DIFF
--- a/libclamav/ole2_extract.c
+++ b/libclamav/ole2_extract.c
@@ -115,7 +115,7 @@ typedef struct ole2_header_tag {
      */
     int32_t sbat_root_start __attribute__((packed));
     uint32_t max_block_no;
-    off_t m_length;
+    size_t m_length;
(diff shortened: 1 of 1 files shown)


shutton (approved):
Seems like the next best thing to checked math.

shutton (approved):
Still LGTM, after much discourse.

PR #745, opened 2022-11-04
Reduce # of time limit checks when in logical sig evaluation

micahsnyder:
While reviewing scan time performance, the time limit checks have a notable effect on scan time. I tested reducing the number of time checks to every n'th signature evaluation. For the given sample which has a large number of very tiny embedded scans, I found that overall scan time was best when checked every 10th signature. Reducing the number of checks any further had no noticable effect.

Bear in mind that these numbers include daily/main/bytecode CVD load time:

| current           | mod 2             | mod 10            |
| ----------------- | ----------------- | ----------------- |
| 63.848 s ±1.188 s | 61.773 s ±0.652 s | 59.831 s ±0.975 s |

| mod 50            | mod 100           | mod 1000          |
| ----------------- | ----------------- | ----------------- |
| 59.279 s ±1.652 s | 59.198 s ±1.147 s | 59.440 s ±1.304 s |

### DIFF
--- a/libclamav/matcher.c
+++ b/libclamav/matcher.c
@@ -959,6 +959,12 @@ static cl_error_t lsig_eval(cli_ctx *ctx, struct cli_matcher *root, struct cli_a
         if (CL_SUCCESS != status) {
             goto done;
         }
+
+        // Check time limit here, because bytecode functions may take a while.
+        status = cli_checktimelimit(ctx);
+        if (CL_SUCCESS != status) {
+            goto done;
+        }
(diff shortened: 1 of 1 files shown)


TheRaynMan (approved):
Confirmed faster speed compared to before the looping, and status will carry the value up the stack. At most, we'd add a comment line to explain the value of 10 came from testing.

micahsnyder:
@TheRaynMan good point. I just added the comment.

PR #747, opened 2022-11-05
Fix a few minor compiler/standards issues (clang-16 porting) (0.103.9)

orlitzky:
The next major version of clang has threatened to enable some strictness flags by default:

  https://bugs.gentoo.org/show_bug.cgi?id=clang-16-porting

Specifically,

* -Werror=implicit-function-declaration
* -Werror=implicit-int
* -Werror=strict-prototypes

Whether or not they all wind up on by default, the problems they catch are easy to fix and usually beneficial for e.g. further compiler diagnostics. Here I've fixed what I had to in order to build v0.103.7.

For the autotools build system, the following autoconf fix (and an `autoreconf -fi`) is also required:

  https://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=commit;h=8b5e2016c7ed2d67f31b03a3d2e361858ff5299b

Otherwise, the checks that are emitted will unduly fail.

### DIFF
--- a/clamonacc/c-thread-pool/thpool.c
+++ b/clamonacc/c-thread-pool/thpool.c
@@ -8,7 +8,7 @@
  *
  ********************************/
 
-#define _POSIX_C_SOURCE 200809L
+#define _GNU_SOURCE
 #include <unistd.h>
 #include <signal.h>
 #include <stdio.h>
--- a/clamonacc/clamonacc.c
+++ b/clamonacc/clamonacc.c
@@ -61,7 +61,7 @@
 pthread_t ddd_pid        = 0;
 pthread_t scan_queue_pid = 0;
 
-static void onas_handle_signals();
+static void onas_handle_signals(void);
 static int startup_checks(struct onas_context *ctx);
 static struct onas_context *g_ctx = NULL;
 
--- a/clamonacc/client/socket.h
+++ b/clamonacc/client/socket.h
@@ -31,4 +31,4 @@ struct onas_sock_t {
 };
 
 cl_error_t onas_set_sock_only_once(struct onas_context *ctx);
-int onas_get_sockd();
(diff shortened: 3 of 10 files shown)


micahsnyder:
Thanks for submitting this.  Skimming over the changes it all looks good to me. 
I imagine this all applies to the `main` branch as well. 

We don't commit new development directly to our release branches, so I've created `dev/0.103.8` and changed this PR to target it instead. 

I also don't like to merge PR's into a development branch until we've changed the clam version number in that branch, so even once this is tested, we'll hold off until that is done.

orlitzky:
> Thanks for submitting this. Skimming over the changes it all looks good to me. I imagine this all applies to the `main` branch as well.

Yes, probably. We have two similar bugs reported against v0.105.x ([871252](https://bugs.gentoo.org/871252), [879625](https://bugs.gentoo.org/879625)), but I'm only maintaining the LTS these days. If you're feeling proactive, you can try to build the main branch yourself with those three `-Werror=...` flags enabled. They work just as well in GCC.

> We don't commit new development directly to our release branches, so I've created `dev/0.103.8` and changed this PR to target it instead.
> 
> I also don't like to merge PR's into a development branch until we've changed the clam version number in that branch, so even once this is tested, we'll hold off until that is done.

No problem. Thanks.

ragusaa (approved):

micahsnyder:
Ugh. We didn't merge this before releasing 0.103.8.  I'm sorry.  We can try to put this towards 0.103.9.   At least we did merge the one that went into main towards future versions.

orlitzky:
NBD. Clang 16 is expected "any day now" so I was peer pressured into patching v0.103.8. But with the LTS series on life-support-only the patches aren't that hard to manage.

PR #762, opened 2022-11-17
Clam 2179 Mitigate crash if image decoder fails

micahsnyder:
If the image decoders (i.e. jpeg, tiff, png, etc.) fail to load an image due to a panic, the application will crash.

This commit attempts to catch those panics and handle the error.

I was able to test this using a sample that crashes an older version of the tiff image decoder by modifying the Cargo.lock file with:
```ini
[[package]]
name = "tiff"
version = "0.7.3"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "7259662e32d1e219321eb309d5f9d898b779769d81b76e762c07c8e5d38fcb65"
dependencies = [
 "flate2",
 "jpeg-decoder",
 "weezl",
]
```

The output is a little verbose when it happens. It prints the panic details:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Format("scan makes use of unset ac huffman table")', /home/micasnyd/.cargo/registry/src/github.com-1ecc6299db9ec823/tiff-0.7.3/src/decoder/image.rs:348:45
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
/mnt/c/Users/micasnyd/Downloads/eofloop_valid_huff.tif.847935d1f8: OK
```
But as you can see, it does not crash.  Before this change, this scan crashes like this:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Format("scan makes use of unset ac huffman table")', /home/micasnyd/.cargo/registry/src/github.com-1ecc6299db9ec823/tiff-0.7.3/src/decoder/image.rs:348:45
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
fish: '~/clams/1.0.0-rc/bin/clamscan -…' terminated by signal SIGABRT (Abort)
```

### DIFF
--- a/libclamav_rust/src/fuzzy_hash.rs
+++ b/libclamav_rust/src/fuzzy_hash.rs
@@ -26,6 +26,7 @@ use std::{
     ffi::CStr,
     mem::ManuallyDrop,
     os::raw::c_char,
+    panic,
     slice,
 };
 
@@ -58,6 +59,9 @@ pub enum FuzzyHashError {
 
     #[error("Failed to load image: {0}")]
     ImageLoad(image::ImageError),
+
+    #[error("Failed to load image due to bug in image decoder")]
+    ImageLoadPanic(),
 
     #[error("Invalid parameter: {0}")]
     InvalidParameter(String),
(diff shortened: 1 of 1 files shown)


shutton (approved):
Looks good.  Might want to revisit later to see if more detail can/should be added into the error.

PR #767, opened 2022-11-18
Fix a few minor compiler/standards issues (clang-16 porting) (for main branch)

micahsnyder:
Copy of https://github.com/Cisco-Talos/clamav/pull/747 forward-ported for `main` branch. 

In addition, this PR:
* Fixes a few new strict prototype issues.
* Enables these warnings as errors by default at the top level CMake to prevent accidentally introducing more of these in the future. 

The description from #747 is as follows...

--- 

The next major version of clang has threatened to enable some strictness flags by default:

  https://bugs.gentoo.org/show_bug.cgi?id=clang-16-porting

Specifically,

* -Werror=implicit-function-declaration
* -Werror=implicit-int
* -Werror=strict-prototypes

Whether or not they all wind up on by default, the problems they catch are easy to fix and usually beneficial for e.g. further compiler diagnostics. Here I've fixed what I had to in order to build v0.103.7.

For the autotools build system, the following autoconf fix (and an `autoreconf -fi`) is also required:

  https://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=commit;h=8b5e2016c7ed2d67f31b03a3d2e361858ff5299b

Otherwise, the checks that are emitted will unduly fail.


micahsnyder:
@ragusaa  I just removed that commit which did this:

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f2c95c35f..74c09bbbb 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -851,6 +851,9 @@ else()
     extract_valid_c_flags(WARNCFLAGS
         -Wall -Wextra
         -Wformat-security
+        -Werror=implicit-function-declaration
+        -Werror=implicit-int
+        -Werror=strict-prototypes
     )
     if(ENABLE_ALL_THE_WARNINGS)
         extract_valid_c_flags(WARNCFLAGS
```

As discussed, we can figure out how to add this in our CI instead, to prevent introducing new warnings.

ragusaa:
This fails to build when running those warnings as errors with clang-14.  The two suggestions I have are 

1. Add the following to the top of crypto.c
`
#if !defined(_WIN32)
#define _GNU_SOURCE 1
#endif
`

2. Change the following line in CMakeLists.txt
`
    if(CMAKE_COMPILER_IS_GNUCXX)
`
to 
`
if(CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang"))
`

micahsnyder:
> This fails to build when running those warnings as errors with clang-14. The two suggestions I have are
> 
>     1. Add the following to the top of crypto.c
>        `#if !defined(_WIN32) #define _GNU_SOURCE 1 #endif`
> 
>     2. Change the following line in CMakeLists.txt
>        `if(CMAKE_COMPILER_IS_GNUCXX)`
>        to
>        `if(CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang"))`

I went with a variation on the #2 option.  I think it is fixed now.

ragusaa (approved):

Issue #771, opened 2022-11-18
Confusing .wdb error message for invalid signature in clamav 1.0


LSchuepbach:
Describe the bug
----------------
ClamAV refuses to load `.wdb` signatures with such a format:
`X:urldefense\.com:.+`

Here's an excerpt of the error message:
```
LibClamAV debug: Loading regex_list
LibClamAV Warning: cli_realloc(): File or section is too large to scan (0 bytes).                      For your safety, ClamAV limits how much memory an operation can allocate to 1073741824 bytes
LibClamAV Error: Can't load /root/test.wdb: Can't allocate memory
LibClamAV Error: cli_loaddbdir: error loading database /root/test.wdb
ERROR: Can't allocate memory
```

Using such a format helps a lot to deal with the FPs coming when a correspondent uses a URL-rewriting phishing protection. urldefense.com is just one example in a long list of such services that I don't want to get hits on.

How to reproduce the problem
----------------------------
I used today's docker's [unstable_base](https://hub.docker.com/layers/clamav/clamav/unstable_base/images/sha256-4f68aba20db9519df03baa621453e38e23492e06338bdd2fc07f864ba3aa0ee6).

Create a file `test.wdb` with the content mentioned above.

Using `X:urldefense\.com:.{1,30}`, ClamAV will not complain while loading, but the SpoofedDomain will still hit.

It works without problem under latest_base (0.105.1)

micahsnyder:
@LSchuepbach 

I looked at it harder and realized that the provided WDB signature is actually invalid and fails to load in 0.105 (and prior) as well, but it fails silently. Fixes during 1.0 development that patched WDB signature load bugs now cause it to fail loudly.

Unfortunately the error message is completely unhelpful and doesn't explain why the signature is invalid. So the reported issue really should still be looked at, but mostly to improve the error handling to detect the issue and explain why the signature is invalid.  That's not as urgent since it's not really a new bug in 1.0 after all.

My understanding right now is a little muddled.   Our documentation claims that the second URL in the signature should have the last 3 characters in the expression as an exact match, as stated here: https://docs.clamav.net/manual/Signatures/PhishSigs.html#wdb-format
However, for similar clicktime protection sites, I have had luck in testing with a format like this:
```
[...]

PR #773, opened 2022-11-18
Clam 2191 wdb load bug

micahsnyder:
- Fix error when loading some WDB databases

  An error occurs when loading the WDB database with this line:

    X:urldefense\.com:.+

  This happens because it goes to add a new suffix for the second domain
  but there is nothing there, so it attempts to undo by shrinking the
  suffix list back to original size. The suffix list in this case goes
  back to zero, and our custom CLI_REALLOC function balks at reallocing to
  zero.

  I think the correct solution is to not free if it goes to zero, just in
  case it is dereferenced later. It at least doesn't leak, so this is a
  safe solution.
    
- Fix formatting for memory allocation warning messages

  The indentation shows up in the message, which is not desired.

Fixes: https://github.com/Cisco-Talos/clamav/issues/771

### DIFF
--- a/libclamav/others_common.c
+++ b/libclamav/others_common.c
@@ -221,8 +221,8 @@ void *cli_malloc(size_t size)
     void *alloc;
 
     if (!size || size > CLI_MAX_ALLOCATION) {
-        cli_warnmsg("cli_malloc(): File or section is too large to scan (%zu bytes). \
-                     For your safety, ClamAV limits how much memory an operation can allocate to %zu bytes\n",
+        cli_warnmsg("cli_malloc(): File or section is too large to scan (%zu bytes). "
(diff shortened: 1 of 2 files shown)


micahsnyder:
My PR is bad. It resolves one specific part of the failure though it still fails overall.

I looked at it harder and realized that the provided WDB signature is actually invalid and fails to load in 0.105 (and prior) as well, but it fails silently.  Fixes during 1.0 development that patched WDB signature load bugs now cause it to fail loudly.

Unfortunately the error message is completely unhelpful and doesn't explain why the signature is invalid.  So the reported issue really should still be looked at, but mostly to improve the error handling to detect the issue andd explain why the signature is invalid.

PR #787, opened 2022-11-30
CMake: Enable selection of a specific python impl

Kangie:
On distros with multiple python impls it can be useful to select a specific version rather than whatever CMake thinks is appopriate.

This patch enables users to instruct CMake to look for a specific version of python by passing `-DPYTHON_FIND_VER`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,20 +223,26 @@ if(ENABLE_TESTS)
     find_package(Libcheck REQUIRED)
 
     # Used to generate the test files and for the application feature test framework
-    find_package(Python3 REQUIRED)
-
-    # First try using pytest from the PATH
-    execute_process(
-        COMMAND pytest --version
-        RESULT_VARIABLE PYTEST_EXIT_CODE
-        ERROR_QUIET OUTPUT_QUIET
-    )
-
-    if(${PYTEST_EXIT_CODE} EQUAL 0)
-        # pytest found in the path.
-        set(PythonTest_COMMAND "pytest;-v")
+    # In distros that support multiple implementations of python it is helpful to specify the impl to use
+    if(DEFINED PYTHON_FIND_VERSION)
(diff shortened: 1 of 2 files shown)


micahsnyder:
Thanks for submitting this contribution, @Kangie 
I will test it momentarily and get back to you on how that works. 

Can you please find a place to document the new cmake option in the `INSTALL.md` file? 
Perhaps somewhere in this section https://github.com/Cisco-Talos/clamav/blob/main/INSTALL.md?plain=1#L356
And perhaps also where it says that Python is a requirement, to give some explanation how Python is found and reference the option.

Kangie:
> Can you please find a place to document the new cmake option in the `INSTALL.md` file? Perhaps somewhere in this section https://github.com/Cisco-Talos/clamav/blob/main/INSTALL.md?plain=1#L356 And perhaps also where it says that Python is a requirement, to give some explanation how Python is found and reference the option.

Will do, I'm travelling at the moment so expect an update in a couple of days. I figured there would be a docs update required, at least there's an opportunity to address any  feedback too. 😄 

There are "better" ways to do this (e.g. setting the magic python3 executable var) but they're incompatible with calling pytest directly, which works in the vast majority of cases; I didn't want to change up the existing logic and remove that, so we got a new variable.

micahsnyder:
> Will do, I'm travelling at the moment so expect an update in a couple of days. 

Sounds good.  Just leave a comment when you have updated it and are ready for a re-review.  I will also be traveling for about a week.  I will respond when I'm back.

Kangie:
@micahsnyder I think that this should cover it. The addition under 'Build Tools' is generic enough to cover LLVM, Python and anything else that comes up in the future.

If we need more detail on the specifics I can add a link to the CMake docs for 'find_package' however I think that this is fine.

micahsnyder (approved):
Thanks @Kangie looks great!  Sorry about the delay. 
Happy holidays!

PR #795, opened 2022-12-15
Fix printf() argument type mismatch warnings

rzvncj:
"format ‘%zu’ expects argument of type ‘size_t’, but argument 3 has type ‘int’ [-Wformat=]"


rzvncj:
I've read the Windows build log and I don't believe that the build error has anything to do with this change.

rzvncj:
Ping? :) As discussed on Discord, the automated check errors are not caused by the patch.

micahsnyder (approved):
Looks good to me. In testing I verified that it resolved the warnings. Thank you for the fix @rzvncj

rzvncj:
> Looks good to me. In testing I verified that it resolved the warnings. Thank you for the fix @rzvncj

Thanks for merging!

PR #800, opened 2022-12-29
Avoid altering signature for .cargo/vendor files

atoomic:
When using the content of the `clamav` tarball in a git repository to version a debian or RedHat package for example. We should never alter the files from cargo or this result in corrupted signature and thus we cannot install the files.

As the repo provides its own `.gitattributes` we cannot easily overwrite it without manually updating `.git/info/attributes`.

Alternatively we could remove the `.gitattributes` file from the tarball when generating it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,6 +9,9 @@
 # Files that should be left untouched (binary is macro for -text -diff)
 *.ref           binary
 
+# Preserve signature for .cargo/vendor files (from the tarabll)
++/.cargo/vendor binary
+
 #
 # Exclude files from exporting
 #


atoomic:
Note that the published tarballs contain `.cargo/vendor` files and at the same time a global `.gitattributes` (at the root) coming from https://github.com/Cisco-Talos/clamav/blob/main/.gitattributes#L3 which set explicitly `* text=auto` 

This change is overwriting that policy for the content of `libclamav_rust` directory so rust files are not updated which could then result in invalid signature from the `cargo-checksum.json` files.

libclamav_rust/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# preserve signature for .cargo/vendor files
+* -text
```
micahsnyder, line 2: The desire for this change makes sense to me, to prevent git from doing end-of-line conversions for any files found in the `.cargo/vendor` directory. 

I'm not sure that adding a file here, or having it apply to all files under `libclamav_rust` makes sense. 

Instead, i wonder if the patch should be:
```diff
diff --git a/.gitattributes b/.gitattributes
index 2253303d2..1cf4e5d9c 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,7 +7,8 @@
 #

 # Files that should be left untouched (binary is macro for -text -diff)
-*.ref           binary
+*.ref                           binary
+libclamav_rust/.cargo/vendor    binary

 #
 # Exclude files from exporting
 ```
 
 or maybe just
 ```diff
diff --git a/.gitattributes b/.gitattributes
index 2253303d2..555b0ce6e 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,6 +8,7 @@

 # Files that should be left untouched (binary is macro for -text -diff)
 *.ref           binary
+.cargo/vendor   binary

 #
 # Exclude files from exporting
```

micahsnyder (changes requested):

libclamav_rust/.gitattributes
toddr: Ideally, whatever change we make here, we need .gitattributes not effectively alter files shipped in the tarball. Even better would be for .gitattributes to NOT ship in the tarball. Is that possible?

atoomic: Indeed removing the `.gitattributes` from the tarball would also solve the issue as I noted in the commit message.
but your suggestion is right, using a single `.gitattributes` file is easier for maintenance and could benefits any other locations where `cargo` packages files are added to the tarball

atoomic:
I ve resubmitted using your suggestion

libclamav_rust/.gitattributes
micahsnyder: To be honest, I misunderstood about removing the `.gitattributes` from the tarball.  But I am happy enough with this solution to explicitly leave the `.cargo/vendor` directory alone. 

Thank you for submitting this pr!

micahsnyder (approved):

PR #801, opened 2022-12-29
Bump minimum Rust version requirement

atoomic:
The rust version required is dictated by `jpeg-decoder` requirements which cannot be compiled without version `1.61` or newer.

Rather than adding a new section in `News.md` for `0.105.3` just bump the entry for `0.105`.


micahsnyder (approved):
Thanks @atoomic this makes sense.

PR #803, opened 2023-01-03
Allow to use system's tomsfastmath library instead of the bundled one.

sebastianas:
By default the bundled tomsfastmath library is used. It can be overriden to use the library from the system if available. Should the system library only be able to handle RSA keys up to 2048 bits then the selftest will fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -466,6 +466,11 @@ if(ZLIB_FOUND)
     set(HAVE_LIBZ 1)
 endif()
 
+find_package(TOMSFASTMATH REQUIRED)
+if(TOMSFASTMATH_FOUND)
+   set(HAVE_TFM_H 1)
+endif()
+
 find_package(BZip2 REQUIRED)
 if(BZIP2_FOUND)
     set(HAVE_BZLIB_H 1)
@@ -1211,6 +1216,16 @@ ${_}                            ${e}${OPENSSL_LIBRARIES}
 ${b}        JSON support:       ${e}
 ${_}            json-c          ${e}${JSONC_INCLUDE_DIRS}
 ${_}                            ${e}${JSONC_LIBRARIES}
+${b}        Math support:       ${e}")
+if(TOMSFASTMATH_BUILTIN)
+message("\
+${_}            tomsfastmath    ${e}built-in")
+else()
+message("\
+${_}            tomsfastmath    ${e}${TOMSFASTMATH_INCLUDE_DIR}
+${_}                            ${e}${TOMSFASTMATH_LIBRARY}")
+endif()
+message("\
 ${b}        Threading support:  ${e}")
 if(WIN32)
 message("\
--- a/CMakeOptions.cmake
+++ b/CMakeOptions.cmake
@@ -116,6 +116,9 @@ option(ENABLE_SYSTEMD
     "Install systemd service files if systemd is found."
     ${ENABLE_SYSTEMD_DEFAULT})
 
+option(ENABLE_EXTERNAL_TOMSFASTMATH
+    "Use system's tomsfastmath instead of internal bundled version.")
+
 # For reference determining target platform:
 #  Rust Targets:  https://doc.rust-lang.org/nightly/rustc/platform-support.html
 option(RUST_COMPILER_TARGET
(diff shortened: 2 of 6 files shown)


sebastianas:
On 2023-01-06 10:22:48 [-0800], Micah Snyder wrote:
> We had this feature but removed it, here: https://github.com/Cisco-Talos/clamav/commit/5c3e866d30695214ba438f38228b8b0952b54987
> 
> A standard build of tomsfastmath does not support wide enough floating
> point numbers for our use case and will not work correctly.  It is
> better (for us) to know that everyone is using tomsfastmath that is
> configured to work for clamav. 

Could we have something that is not announced? I kind of need to use the
in-tree lib rather than the bundled copy.

You could as upstream to update lib and make a release ;) This is the
only change right?

Sebastian

micahsnyder:
> Could we have something that is not announced? 

It is included in the major changes in the change log;
- https://blog.clamav.net/2022/11/clamav-100-lts-released.html
- https://github.com/Cisco-Talos/clamav/blob/clamav-1.0.0/NEWS.md?plain=1#L45-L50

> I kind of need to use the in-tree lib rather than the bundled copy. 

I feel like "need" is an exaggeration. It is not your responsibility to de-bundle bundled sources. I see the appear if a bunch of other packages are using tomsfastmath to be able to push out 1 update to tomsfastmath if there is a security issue and have that cover them all. I don't know how to check what other packages depend on https://packages.debian.org/source/bullseye/tomsfastmath.  Though that sort of ties into my next point.

> You could as upstream to update lib and make a release ;) 

I don't know if changing from the defaults will break other software packages that expect the default settings.  If there are no other packages that depend on tomsfastmath, then it seems even less critical that this tomsfastmath is de-bundled from clamav.

This idea is not invalid, but it doesn't take into consideration all of the other distros that may use clamav and think "hey, external tomsfast seems better, let's do that!" We obviously can't find and "fix" tomsfastmath packages for every distro.

> This is the only change right? Sebastian

We also build with these defined:
- `TFM_CHECK`
- `TMP_NO_ASM`

Set through the build system in: https://github.com/Cisco-Talos/clamav/commit/375ecf678c714623e6fb5c0119d1bec98dc700dd#diff-6d5ce841c8f84bb1d6383da4bf5fa35316926e2e0668ad71aa137d7bd5429c43R26-R34

sebastianas:
On 2023-01-23 11:43:11 [-0800], Micah Snyder wrote:
> 
> > Could we have something that is not announced? 
> 
> It is included in the major changes in the change log;
> - https://blog.clamav.net/2022/11/clamav-100-lts-released.html
> - https://github.com/Cisco-Talos/clamav/blob/clamav-1.0.0/NEWS.md?plain=1#L45-L50

I meant to support system's lib without making a big fuss about.

> > I kind of need to use the in-tree lib rather than the bundled copy. 
> 
> I feel like "need" is an exaggeration. It is not your responsibility
> to de-bundle bundled sources. I see the appear if a bunch of other
> packages are using tomsfastmath to be able to push out 1 update to
> tomsfastmath if there is a security issue and have that cover them
> all. I don't know how to check what other packages depend on
> https://packages.debian.org/source/bullseye/tomsfastmath.  Though that
> sort of ties into my next point.

$ apt-cache rdepends libtfm1
libtfm1
Reverse Depends:
  libtfm-dev
  libtfm1-dbgsym
  libclamav11

In general the policy is to use system's libs. For this reasons I also
provided the glue layer for libmspack.

> > You could as upstream to update lib and make a release ;) 
> 
> I don't know if changing from the defaults will break other software
> packages that expect the default settings.  If there are no other
> packages that depend on tomsfastmath, then it seems even less critical
> that this tomsfastmath is de-bundled from clamav.
> 
> This idea is not invalid, but it doesn't take into consideration all
> of the other distros that may use clamav and think "hey, external
> tomsfast seems better, let's do that!" We obviously can't find and
> "fix" tomsfastmath packages for every distro.

Hmm. I started looking into replacing tfm with openssl bignum (BN). That
should fix all things are is there something that is unique to tfm? I
just started so I may discover it soon - or not ;)

> > This is the only change right? Sebastian
> 
> We also build with these defined:
> - `TFM_CHECK`
> - `TMP_NO_ASM`

That should be what we do.


Sebastian

sebastianas:
This is obsoleted by #840

micahsnyder:
Closing because not desired out of compatibility concerns with default TFM build config for other distros that may use it, and because obsoleted by #840 as Sebastian said.

PR #776, opened 2022-11-19
Use version-script with cmake builds

opoplawski:
This is an attempt to resurrect symbol versioning and fix #775 .  One problem is the new unit tests that link to libclamav and test internal functions do not have access to the symbols.  I started adding them to the symbol map, but I don't think this is the proper solution.  I think instead there perhaps be a static library that is used for the unit tests?

PR #804, opened 2023-01-03
Use a symbol file

sebastianas:
The symbol file is needed to limit the exported symbols. Otherwise all public symbols are exported and library management becomes a mess in Debian.

I don't know if the added symbols are meant for public or private use so I added them to the private section.

Sebastian


micahsnyder:
Thank you @sebastianas this looks to solve a problem I had trouble solving a few years ago.  I

For reference, here's the link to the other PR where Sebastian added the same for libclamunrar and libclamunrar_iface: https://github.com/Cisco-Talos/clamav/pull/816 

We'll test this as soon as we get a chance.

sebastianas:
This and a bunch of other of pulls fail in the winows-build category but I think now that this has nothing to do with the change.
This is a friendly ping since I believe that I have no to-do pending.

micahsnyder:
> This and a bunch of other of pulls fail in the winows-build category but I think now that this has nothing to do with the change. This is a friendly ping since I believe that I have no to-do pending.

Yup the windows build has an issue for all PR's right now.  Our internal Jenkins pipeline builds the installers and tests on windows 32bit and 64bit with more tests on both windows 7 and 10 and it hasn't had any issues.  So I haven't put a priority on fixing the github actions windows pipeline.  For now just don't worry about it.

micahsnyder (changes requested):
The macOS build in github actions failed with:
```
...
[ 61%] Linking C shared library libclamav.dylib
ld: unknown option: --version-script=/Users/runner/work/clamav/clamav/libclamav/libclamav.map
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libclamav/libclamav.11.0.0.dylib] Error 1
make[1]: *** [libclamav/CMakeFiles/clamav.dir/all] Error 2
make: *** [all] Error 2
Error: Process completed with exit code 2.
```

I just confirmed that it failed the same way for macOS in Jenkins. 
I think we need to disable the link flag for apple, like this: https://github.com/PaddlePaddle/Paddle/pull/9206/files

The internal Jenkins build for Windows x64 and x86 didn't have any issue. Testing on FreeBSD also looked good.

micahsnyder:
I forgot to mention -- @opoplawski has a similar patch work in progress with some additional changes for versioning the API:

https://github.com/Cisco-Talos/clamav/pull/776/files 

Is there anything you want to add from Orion's work (with credit, of course)?  Or else after merging this if Orion is interested in rebasing his work-in-progress PR.

opoplawski:
Looks like this adds `LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/libclamav.map`, which is good.  I went with:
```
    if(UNIX AND NOT APPLE)
        target_link_options( clamav PRIVATE "-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libclamav.map")
    endif()
```
to exclude Apple.  I'm not sure which is preferred in general: `target_link_options` or `set_target_properties(clamav PROPERTIES LINK_FLAGS`.

We have different sorting in libclamav.map so hard to compare in detail.  The one thing I tried to do differently was to try to track what version some items were added, but I don't know if this is really all that useful here.

I didn't address libfreshclam.map, but I did add the link flags for libclamunrar_iface as well.

sebastianas:
On 2023-03-09 16:43:30 [-0800], Orion Poplawski wrote:
> Looks like this adds `LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/libclamav.map`, which is good.  I went with:
> ```
>     if(UNIX AND NOT APPLE)
>         target_link_options( clamav PRIVATE "-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libclamav.map")
>     endif()
> ```
> to exclude Apple.  I'm not sure which is preferred in general: `target_link_options` or `set_target_properties(clamav PROPERTIES LINK_FLAGS`.

I'm going to update my pull and exclude the Apple, too. If we would
really care about apple we could use check_linker_flag()
	https://cmake.org/cmake/help/latest/module/CheckLinkerFlag.html

however it is available since 3.18. The problem seems to be that the
apple box is not using a GNU toolchain. As I young boy on AIX, HP-UX or
even SUN: my first steps were always to get a GNU toolchain & tools. It
turned out to be time spent wise ;)

> We have different sorting in libclamav.map so hard to compare in
> detail.  The one thing I tried to do differently was to try to track
> what version some items were added, but I don't know if this is really
> all that useful here.

Sorted it on one commit and added missing functions in the second.
You also added "CLAMAV_0.104.0" which seems you intend to maintain
version compatibility. As someone who maintains this package in Debian I
don't think this makes sense due to the way cl_retflevel() is used.

You *must* ensure that this symbol points always to the latest version.
The problem is that if you compile something against libclamav v1.0.1 it
will record CL_FLEVEL 161. If you try to run this binary against
libclamav v1.0.0 it will complain because CL_FLEVEL is 160 there.
Otherwise it is fine ABI wise.
For that reason I ignore the versions and ensure that everything that
was compiled against libclamav X depends on X or later. Therefore I
don't see any added value in using more detailed symbol versions.

Sebastian

micahsnyder:
I like that you've sorted them alphabetically.  That's certainly easier to reference.

After a fair bit of review I think @opoplawski 's PR has the advantage for two, maybe 3 reasons.

1. I think `target_link_options` is the newer option. According to [this documentation for LINK_DEPENDS](`https://cmake.org/cmake/help/latest/prop_tgt/LINK_DEPENDS.html`), the LINK_DEPENDS property is supported only by [Ninja](https://cmake.org/cmake/help/latest/generator/Ninja.html#generator:Ninja) and [Makefile Generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#makefile-generators).  Meanwhile, [the docs say that `target_link_options` was added in 3.13](https://cmake.org/cmake/help/latest/command/target_link_options.html) and does not list that limitation.  So I think it may be the newer (preferred?) option.   Regardless, I like that `target_link_options` requires fewer lines of code to implement.

2. Orion's PR includes using libclamunrar_iface.map, which is missing here and is probably desirable. 

3. I agree with you @sebastianas that clamscan/clamd/etc require a matching functionality level (FLEVEL) or else will complain. Our programs use this logic:
   ```c
   int check_flevel(void)
   {
       if (cl_retflevel() < CL_FLEVEL) {
           fprintf(stderr, "ERROR: This tool requires libclamav with functionality level %u or higher (current f-level: %u)\n", CL_FLEVEL, cl_retflevel());
           return 1;
       }
       return 0;
   }
   ```
   But it's likely other downstream projects will not be so fussy, and are willing to swap in a newer libclamav.  So it is possible they will benefit from the versioning that Orion has added. 

In summary, I think it's best to go with Orion's solution so I will close this one.  If you're interested in submitting another PR to alphabetize the symbols after Orion's work has been merged, that would be a pleasant change even if it's not functional.  

That said -- I'm seeing a failure building the check_clamav unitttest program when building with static library dependencies. So I may have to ask for a change to fix that.  Working that out right now.

sebastianas:
On 2023-03-17 16:49:40 [-0700], Micah Snyder wrote:
> After a fair bit of review I think @opoplawski 's PR has the advantage for two, maybe 3 reasons.
…
> 3. I agree with you @sebastianas that clamscan/clamd/etc require a matching functionality level (FLEVEL) or else will complain. Our programs use this logic:
>    ```c
>    int check_flevel(void)
>    {
>        if (cl_retflevel() < CL_FLEVEL) {
>            fprintf(stderr, "ERROR: This tool requires libclamav with functionality level %u or higher (current f-level: %u)\n", CL_FLEVEL, cl_retflevel());
>            return 1;
>        }
>        return 0;
>    }
>    ```
>    But it's likely other downstream projects will not be so fussy, and
>    are willing to swap in a newer libclamav.  So it is possible they
>    will benefit from the versioning that Orion has added. 

How do they benefit from that versioning? I don't see added value in
additional version-table. Having a split between PUBLIC and PRIVATE
helps in terms "here I need to pay attention to ABI changes and here I
don't".
glibc is using this to provide two (or more) implementation of a certain
function so the default behaviour can change and is still provided to an
older version without recompiling. So you could make this work with
check_flevel() if you want but this requires extra work.

Also: If I compile X against libclamav Y then X requires libclamav
>= Y. This relationship is not recorded in terms of check_flevel(). To
"fix" this you would need to provide different check_flevel()
implementations for "older" Version of libclamav so programs compiled
against libclamav-1.0.0 have a different function than those compiled
against libclamav-1.0.1. And you would have a set function introduced in
1.0.0 and 1.0.1 (and later if any).

The point of having these versioned functions is exactly to provide one
function which remains compatible in a newer library without
recompiling. glibc is using this extensively.
This can also be used to record a dependency on a package. So you
compile against libclamav-1.0.1 you could also use libclamav-1.0.0 _if_
the lowest used symbol is from libclamav-1.0.0. But the way
check_flevel() works, you need to use libclamav-1.0.1. Also, must not
rename version strings (i.e. for check_flevel() from 1.0.0 to 1.0.1)
because a function is missing which means everything compiled against
1.0.0 stops working because a symbol is missing. You only change
existing entries during so-major change).

According to my memory the recorded version was 0.10x.y (x,y slipped my
memory). This should start with the 1.0.0 version (or whatever the
main branch provides).

> In summary, I think it's best to go with Orion's solution so I will
> close this one.  If you're interested in submitting another PR to
> alphabetize the symbols after Orion's work has been merged, that would
> be a pleasant change even if it's not functional.  

I don't care about alphabetize the symbols it was just easier to parse.
I don't care which way is used for symbol versions. I just need one for
my distro work and I simply made the old way working which was around
before the switch to cmake. Eitherway I'm fine with what you end up but
I would appreciate not to have any extra work (say if I need to patch
out the versions or so because I can't recompile every downstream user
during a new clamav release with a SO-change).

> That said -- I'm seeing a failure building the check_clamav unitttest
> program when building with static library dependencies. So I may have
> to ask for a change to fix that.  Working that out right now. 

oki.

Sebastian

opoplawski:
I'm not super familiar with symbol versioning - but as I understand it another use it to track when symbols were added.  So if foo() were added in clamav 1.2 libclamav.so.12, but there wasn't a need for a soname bump from clamav 1.1 libclamav.so.12, programs compiled against libclamav from 1.2 would get marked (at least by rpm) as requiring libclamav.so.12(CLAMAV_1.2) and could not accidentally be installed on systems with clamav 1.1.  This was what I was mainly going for.

As such, marking things now with CLAMAV_0.104.0 is probably pointless, but not harmful I think.

PR #807, opened 2023-01-06
Clam 2214 bounds checking regex suffix

ragusaa:


ragusaa:
> I wonders if this PR also fix regex in Yara too?

I am sorry, I must have missed your comment.  Do you happen to have a testcase I can try with this PR?

net1:
It is back-references regex and I just found that back-references are not support in Yara rules.

libclamav/regex_suffix.c
```diff
@@ -179,8 +184,25 @@ static void destroy_tree(struct node *n)
...
-static uint8_t *parse_char_class(const uint8_t *pat, size_t *pos)
+static uint8_t *parse_char_class(const uint8_t *pat, size_t patSize, size_t *pos)
 {
+
+#ifndef inc_pos
+#define add_pos(posPtr, incVal, posMax)                              \
+    {                                                                \
+        do {                                                         \
+            if (((*posPtr) + incVal) >= posMax) {                    \
+                cli_warnmsg("parse_char_class: Invalid char class"); \
```
micahsnyder, line 195: Missing a `\n` here.

libclamav/regex_suffix.c
```diff
@@ -480,7 +505,7 @@ cl_error_t cli_regex2suffix(const char *pattern, regex_t *preg, suffix_callback
                cli_errmsg("cli_regex2suffix: unable to strdup regex.pattern");
```
micahsnyder, line 505: Also missing a new line `\n` here and in a few other error messages in this file.

libclamav/regex_suffix.c
```diff
@@ -179,8 +184,25 @@ static void destroy_tree(struct node *n)
     FREE(n);
 }
 
-static uint8_t *parse_char_class(const uint8_t *pat, size_t *pos)
+static uint8_t *parse_char_class(const uint8_t *pat, size_t patSize, size_t *pos)
 {
+
+#ifndef inc_pos
+#define add_pos(posPtr, incVal, posMax)                              \
```
micahsnyder, line 191: And Interesting solution to limit duplicate code.  I'm not a huge fan of macros because they're tough to debug. But it doesn't really make sense to do a static inline function for a  single comparison.  I won't complain too much I guess. 

So my only is I think we should capitalize as `INC_POS` and `ADD_POS` like we do with other macros.

micahsnyder (changes requested):
Testing went well.
Outside of the above, looks good to me.

micahsnyder (approved):

PR #811, opened 2023-01-19
Clam 2212 Moved structure free to after last use

ragusaa:


micahsnyder (approved):
Well this one is pretty self explanatory! Ha...

PR #814, opened 2023-01-22
libclamav/pe: Use endian wrapper in more places.

sebastianas:
A few user of VirtualAddress and Size in cli_exe_info::pe_image_data_dir don't use the endian wrapper while other places do. This leads to testsuite failures on big endian machines.

Use the endian wrapper in all places across pe.c for the two members.

### DIFF
--- a/libclamav/matcher.c
+++ b/libclamav/matcher.c
@@ -766,7 +766,7 @@ int32_t cli_bcapi_matchicon(struct cli_bc_ctx *ctx, const uint8_t *grp1, int32_t
             !ctx->hooks.pedata->dirs[2].Size)
             info.res_addr = 0;
         else
-            info.res_addr = le32_to_host(ctx->hooks.pedata->dirs[2].VirtualAddress);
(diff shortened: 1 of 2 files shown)


micahsnyder:
This looks really good to me.  Thank you @sebastianas!  I will spend more time with it, testing, etc as soon as I am able.

micahsnyder (approved):
This looks great to me.  I tried to be as thorough as I could in reviewing the code changes. 
Testing went well (at least on the little endian devices we have in our pipelines).  Will merge.

sebastianas:
On 2023-03-17 11:31:57 [-0700], Micah Snyder wrote:
> Merged #814 into main.

Could this also go into the rel/1.0 branch? It is a regression coming
from the previous LTS (0.103.8).

Sebastian

micahsnyder:
Sure let's do that.  But I did find that we messed up something and coverity is angry:
image

And for the other variable:
image

I had put in a ticket for us to try to solve that  next week, but if you want to help you're welcome to.
When we get that fix we can cherry-pick both commits for PR targeting the `dev/1.0.2` branch.

PR #816, opened 2023-01-22
Add a version script for libclamunrar and libclamunrar_iface.

sebastianas:
Without a version script all symbols will be exported which are public within the libclamunrar* libraries. This is true for the officially exported symbols as well as all the public symbols which are used within libraries and are not inteded for export.

There is already a .map preset to limit the exported symbols of libclamunrar and libclamunrar_iface. Use it.

### DIFF
--- a/libclamunrar/CMakeLists.txt
+++ b/libclamunrar/CMakeLists.txt
@@ -70,7 +70,8 @@ if(ENABLE_SHARED_LIB)
     add_library( clamunrar SHARED )
     set_target_properties(clamunrar PROPERTIES
         VERSION ${LIBCLAMAV_VERSION}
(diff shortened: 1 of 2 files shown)


micahsnyder:
Thanks for submitting this Sebastian.  This week has been hectic and I expect that week will be as well. I will look at this/test it as soon as I'm able.

micahsnyder:
Since I closed your other version script PR ( #804 ) for the reasons mentioned there, I will close this as well.  

But I do see that #776  lacks the logic to use the libclamunrar.map.  I am testing adding that new in personal branch.  If that goes well, I'll just push it onto #776 along with a fix for the static library linking issue mentioned on that PR.

Thank you Sebastian for working on this and encouraging it even though I'm not merging these specific PR's.

Issue #750, opened 2022-11-08
clamdscan man page missing `-c` shorthand for `--config-file` option


micahsnyder:
`man clamdscan` does not list the `-c` short-hand for the `--config-file` option.

PR #821, opened 2023-01-28
Add -c short-hand in clamdscan help and man

b1tg:
fix https://github.com/Cisco-Talos/clamav/issues/750


micahsnyder:
Thanks @b1tg for the help. I made one tiny tweak, but otherwise looked good to me.

PR #822, opened 2023-02-02
clamonacc: add logging to watchpoint hierarchy add

m-sola:
Patch provides more insight into error conditions which may arise when adding a directory to the watch hierarchy. If a specific file caused the issue, the filename is provided to help users with any troubleshooting needed.

### DIFF
--- a/clamonacc/inotif/inotif.c
+++ b/clamonacc/inotif/inotif.c
@@ -202,15 +202,20 @@ static int onas_ddd_watch_hierarchy(const char *pathname, size_t len, int fd, ui
     struct onas_element *elem = NULL;
     int wd                    = 0;
 
-    if (onas_ht_get(ddd_ht, pathname, len, &elem) != CL_SUCCESS) return CL_EARG;
+    if (onas_ht_get(ddd_ht, pathname, len, &elem) != CL_SUCCESS) {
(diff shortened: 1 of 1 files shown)


clamonacc/inotif/inotif.c
```diff
@@ -236,13 +245,15 @@ static int onas_ddd_watch_hierarchy(const char *pathname, size_t len, int fd, ui
         size_t size      = len + strlen(curr->dirname) + 2;
         char *child_path = (char *)cli_malloc(size);
         if (child_path == NULL)
+            logg(LOGG_ERROR, "ClamInotif: out of memory when when adding child for %s\n", hnode->pathname);
```
micahsnyder, line 248: You're missing `{ }` curlies here, so it fails to watch any directories with CL_EMEM.

micahsnyder (changes requested):

micahsnyder:
There are also some places where `err` (a `cl_error_t` return code) is printed directly.  E.g.
```c
            logg(LOGG_ERROR, "ClamInotif: could not parse include list (%d)\n", err);
```
It would probably help make if we switch to `cl_strerror`, like:
```c
            logg(LOGG_ERROR, "ClamInotif: could not parse include list (%s)\n", cl_strerror(err));
```

micahsnyder (approved):

PR #835, opened 2023-02-13
Allow passing in custom RUSTFLAGS

opoplawski:
Also, according to https://cmake.org/cmake/help/latest/command/add_custom_command.html the `ARGS` option is ignored.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -236,7 +236,7 @@ function(add_rust_executable)
     # Build the executable.
     add_custom_command(
         OUTPUT "${OUTPUT}"
-        COMMAND ${CMAKE_COMMAND} -E env "CARGO_TARGET_DIR=${ARGS_BINARY_DIRECTORY}" ${cargo_EXECUTABLE} ARGS ${MY_CARGO_ARGS}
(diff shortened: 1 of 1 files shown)


micahsnyder:
Most of this makes sense to me, with the `string(APPEND RUSTFLAGS " -g")` and `string(STRIP "${RUSTFLAGS}" RUSTFLAGS)`.
Removing `ARGS` also makes sense, after looking at the docs to see that the minimum cmake version we require says it is optional/ignored. 

I am a little confused about one part of this change. Why do you want to remove the quotes around `${RUSTFLAGS}`?  It seems like that might break things if you have more than one flag.

micahsnyder (approved):
> The quotes caused problems. It seems like cmake automatically quotes the arguments.

Very interesting.  I didn't expect it to work this way, but I did some manual testing using options `-D RUSTFLAGS="-g -v"` and it seems good.  

We may also consider adding some documentation to the top of this file explaining that you can set RUSTFLAGS cmake variable before using this module. But I don't feel strongly about it.   I ran this through our automated tests a couple weeks ago and that looked good.  May as well merge now. 

Thanks for the PR @opoplawski

PR #837, opened 2023-02-14
Clam 1288 Signature regex filter search optimization

micahsnyder:
PDB, WDB, and CDB signatures can use a regex feature to match domain names. At one time in the process we ran a filter search to speed up filtering out non-matching static patterns but were accidentally discarding the result. When tested, it turns out it wasn't working correctly anyways.

Since then, we've fixed some bugs and upgraded the regex implementation to the latest version. After re-testing, I have found that the filter_search() appears to be working correctly now, both in the unit tests and with the existing CDB signatures.

### DIFF
--- a/libclamav/regex_list.c
+++ b/libclamav/regex_list.c
@@ -156,7 +156,7 @@ cl_error_t regex_list_match(struct regex_matcher *matcher, char *real_url, const
     char *buffer  = NULL;
     char *bufrev  = NULL;
     cl_error_t rc = CL_SUCCESS;
-    // int filter_search_rc = 0;
+    int filter_search_rc = 0;
     int root;
     struct cli_ac_data mdata;
     struct cli_ac_result *res = NULL;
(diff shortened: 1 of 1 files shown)


micahsnyder:
I went back to figure out when the test was fixed. I didn't spent a lot of time testing, but I did find that with this change version 0.104 has the following test failure:
```
2: Running suite(s): cl_suite
2:  cli
2:  jsnorm
2:  str
2:  regex
2:  disasm
2:  unique
2:  matchers
2:  htmlnorm
2:  bytecode
2: 99%: Checks: 1176, Failures: 1, Errors: 0
2: /home/micasnyd/workspace/clamav-micah/unit_tests/check_regex.c:295:F:regex_list:regex_list_match_test:13: regex_list_match
```
That failure is gone as of 0.105, and of course is not present in main either. 

Outside of some expected failures in the CI pipelines, everything looks good.

PR #838, opened 2023-02-15
Test: Check that a trusting a cert won't trust whole archive

micahsnyder:
This test scans a ZIP containing:
1. An executable signed with a trusted certificate
2. A file that will alert because of a signature match

The goal is to verify that the scan isn't terminated early and that the entire archive isn't trusted simply because the inner executable is trusted.
That is, that the "CL_VERIFIED" status does not propagate up out of the magic_scan of a specific layer.

Note: this is not a regression test. No such bug exists. I'm just adding it for safety going forward.


micahsnyder:
Rebased to see it pass the copyright tests and fuzz tests.

PR #841, opened 2023-02-18
Add cli options for all config options (except for deprecated ones)

abautu:
I added long-named CLI options for all config options that did not have one and are not deprecated. This would allow clamd to be started without a config file and configured from command line only.

### DIFF
--- a/common/optparser.c
+++ b/common/optparser.c
@@ -254,23 +254,23 @@ const struct clam_option __clam_options[] = {
 
     {"LogFile", "log", 'l', CLOPT_TYPE_STRING, NULL, -1, NULL, 0, OPT_CLAMD | OPT_MILTER | OPT_CLAMSCAN | OPT_CLAMDSCAN | OPT_CLAMONACC, "Save all reports to a log file.", "/tmp/clamav.log"},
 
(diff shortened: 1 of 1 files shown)


micahsnyder:
I appreciate the desire for this change.  It's a pretty big one, though. 

This change would also require:
- an update to the `--help` strings to add all of these new options for each of the affected programs
- an update to the manpages to add all of these new options for each of the affected programs

Before you put more work into this -- let me talk to the team about it to decide if we want to support this.  I'll get back to you.

micahsnyder (changes requested):
Thank you for your patience.  After speaking with the team about it and considering the effect, we decided we actually really like this change.  But I do have some requests. 

At first we considered asking you to add new options  to make it so  we can have all command line options mirror the config options, and then deprecate the old ones.  However, I think that is perhaps too much to ask.  

Another thing we considered was if it made sense to add a feature like `-C ConfigFileOption=Value` to make it so any config file option could be passed in as a command line option, and so we wouldn't clutter the `--help` strings and manpages with a lot more options.  However, this too seemed like a big ask and clunky considering we already support many if not most features in both command-line option and config option form.  

In the end, we thought it best to accept your approach, but with some minor corrections.  I think I found all of the issues with consistency with existing options and have added comments for each.

But in addition to the individual corrections, we will need to add all of the new command line options to the `--help` messages and theand manpage documentation files for the affected programs.  Can you please add these?

micahsnyder:
@abautu are you ready for re-review?  I see you merged some more changes into your PR.  Although since then a merge conflict has come up.

micahsnyder:
Also I did not see any answer to the comments I left on the PR.

PR #852, opened 2023-03-06
Clam 571 sigtool vba extraction improvements

ragusaa:
This PR deduplicates some vba extraction logic so sigtool works the same way libclamav does, and so it can extract more things.

### DIFF
--- a/libclamav/clamav.h
+++ b/libclamav/clamav.h
@@ -814,6 +814,27 @@ typedef int (*clcb_file_props)(const char *j_propstr, int rc, void *cbdata);
  */
 extern void cl_engine_set_clcb_file_props(struct cl_engine *engine, clcb_file_props callback);
 
+/**
+ * @brief generic data callback function.
+ *
+ * Callback handler prototype for callbacks passing back data and application context.
+ *
+ * @param data      A pointer to some data. Should be treated as read-only and may be freed after callback.
+ * @param data_len  The length of data.
+ * @param cbdata    Opaque application provided data.
+ */
+typedef int (*clcb_generic_data)(const unsigned char *const data, const size_t data_len, void *cbdata);
+
+/**
+ * @brief Set a custom VBA macro callback function.
+ *
+ * Caution: changing options for an engine that is in-use is not thread-safe!
+ *
+ * @param engine    The initialized scanning engine.
+ * @param callback  The callback function pointer.
+ */
+extern void cl_engine_set_clcb_vba(struct cl_engine *engine, clcb_generic_data callback);
+
 /* ----------------------------------------------------------------------------
  * Statistics/telemetry gathering callbacks.
  *
--- a/libclamav/others.h
+++ b/libclamav/others.h
@@ -416,6 +416,7 @@ struct cl_engine {
     void *cb_sigload_ctx;
     clcb_hash cb_hash;
     clcb_meta cb_meta;
+    clcb_generic_data cb_vba;
     clcb_file_props cb_file_props;
     clcb_progress cb_sigload_progress;
     void *cb_sigload_progress_ctx;
--- a/libclamav/vba_extract.c
+++ b/libclamav/vba_extract.c
@@ -1259,10 +1259,20 @@ cl_error_t cli_vba_readdir_new(cli_ctx *ctx, const char *dir, struct uniq *U, co
                         module_data_utf8_size = vba_normalize(module_data_utf8, module_data_utf8_size);
 
                         CLI_WRITEN(module_data_utf8, module_data_utf8_size);
+
+                        if (NULL != ctx->engine->cb_vba) {
+                            ctx->engine->cb_vba(module_data_utf8, module_data_utf8_size, ctx->cb_ctx);
+                        }
+
                         module_stream_found = 1;
                         free(module_data_utf8);
                         module_data_utf8 = NULL;
                     } else {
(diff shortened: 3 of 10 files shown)


micahsnyder:
In testing I tried using the ` --vba-hex=FILE ` option and it failed:
```

❯ ./install/bin/sigtool --vba-hex=$HOME/Downloads/1b96c0ad7ce83a573ec7770435655fa2c82c42ea43bd012dd0b64444941c9432
scanfile: Invalid args.
```

sigtool/sigtool.c
```diff
@@ -1566,6 +1573,404 @@ static int listsigs(const struct optstruct *opts, int mode)
...
+        }
+#endif
+        printf("%c", metaString[i]);
+    }
+    printf("\n");
+
+    return 0;
+}
+
+static bool setTempDir(struct cl_engine *engine, const struct optstruct *opts)
```
micahsnyder, line 1842: I wonder if there are other features in sigtool that could make use of setting the temp directory like this, and also the option to leaving temp files.

sigtool/sigtool.c
```diff
@@ -1566,6 +1573,404 @@ static int listsigs(const struct optstruct *opts, int mode)
...
+        if (rlim.rlim_cur < (rlim_t)cl_engine_get_num(engine, CL_ENGINE_MAX_SCANSIZE, NULL))
+            logg(LOGG_WARNING, "System limit for file size is lower than engine->maxscansize\n");
+    } else {
+        logg(LOGG_WARNING, "Cannot obtain resource limits for file size\n");
+    }
+#endif
+
+    /*Enable options that are needed to parse ole2 and ooxml files for vba macros.  */
+    options.parse |= CL_SCAN_PARSE_ARCHIVE;
+    options.parse |= CL_SCAN_PARSE_XMLDOCS;
```
micahsnyder, line 1937: Do we also need `CL_SCAN_PARSE_OLE2`?

sigtool/sigtool.c
```diff
@@ -1566,6 +1573,404 @@ static int listsigs(const struct optstruct *opts, int mode)
...
+    filename = optget(opts, "vba")->strarg;
+    if (NULL == filename) {
+        filename = optget(opts, "vba-hex")->strarg;
+    }
+    scanfile(filename, engine, opts, &options, &scanFiles);
+
+    for (sf = scanFiles.lst; NULL != sf;) {
+        scanfile(sf->fileName, engine, opts, &options, &scanFiles);
+        sf = removeScanFile(&scanFiles, sf);
+    }
```
micahsnyder, line 1958: cl_scandesc_callback is already a recursive process, so I would think that the process of adding files from the json to scan later would not be required.  Maybe we can message offline about it so I can understand this better

micahsnyder (changes requested):

sigtool/sigtool.c
```diff
@@ -1566,6 +1573,404 @@ static int listsigs(const struct optstruct *opts, int mode)
...
+        tempdir = opt->strarg;
+    } else {
+        tempdir = cli_gettmpdir();
+    }
+
+    /*
+     * We ALWAYS tell the engine to leave-temps, so that we can re-scan them to extract all the VBA, 
+     * so we create another temp subdirectory, so that we know what to delete when we go to clean them up.
+     */
+    if (!optget(opts, "leave-temps")->enabled) {
```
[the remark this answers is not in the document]
ragusaa: I like that idea.  I am find changing it, but the behavior will be different than clamscan.

libclamav/vba_extract.c
```diff
@@ -1255,6 +1255,10 @@ cl_error_t cli_vba_readdir_new(cli_ctx *ctx, const char *dir, struct uniq *U, co
 
                     close(module_fd);
 
+                    if (NULL != ctx->engine->cb_vba) {
+                        ctx->engine->cb_vba(module_data, module_data_size, ctx->cb_ctx);
+                    }
+
                     if (CL_SUCCESS == cli_codepage_to_utf8((char *)module_data, module_data_size, codepage, (char **)&module_data_utf8, &module_data_utf8_size)) {
                         module_data_utf8_size = vba_normalize(module_data_utf8, module_data_utf8_size);
```
ragusaa, line 1264: change callback to normalized data.

libclamav/clamav.h
```diff
@@ -814,6 +814,28 @@ typedef int (*clcb_file_props)(const char *j_propstr, int rc, void *cbdata);
...
+
+/**
+ * @brief Set a custom VBA macro callback function.
+ *
+ * Caution: changing options for an engine that is in-use is not thread-safe!
+ *
+ * @param engine    The initialized scanning engine.
+ * @param callback  The callback function pointer.
+ */
+extern void cl_engine_set_clcb_vba(struct cl_engine *engine, clcb_vba callback);
```
micahsnyder, line 837:
```suggestion
extern void cl_engine_set_clcb_vba(struct cl_engine *engine, clcb_generic_data callback);
```

libclamav/others.h
```diff
@@ -416,6 +416,7 @@ struct cl_engine {
     void *cb_sigload_ctx;
     clcb_hash cb_hash;
     clcb_meta cb_meta;
+    clcb_vba cb_vba;
```
micahsnyder, line 419:
```suggestion
    clcb_generic_data cb_vba;
```

micahsnyder (changes requested):

PR #867, opened 2023-03-22
Fix #820 - add new option --fail-if-cvd-older-than=days

### LINKED ISSUE #820: clamscan: new option --fail-when-outdated

rzvncj:
When passed, causes clamscan to exit with a non-zero return code if the virus database is older than the specified number of days.

### DIFF
--- a/clamscan/manager.c
+++ b/clamscan/manager.c
@@ -1258,6 +1258,13 @@ int scanmanager(const struct optstruct *opts)
 
     if ((opt = optget(opts, "database"))->active) {
         while (opt) {
+            if (optget(opts, "fail-if-cvd-older-than")->enabled) {
+                if (check_if_cvd_outdated(opt->strarg, optget(opts, "fail-if-cvd-older-than")->numarg) != CL_SUCCESS) {
+                    ret = 2;
+                    goto done;
+                }
+            }
+
             if ((ret = cl_load(opt->strarg, engine, &info.sigs, dboptions))) {
                 logg(LOGG_ERROR, "%s\n", cl_strerror(ret));
 
@@ -1269,6 +1276,13 @@ int scanmanager(const struct optstruct *opts)
         }
     } else {
         char *dbdir = freshdbdir();
+
+        if (optget(opts, "fail-if-cvd-older-than")->enabled) {
+            if (check_if_cvd_outdated(dbdir, optget(opts, "fail-if-cvd-older-than")->numarg) != CL_SUCCESS) {
+                ret = 2;
+                goto done;
+            }
+        }
 
         if ((ret = cl_load(dbdir, engine, &info.sigs, dboptions))) {
             logg(LOGG_ERROR, "%s\n", cl_strerror(ret));
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -106,6 +106,7 @@ static void help(void)
     printf("    --debug                                 Enable debug mode\n");
     printf("    --log=FILE               -l FILE        Log into FILE\n");
     printf("    --config-file=FILE       -c FILE        Read configuration from FILE\n");
+    printf("    --fail-if-cvd-older-than=days           Return with a nonzero error code if virus database outdated.\n");
     printf("\n");
     printf("Pass in - as the filename for stdin.\n");
     printf("\n");
@@ -651,6 +652,12 @@ int main(int argc, char **argv)
             svc_register("clamd");
         }
 #endif
+        if (optget(opts, "fail-if-cvd-older-than")->enabled) {
+            if (check_if_cvd_outdated(dbdir, optget(opts, "fail-if-cvd-older-than")->numarg) != CL_SUCCESS) {
+                ret = 1;
+                break;
+            }
+        }
 
         if ((ret = cl_load(dbdir, engine, &sigs, dboptions))) {
             logg(LOGG_ERROR, "%s\n", cl_strerror(ret));
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -255,6 +255,7 @@ void help(void)
     mprintf(LOGG_INFO, "                                         A JSON file will dropped to the temp directory if --leave-temps is enabled.\n");
     mprintf(LOGG_INFO, "    --database=FILE/DIR   -d FILE/DIR    Load virus database from FILE or load all supported db files from DIR\n");
     mprintf(LOGG_INFO, "    --official-db-only[=yes/no(*)]       Only load official signatures\n");
+    mprintf(LOGG_INFO, "    --fail-if-cvd-older-than=days        Return with a nonzero error code if virus database outdated.\n");
     mprintf(LOGG_INFO, "    --log=FILE            -l FILE        Save scan report to FILE\n");
     mprintf(LOGG_INFO, "    --recursive[=yes/no(*)]  -r          Scan subdirectories recursively\n");
     mprintf(LOGG_INFO, "    --allmatch[=yes/no(*)]   -z          Continue scanning within file after finding a match\n");
--- a/common/misc.c
+++ b/common/misc.c
@@ -44,7 +44,6 @@
 #include <errno.h>
 
 // libclamav
-#include "clamav.h"
 #include "cvd.h"
 #include "others.h" /* for cli_rmdirs() */
 #include "regex/regex.h"
@@ -486,4 +485,22 @@ unsigned int countlines(const char *filename)
 
     fclose(fh);
     return lines;
+}
+
+cl_error_t check_if_cvd_outdated(const char *path, long long days)
+{
+    cl_error_t status;
+    time_t cvd_age;
+
+    if ((status = cl_cvdgetage(path, &cvd_age)) != CL_SUCCESS) {
+        logg(LOGG_ERROR, "%s\n", cl_strerror(status));
+        return status;
+    }
+
+    if (days * 86400 < cvd_age) {
+        logg(LOGG_ERROR, "Virus database is older than %lld days!\n", days);
+        return CL_ECVD;
+    }
+
+    return CL_SUCCESS;
 }
--- a/common/misc.h
+++ b/common/misc.h
@@ -28,6 +28,7 @@
 #endif
 #include <stdbool.h>
 
+#include "clamav.h"
 #include "platform.h"
 #include "optparser.h"
 /* Maximum filenames under various systems - njh */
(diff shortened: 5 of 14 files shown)


rzvncj:
Please let me know if this is a good start. I'll continue with `FailIfCvdOlderThan` after we're good with the clamscan part (probably in its own commit).

clamscan/manager.c
```diff
@@ -1270,6 +1289,25 @@ int scanmanager(const struct optstruct *opts)
...
         char *dbdir = freshdbdir();
 
+        if (optget(opts, "fail-if-cvd-older-than")->enabled) {
+            time_t cvd_age;
+            const long long days = optget(opts, "fail-if-cvd-older-than")->numarg;
+
+            if ((ret = cl_cvdgetage(opt->strarg, &cvd_age)) != CL_SUCCESS) {
+                logg(LOGG_ERROR, "%s\n", cl_strerror(ret));
+
+                ret = 2;
```
micahsnyder, line 1299: I imagine you don't need to override `ret` here.

clamscan/manager.c
```diff
@@ -1258,6 +1258,25 @@ int scanmanager(const struct optstruct *opts)
...
     if ((opt = optget(opts, "database"))->active) {
         while (opt) {
+            if (optget(opts, "fail-if-cvd-older-than")->enabled) {
+                time_t cvd_age;
+                const long long days = optget(opts, "fail-if-cvd-older-than")->numarg;
+
+                if ((ret = cl_cvdgetage(opt->strarg, &cvd_age)) != CL_SUCCESS) {
+                    logg(LOGG_ERROR, "%s\n", cl_strerror(ret));
+
+                    ret = 2;
```
micahsnyder, line 1268: I imagine you don't need to override `ret` here.

micahsnyder (commented):
This is a great start! 

I have some suggested changes but nothing functional. 

I forgot to mention this on the previous PR, but in the spirit of #841, I would suggest adding both a `clamd.conf` option as well as a `--fail-if-cvd-older-than` command line option when you extend this for `clamd`. Which of course means adding documentatino as well to
- `clamd.conf.sample`
- `clamd.conf` manpage
- `clamd` manpage

clamscan/manager.c
rzvncj: I was just following the example below that line, with `cl_load()`, where it still gets set to `2` even though `ret` is non-zero. I don't know why that's been done, but I thought I shouldn't change behaviour. Of course I'm happy to remove the `ret = 2;` line if you think I should.
```c++
if ((ret = cl_load(opt->strarg, engine, &info.sigs, dboptions))) {
    logg(LOGG_ERROR, "%s\n", cl_strerror(ret));
                
    ret = 2;
    goto done;
}
```

clamscan/manager.c
rzvncj: Please see previous comment on overriding `ret`.

rzvncj: > I guess nevermind on this as per my reply to the previous comment. Sorry!

No problem at all, such is collaborating on software! :) I'll get rid of `CL_EOUTDATED_CVD` in that case (and just use the old generic `CL_ECVD`).

I guess nevermind on this as per my reply to the previous comment. Sorry!

rzvncj:
I assume the clamscan part is OK, so I'll start work shortly on the clamd part. Thanks for the quick review so far!

clamscan/manager.c
```diff
@@ -1031,6 +1031,24 @@ static int scan_files(struct cl_engine *engine, const struct optstruct *opts, st
...
+    }
+
+    if (days * 86400 < cvd_age) {
+        logg(LOGG_ERROR, "Virus database is older than %lld days!\n", days);
+        return CL_ECVD;
+    }
+
+    return CL_SUCCESS;
+}
+
```
micahsnyder, line 1051: If you want to put it in `common` you can minimize duplicate code when adding this feature to `clamd`.  Maybe in `common/misc.c` / `.h`.

micahsnyder:
As a heads up, we're prepping the 1.1 release candidate early next week.  It'd be fun to get this in there, but if not it will be included in 1.2 which will be released around August/September-ish.

rzvncj:
> As a heads up, we're prepping the 1.1 release candidate early next week. It'd be fun to get this in there, but if not it will have to wait for merge until after release in a few weeks, and be included in the following version (1.2) which will be released around August/September-ish.
> 
> Edit: To clarify, we don't like adding new features between the release candidate and release, though we may merge minor bug fixes during that time.

Understood. No problem, I'll do my best to get to the clamd part ASAP. Thanks!

rzvncj:
There's an additional clamd `cl_load()` call in server-th.c (in `reload_th()`) that I've not treated. I'm assuming that that code only runs after an actual database update so it would be pointless to guard it with `check_if_cvd_outdated()` but I could be wrong. Please let me know if that's something I should look at again. Thanks!

rzvncj:
Also, if you prefer this PR to be in a single commit please let me know and I'm happy to squash and force-commit the result.

micahsnyder:
> There's an additional clamd `cl_load()` call in server-th.c (in `reload_th()`) that I've not treated. I'm assuming that that code only runs after an actual database update so it would be pointless to guard it with `check_if_cvd_outdated()` but I could be wrong. Please let me know if that's something I should look at again. Thanks!

The `reload_th` use of `cl_load()` is also done on the self-check that runs  every x minutes (default every 10 minutes).
I can imagine someone complaining that they thought it would fail if the database hasn't been updated.  But I don't have any strong feels about whether or not is *should* fail to reload if the database ages past the threshold. 

If we do make the reload fail for old cvds, then to solve it I think you'd have to add an engine option to set the max database age.  That's maybe more work than you want to do?

If we don't make the reload fail for old cvds, then we should adjust the wording in the `clamd.conf.sample` and clamd.conf manpage to say something along the lines of "This will not cause a database reload to fail if the database found on reload is older than this limit."

> Also, if you prefer this PR to be in a single commit please let me know and I'm happy to squash and force-commit the result.

Don't worry about it, I will squash the commits and reword slightly when I merge it.

Speaking of which, I pushed a change to the win32 equivalent of the `clamd.conf.sample` for you, because I forgot to mention that it exists and must also be updated.

I'd like your opinion on what you think we should do regarding the `reload_th` `cl_load`.  And @candrews too if you have thoughts you're welcome to chime in.

micahsnyder (changes requested):
In terms of code and documentation, everything looks great to me. 

I manually tested clamd and clamscan behavior, including verifying that the clamd reload is not affected by the new options.  

Marking as "request changes" because we'll need either a change to the clamd.conf documentation if we choose to keep the reload behavior as-is, or because we'd need more clamd changes to support the option in the reload use case.

rzvncj:
> > There's an additional clamd `cl_load()` call in server-th.c (in `reload_th()`) that I've not treated. I'm assuming that that code only runs after an actual database update so it would be pointless to guard it with `check_if_cvd_outdated()` but I could be wrong. Please let me know if that's something I should look at again. Thanks!
> 
> The `reload_th` use of `cl_load()` is also done on the self-check that runs every x minutes (default every 10 minutes). I can imagine someone complaining that they thought it would fail if the database hasn't been updated. But I don't have any strong feels about whether or not is _should_ fail to reload if the database ages past the threshold.
> 
> If we do make the reload fail for old cvds, then to solve it I think you'd have to add an engine option to set the max database age. That's maybe more work than you want to do?
> 
> If we don't make the reload fail for old cvds, then we should adjust the wording in the `clamd.conf.sample` and clamd.conf manpage to say something along the lines of "This will not cause a database reload to fail if the database found on reload is older than this limit."

The amount of needed work is irrelevant to me, there are only two important factors here IMHO:

1. Do we want this to be done ASAP (so it makes it into the upcoming release before the freeze this week)?
2. Do users want this feature? I can't really tell, because it's not something I thought of using before seeing @candrews' open issue, so in this respect his opinion probably carries more weight than mine.

If it's OK to postpone this for the next release (i.e. the answer to the first question is "not necessarily"), and if it makes sense for the `cl_load()` thread to fail, then I have no problem to continue working on this.

> > Also, if you prefer this PR to be in a single commit please let me know and I'm happy to squash and force-commit the result.
> 
> Don't worry about it, I will squash the commits and reword slightly when I merge it.
> 
> Speaking of which, I pushed a change to the win32 equivalent of the `clamd.conf.sample` for you, because I forgot to mention that it exists and must also be updated.

Thanks! Sorry about missing the Windows part - I guess it shows I'm a UNIX person. :)

micahsnyder:
> The amount of needed work is irrelevant to me, there are only two important factors here IMHO:
> 
> 1. Do we want this to be done ASAP (so it makes it into the upcoming release before the freeze this week)?
> 2. Do users want this feature? I can't really tell, because it's not something I thought of using before seeing @candrews' open issue, so in this respect his opinion probably carries more weight than mine.
> 
> If it's OK to postpone this for the next release (i.e. the answer to the first question is "not necessarily"), and if it makes sense for the `cl_load()` thread to fail, then I have no problem to continue working on this.

I tested this change to the `reload_th` behavior to check the CVD age:
[cvd_age_check_clamd_reload.patch](https://github.com/Cisco-Talos/clamav/files/11083673/cvd_age_check_clamd_reload.patch)

What I realized with this patch is that failing the CVD update does not cause clamd to die.  Instead it intentionally continues on with the existing databases, which is very likely the preferred behavior.  

So then I thought about it a little more about the problem. I realize that the scenario I was trying to account for should not happen.  That is, it should only try to run the `reload_th` function if there was an update to the databases.  And if there was an update to the databases then the CVD age should not exceed the limit. 

In short, I'm happy with the PR as-is and don't care about any further documentation change after all. 

> Thanks! Sorry about missing the Windows part - I guess it shows I'm a UNIX person. :)

😆 No worries, I should have said something. 

I'll make sure this all passes our automated Jenkins tests, and if so will merge it.

micahsnyder (approved):
I had to make a small change because initializing arrays with `= {};` doesn't work on windows -- but you can initialize with `= {0};`.  Outside of that testing looked good.

micahsnyder:
Actually - one more change.  The new clamav.h API in libclamav.map should be in a different namespace: `CLAMAV_1.1.0`.

rzvncj:
Thanks!

micahsnyder:
Thank you so much for the help @rzvncj

candrews:
Thank you all! This is really fantastic!

rzvncj:
> Thank you so much for the help @rzvncj

You're welcome! It's been quick and pleasant. Thanks for the prompt reviews!

PR #868, opened 2023-03-22
Fix overread in js normalization

m-sola:


micahsnyder (approved):
This fix looks looks good to me.  Testing locally and in jenkins pipeline went well.

PR #840, opened 2023-02-18
libclamav: Use OpenSSL' BN instead tomfastmath.

sebastianas:
Use OpenSSL's big number/ multiprecision integer arithmetics
functionality to replace tomfastmath.

This is a first shot at doing just this. Further improvement could be
use more RSA-signature verification from OpenSSL in crtmgr_rsa_verify()
and less self parsing.
_padding_check_PKCS1_type_1() has been borrowed from OpenSSL to make
further replacements easier.

This would make the tomfastmath pull obsolete ;)

PR #871, opened 2023-03-25
Clam 2291 rc soversion news

micahsnyder:
Set release candidate version suffix and bump SOVERSION 

Add news for recent 1.0.1, 0.105.2, and 0.103.8 patch versions

Update news for 1.1.0 release

Also small random fix:
- CMake: Fix issue generating Rust bindings with -D MAINTAINER_MODE=ON


micahsnyder:
@ragusaa TBH I wondered about that too.  We haven't touched that line in a couple years, and when I looked at the "Properties" page for `clamscan.exe`, `sigtool.exe`, etc I found that they don't have the `-devel` suffix in there.  So I think maybe it isn't even used??  I wonder if we can delete it entirely, or else fix it somehow so it does show up somewhere.  Or maybe I'm looking in the wrong place.

ragusaa (approved):

micahsnyder:
I'll update this with changes to the NEWS.md after:
1. we merge #852, #867, and #840 
2. our review of the release notes (in sharepoint) is complete

So maybe this evening or tomorrow I hope

Issue #731, opened 2022-10-24
clamd.conf manpage missing `DisableCache` config option


micahsnyder:
The `clamd.conf` manpage is missing the `DisableCache` config option. 

Documentation issue observed by Discord user, here: https://discord.com/channels/636023333074370595/1034157716807094363

micahsnyder:
Related: the `clamdscan` manpage is missing the `-c` shorthand for `--config-file` option. See https://github.com/Cisco-Talos/clamav/issues/731

candrews:
> Related: the `clamdscan` manpage is missing the `-c` shorthand for `--config-file` option.

That option is now documented: https://github.com/Cisco-Talos/clamav/commit/f227bc5ffcefd4c216b487d81268465cdc6a0075

PR #878, opened 2023-03-29
fix: Document DisableCache in clamd.conf man page

candrews:
Closes: https://github.com/Cisco-Talos/clamav/issues/731


micahsnyder:
Thanks for this! Nice find.

micahsnyder (approved):

PR #881, opened 2023-03-31
unit_tests/freshclam_test.py: fix unit test

arjendekorte:
Somehow the tests returns a different result than expected. Not sure if the expected result should be changed (this patch) or that the message emitted by freshclam should be changed.

### DIFF
--- a/unit_tests/freshclam_test.py
+++ b/unit_tests/freshclam_test.py
@@ -671,7 +671,7 @@ class TC(testcase.TestCase):
         assert output.ec == 0  # success
 
         expected_stdout = [
-            'already up-to-date',
+            'database is up-to-date',
         ]
         unexpected_results = [
             'test.cld updated \\(version: 6',


arjendekorte:
String to match is generated in libfreshclam/libfreshclam_internal.c line 2144

opoplawski:
This certainly does resolve the test failure.  Still to be determined I suppose if this is the correct fix.

arjendekorte:
Indeed. Changing line 2144 in libfreshclam/libfreshclam_internal.c will have the same result, but I have no idea either whether or not this is intentionally different.

micahsnyder:
After thinking about this for some time, I suspect you are carrying a patch that makes Debug-level messages go to `stderr`.  This test is checking for a Debug-level message in `stdout`.  Can you confirm?

@opoplawski's proposed change switches to look for an Info-level message in `stdout` that is roughly equivalent. I'm also happy with this change.

micahsnyder (approved):

micahsnyder:
Before I merge, if you can confirm the debug/stderr suspicion that may allow us to fix up the commit message.

opoplawski:
That patch fixes the test for me as well.  FWIW - my ctest line is:
```
+ /usr/bin/ctest --test-dir redhat-linux-build --output-on-failure --force-new-ctest-process -j4 -- -E valgrind
```

micahsnyder:
I merged the commit and updated the commit message to explain the reason for the change and the outstanding issue. I am still curious to solve why debug-level messages aren't bring printed for you.

PR #884, opened 2023-04-03
Update cbindgen 0.24

opoplawski:
Not quite sure what the process for updating rust deps is.  But we're building clamav okay in Fedora against 0.24.3.


micahsnyder:
This change makes sense to me.
I think the Cargo.lock file also has to be updated.  I can do that and tack it on to the PR.

micahsnyder (approved):
Oops, I forgot to check on the test results and merge this.  Looks good, will merge.

PR #886, opened 2023-04-03
Don't set parent's "don't cache" to match child

candrews:
Don't set the parent layer's "don't cache" flag to match the child.

`emax_reached()` already does the same thing so doing it again is unnecessary.

See https://github.com/Cisco-Talos/clamav/commit/db013a2bfda869072f72f22441e9be21c08580f7 to see when the code that this commit removes was added.

### DIFF
--- a/libclamav/matcher.c
+++ b/libclamav/matcher.c
@@ -803,10 +803,6 @@ cl_error_t cli_scan_desc(int desc, cli_ctx *ctx, cli_file_t ftype, bool filetype
 
     status = cli_scan_fmap(ctx, ftype, filetype_only, ftoffset, acmode, acres, NULL);
 
-    map->dont_cache_flag = ctx->fmap->dont_cache_flag; /* Set the parent layer's "don't cache" flag to match the child.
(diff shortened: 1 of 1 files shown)


micahsnyder:
Ah I see you feel the same as I felt when writing the comment: 
> `TODO: This may not be needed since `emax_reached()` should've already done that for us.`

Yeah I think it should be safe to merge your PR.  It's a very minor optimization and shouldn't change the behavior at all.  Right?

candrews:
>  It's a very minor optimization

It increases cacheability which can be a major optimization :-)

>  and shouldn't change the behavior at all. Right?

Yes, correct.

candrews:
You are, of course, correct, and I was mistaken.

I got this PR confused with another effort - I apologize for the mistake.

This PR is, as you said, a very minor optimization and shouldn't change the behavior at all, and I believe it is safe to merge.

micahsnyder (approved):
I did a little bit more testing and confirmed I'm very comfortable merging this PR.

PR #888, opened 2023-04-04
Jenkins: Moving appcheck-pipeline to daily triggers

rsundriyal:
We are moving our appcheck pipeline from ClamAV project and instead will trigger them once a day.
Right now, it triggers with every PR and merges to the main branch.


micahsnyder (approved):
Looks good, pipeline ran as expected.  We should backport this for:
- `dev/1.0.2`
- `dev/0.105.3`
- `dev/0.103.9`

PR #892, opened 2023-04-11
Clam 2269 cdiff unlink

micahsnyder:
Fix issue applying cdiffs that unlink a file (e.g. dropping a CBC signature from bytecode.cvd). 
Also some minor code cleanup to resolve linter errors.


micahsnyder:
Moved commits from this over to https://github.com/Cisco-Talos/clamav/pull/893 because there were some additional related bugs discovered when working on that and it made sense to fix everything in one PR to get all the tests to pass.

PR #893, opened 2023-04-11
Clam 2269 & 2271: CDIFF: unlink, close issues. Freshclam: download whole CVD if failure applying CDIFF

micahsnyder:
In the event that there is an issue with the CDIFF process, freshclam is treating it as thought no patch was downloaded.

If freshclam fails to apply the patch because of an issue with the patch, or some bug in the CDIFF module, it should retry for the whole CVD.

### DIFF
--- a/libclamav_rust/src/cdiff.rs
+++ b/libclamav_rust/src/cdiff.rs
@@ -183,6 +183,9 @@ pub enum InputError {
 
     #[error("no final newline")]
     MissingNL,
+
+    #[error("Database file is still open: {0}")]
+    DBStillOpen(String),
 }
 
 /// Errors encountered while processing
@@ -283,6 +286,32 @@ impl<'a> DelOp<'a> {
             .ok_or(InputError::MissingParameter("DEL", "orig_line"))?;
 
         Ok(DelOp { line_no, del_line })
+    }
+}
+
+#[derive(Debug)]
+pub struct UnlinkOp<'a> {
+    db_name: &'a str,
+}
+
+/// Method to parse the cdiff line describing an unlink operation
+impl<'a> UnlinkOp<'a> {
+    pub fn new(data: &'a [u8]) -> Result<Self, InputError> {
+        let mut iter = data.split(|b| *b == b' ');
+        let db_name = str::from_utf8(
+            iter.next()
+                .ok_or(InputError::MissingParameter("UNLINK", "db_name"))?,
+        )
+        .map_err(|_| InputError::NotUnicode("database name"))?;
+
+        if !db_name
+            .chars()
+            .all(|x: char| x.is_alphanumeric() || x == '\\' || x == '/' || x == '.')
+        {
+            return Err(InputError::ForbiddenCharactersInDB(db_name.to_owned()));
+        }
+
+        Ok(UnlinkOp { db_name })
     }
 }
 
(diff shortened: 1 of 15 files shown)


ragusaa:
Code looks good to me, however tests 8 and 9 fail when I run ctest.

micahsnyder:
> Code looks good to me, however tests 8 and 9 fail when I run ctest.

OH NO! How'd I miss that? Will look into it and get back to you.

unit_tests/sigtool_test.py
```diff
@@ -79,3 +81,78 @@ def test_sigtool_00_version(self):
...
+
+        # Apply 1st cdiff.
+
+        command = '{valgrind} {valgrind_args} {sigtool} --run-cdiff={cdiff}'.format(
+            valgrind=TC.valgrind, valgrind_args=TC.valgrind_args, sigtool=TC.sigtool,
+            cdiff=TC.path_source / 'unit_tests' / 'input' / 'freshclam_testfiles' / 'test-2.cdiff'
+        )
+        output = self.execute_command(command)
+
+        # Apply 2nd cdiff. This one failed because of the UNLINK bug.
```
micahsnyder, line 113: comment is wrong, this is from the CLOSE bug for new files.

unit_tests/sigtool_test.py
```diff
@@ -79,3 +81,78 @@ def test_sigtool_00_version(self):
...
+            cdiff=TC.path_source / 'unit_tests' / 'input' / 'freshclam_testfiles' / 'test-3.cdiff'
+        )
+        output = self.execute_command(command)
+
+        assert output.ec == 1  # failure
+
+        # Verify that the `error!()` message was printed to stderr.
+        # This didn't happen before when we didn't initialize the rust logging lib at the top of sigtool.
+        expected_results = [
+            'LibClamAV Error'.format(TC.version),
```
micahsnyder, line 156: the `.format()` is not required. No paramters!

libclamav_rust/src/cdiff.rs
```diff
@@ -286,6 +289,32 @@ impl<'a> DelOp<'a> {
...
+        let mut iter = data.split(|b| *b == b' ');
+        let db_name = str::from_utf8(
+            iter.next()
+                .ok_or(InputError::MissingParameter("UNLINK", "db_name"))?,
+        )
+        .map_err(|_| InputError::NotUnicode("database name"))?;
+
+        if !db_name
+            .chars()
+            .all(|x: char| x.is_alphanumeric() || x == '\\' || x == '/' || x == '.')
```
micahsnyder, line 309: the `\\` and `/` should NOT be allowed, but the alphanumeric or `.` SHOULD be allowed. 
This is copypaste from a check elsewhere in the `cdiff.rs`. we can check if the filename is purely just `.` or `..` (that's not in the original cdiff code, but we could add it).

micahsnyder: Consider switching this to pathbuf checks for a parent with std::path.

libclamav_rust/src/cdiff.rs
```diff
@@ -97,8 +97,8 @@ pub enum CdiffError {
     ///
     /// This error *may* wrap a processing error if the command has side effects
     /// (e.g., MOVE or CLOSE)
-    #[error("{1} on line {0}")]
-    Input(usize, InputError),
+    #[error("{1} on line {0}: {2}")]
+    Input(usize, InputError, String),
```
micahsnyder, line 101:
```rust
Input {
   line: usize,
   err: InputError,
   operation: String,
}
```

will have to go fix them all anywhere `Input` is used.

shutton (changes requested):
Changes as discussed live.

libclamav_rust/src/cdiff.rs
```diff
@@ -960,7 +1027,10 @@ fn process_line(ctx: &mut Context, line: &[u8]) -> Result<(), InputError> {
             cmd_move(ctx, move_op)
         }
         b"CLOSE" => cmd_close(ctx),
-        b"UNLINK" => cmd_unlink(ctx),
+        b"UNLINK" => {
+            let unlink_op = UnlinkOp::new(remainder.unwrap())?;
+            cmd_unlink(ctx, unlink_op)
+        }
         _ => Err(InputError::UnknownCommand(
             String::from_utf8(cmd.to_owned()).unwrap(),
```
shutton, line 1035: Not your code, but there's the potential to thrown an error while building the error here. I would recommend using `::from_utf8_lossy()` to avoid panicking if the content is non-UTF-8

shutton (changes requested):
Just that one utf8 conversion issue

ragusaa (approved):

libclamav_rust/src/cdiff.rs
micahsnyder: Good idea.  I've made the change.  Merging.

micahsnyder:
Just fixed some python 3.5 compatibility issues because python 3.6 isn't available on all machines we test on, even though python 3.5 is past enough of life.  Sigh.  Anyways all looked good now in the test pipeline.  Will merge.

PR #894, opened 2023-04-12
2287 - fix ole2 vba temp file leak

m-sola:
Previous behaviour would remove temp files by deleting the subdirectory This caused issues in cases (on Windows) where subdirectories aren't created due to performance concerns

This commit removes tempfiles individually if keeptemp is off

Original patch authored by Thomas Vy

### DIFF
--- a/libclamav/scanners.c
+++ b/libclamav/scanners.c
@@ -1633,6 +1633,7 @@ static cl_error_t cli_ole2_tempdir_scan_vba_new(const char *dir, cli_ctx *ctx, s
     char path[PATH_MAX];
     char filename[PATH_MAX];
     int tempfd = -1;
+    char * tempfile = NULL;
 
     if (CL_SUCCESS != (ret = uniq_get(U, "dir", 3, &hash, &hashcnt))) {
         cli_dbgmsg("cli_ole2_tempdir_scan_vba_new: uniq_get('dir') failed with ret code (%d)!\n", ret);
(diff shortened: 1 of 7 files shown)


libclamav/scanners.c
```diff
@@ -1633,7 +1633,7 @@ static cl_error_t cli_ole2_tempdir_scan_vba_new(const char *dir, cli_ctx *ctx, s
     char path[PATH_MAX];
     char filename[PATH_MAX];
     int tempfd = -1;
-
+    char * tempfile = NULL;
```
micahsnyder, line 1636: Could use a new line after this, and clang-format to align the `=`.

libclamav/scanners.c
```diff
@@ -1703,6 +1709,11 @@ static cl_error_t cli_ole2_tempdir_scan_vba_new(const char *dir, cli_ctx *ctx, s
         close(tempfd);
         tempfd = -1;
     }
+    if (tempfile && !ctx->engine->keeptmp) {
```
micahsnyder, line 1712: Only removing the tempfile should be wrapped with the `keeptmp` check. We still need to free the variable and set to NULL either way. 
E.g. like is done in `xlm_extract.c`:
```c
    if (tempfile && !ctx->engine->keeptmp) {
        remove(tempfile);
    }
    FREE(tempfile);
```

libclamav/scanners.c
```diff
@@ -1690,9 +1690,15 @@ static cl_error_t cli_ole2_tempdir_scan_vba_new(const char *dir, cli_ctx *ctx, s
             if (CL_SUCCESS != ret) {
                 goto done;
             }
-
+            
             close(tempfd);
             tempfd = -1;
+            if (tempfile && !ctx->engine->keeptmp) {
```
micahsnyder, line 1696: Would be nice to have a newline before the `if()`, for readability.

micahsnyder (changes requested):

micahsnyder:
Looks great.  Good idea @m-sola doing a full clang-format and doing as as separate commit.  Looks like other format issues slipped throughout dev and this cleans them all up.

micahsnyder (approved):

PR #897, opened 2023-04-18
clam-2299 fix memcpy negative size param

m-sola:
During fuzzing, junk data could result in an underflow when calculating chunk size during html_normalization

### DIFF
--- a/libclamav/htmlnorm.c
+++ b/libclamav/htmlnorm.c
@@ -1824,6 +1824,11 @@ static bool cli_html_normalise(cli_ctx *ctx, int fd, m_area_t *m_area, const cha
         }
 
         if (in_tag == TAG_STYLE) {
+            if (ptr < style_begin) {
+                cli_errmsg("cli_html_normalise: style chunk size underflow\n");
+                goto done;
+            }
+
             size_t chunk_size = ptr - style_begin;
 
             if (style_buff == NULL) {


libclamav/htmlnorm.c
```diff
@@ -1824,8 +1824,13 @@ static bool cli_html_normalise(cli_ctx *ctx, int fd, m_area_t *m_area, const cha
...
         if (in_tag == TAG_STYLE) {
-            size_t chunk_size = ptr - style_begin;
-
+            size_t chunk_size = 0;
+            if ((int64_t)(ptr - style_begin) > 0) {
+                chunk_size = ptr - style_begin;
+            } else {
+                cli_errmsg("cli_html_normalise: style chunk size underflow\n");
+                goto done;
+            }
```
micahsnyder, line 1833: Instead of having it underflow and then casting to a signed integer to verify the underflow, it would be better to do this:
```suggestion

            if (ptr < style_begin) {
                cli_errmsg("cli_html_normalise: style chunk size underflow\n");
                goto done;
            }

            chunk_size = ptr - style_begin;

```

Outside of this I think your solution is good.  

I wrote the new code where the bug occurs so it didn't take me too long to get into the right headspace for it.  But I spent a few hours trying to understand how it got into this state and figure out if this is the only required fix.

In short, the code just after this is the problem.

It has this logic that "rewinds" the `ptr` variable to a point b efore the `style_begin`:
```c
        if (look_for_screnc && ptr_screnc) {
            /* start found, and stuff before it already processed */
            ptr        = ptr_screnc;
            ptr_screnc = NULL;
            state      = HTML_JSDECODE_LENGTH;
            next_state = HTML_BAD_STATE;
            continue;
        }
```

The intention of that rewind is to go revisit the stuff after the `"#@~^"` marker. It reads the length of the encoded javascript and attempts to decode it.   But the problem is that after it's done with decoding, it continues on at a place *before* where it left off.   So that's why `ptr` < `style_begin`.  

In my opinion, the "rewind and decode" logic is ill conceived.  It should have decoded it right away, or else called a function that decodes it.  It should never rewind the pointer.  

But I think rewriting how that all works out outside the scope of this fix.  What you have is good to resolve the fuzz bug.  When will we rework that?  Probably never.  Unless there is an htmlnorm rewrite in Rust.

micahsnyder (changes requested):
Other than my above suggestion and otherwise I think this approach to fixing the issue is good.

micahsnyder (approved):
Tests look good.

PR #898, opened 2023-04-19
Jenkins: switch to ubuntu for package stage for newer Rust. CMake-Rust: vendor dependencies in top level directory

micahsnyder:
### Problem statement

Adding Cargo.lock file caused the cmake tarball build that we run on Jenkins to fail unless we ran `cargo update` directly before (as seen in the Jenkinsfile).

We wanted to make it "just work" and so set out to update the FindRust.cmake to run `cargo update` for us as a part of the build.

### Summary of this PR

Initially we wanted to update the FindRust.cmake to run `cargo update` for us as a part of the build.  What we didn't realize is that `cargo update` was in fact modifying the Cargo.lock file and rolling BACK the dependency version numbers to match what Cargo thought was available for the current version of Cargo/Rust. 

My understanding now is that the reason the build was failing to begin with wasn't because it was failing to update, but because Rust itself was too old to build the latest version of the `image` crate.  So now this PR has been updated to do two things:

1. Don't run `cargo update`.  Instead, use a different Jenkins node that has access to a more recent version of Rust.
2. Change how we vendor the dependencies. Vendor them at the top level `SRC/.cargo/vendor` **instead** of in a `.cargo/vendor` subdirectory for **each** Rust crate. This second part is only because I happened to make this improvement while working on the FindRust.cmake and not because it is actually required to fix the issue. 

Regarding the 2nd change (the more complicated part of the PR):  This change was required to be able to easily check if the Rust dependencies were already vendored to know if we needed to do an "offline build" or if we should run cargo update.   We're not doing that after all, so the 2nd change is not technically required.  But it is an improvement, which is why I thought I'd keep it in the PR.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,7 +10,7 @@
 *.ref           binary
 
 # Preserve signature for .cargo/vendor files (from the tarabll)
-+/.cargo/vendor binary
+.cargo/vendor   binary
 
 #
 # Exclude files from exporting
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -47,7 +47,7 @@ properties(
     ]
 )
 
-node('master') {
+node('ubuntu-18-x64') {
     stage('Generate Tarball') {
         cleanWs()
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -68,7 +68,6 @@ node('master') {
 
         dir(path: 'build') {
             sh """# CPack
-                cargo update
                 cmake .. -D VENDOR_DEPENDENCIES=ON
                 cpack --config CPackSourceConfig.cmake """
             archiveArtifacts(artifacts: "clamav-${params.VERSION}*.tar.gz", onlyIfSuccessful: true)
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -184,28 +184,28 @@ function(cargo_vendor)
     set(oneValueArgs TARGET SOURCE_DIRECTORY BINARY_DIRECTORY)
     cmake_parse_arguments(ARGS "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
-    if(NOT EXISTS ${ARGS_SOURCE_DIRECTORY}/.cargo/config.toml)
-        # Vendor the dependencies and create .cargo/config.toml
-        # Vendored dependencies will be used during the build.
-        # This will allow us to package vendored dependencies in source tarballs
-        # for online builds when we run `cpack --config CPackSourceConfig.cmake`
-        message(STATUS "Running `cargo vendor` to collect dependencies for ${ARGS_TARGET}. This may take a while if the local crates.io index needs to be updated ...")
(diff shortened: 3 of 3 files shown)


shutton (approved):
LGTM, so long as you're seeing what you expect with the Cargo cache.

micahsnyder:
The effort to force cargo to update online has failed.  `cargo update` still fails on our FreeBSD machine wtih the added `Cargo.lock` file, like it's not even trying to check for newer versions of the packages.

The specific issue we're trying to fix is this error message when trying to build or vendor the dependencies with the `Cargo.lock` file where the version specified in the lock file newer than is known about in the local cache of the crates.io registry index.

The error looks like this:
```
  Failed!
      Updating crates.io index

  error: failed to sync

  Caused by:
    failed to load pkg lockfile

  Caused by:
    failed to select a version for the requirement `image = "^0.24"`
    candidate versions found which didn't match: 0.24.5, 0.24.4, 0.24.3, ...
    location searched: crates.io index
    required by package `clamav_rust v0.0.1 (/usr/local/jenkins/workspace/clamav_PR-898/libclamav_rust)`
```

Frankly we shouldn't need this "fix" anyways. But it seems that `cargo vendor` (and presumably `cargo build`) aren't checking for newer versions of the packages on our FreeBSD machines and then failing because the package version required in `Cargo.lock` is newer than is seen in the local cache of the crates.io index.  

Even with this change, it seems that even `cargo update` isn't using the internet to update the local cache of the crates.io registry (i.e. `$HOME/.cargo/registry`) in our FreeBSD machines.  I blame the FreeBSD-port of `cargo`.  Because this doesn't happen anywhere else I've tested.

I believe the solution if someone encounters this is probably to clear the cargo cache and try again.  But that seems too heavy handed for us to do as a part of our build system.  You can do it like:
```
cargo install cargo-cache
cargo cache -a -r registry-index
```
In theory we could detect the `failed to load pkg lockfile` message in the failure and try to do the above , but that feels like too much.

I've opted to try adding these commands to our Jenkinsfile.   Others may need it as well. 

I will add an FAQ explaining this situation in case anyone else runs into it.

micahsnyder:
Update: I had trouble with clearing the cache.  Still reproduced even after manually clearing the cache.  Instead, I've tried something novel... using Ubuntu instead of FreeBSD.  Works like a charm. 

The problem may have been that our FreeBSD box is stuck Rust v1.58.0 and that the image crate is requiring v1.61.0 (https://github.com/image-rs/image/commit/11492233cd59fc0e2701ab12cd00607e9bbf125c).   If so, that's one godawful error message.   If not, it's an awful message to... so... 🤷 

Anyways, Using Ubuntu is letting us build the package without upgrading the FreeBSD box on short notice to get access to a newer version of the rust package.

@shutton  if you could please take a look now at what I've done, you can check the individual commit messages for a good description of each. I think I'm happy with this now.

micahsnyder:
Canceled in favor of https://github.com/Cisco-Talos/clamav/pull/907.

PR #902, opened 2023-04-20
Clam 2269 2271 cdiff issues 0.105

micahsnyder:
backport of https://github.com/Cisco-Talos/clamav/pull/893 for 0.105


micahsnyder:
Oops! Nevermind. Publishing 1.1. means that 0.105 is past end of life for new patch versions as per the EOL policy.

PR #904, opened 2023-04-25
tools: update Github Actions workflows

VoltrexKeyva:
- Update the actions to their latest versions.
- Update and pin the platform versions.
- Use Clang-Format 16.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,40 +2,65 @@
 Language:        Cpp
 AccessModifierOffset: -2
 AlignAfterOpenBracket: Align
-AlignConsecutiveMacros: false
-AlignConsecutiveAssignments: true
-AlignConsecutiveBitFields: false
-AlignConsecutiveDeclarations: false
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments:
+  Enabled:         true
+  AcrossEmptyLines: false
+  AcrossComments:  false
+  AlignCompound:   false
+  PadOperators:    true
+AlignConsecutiveBitFields:
+  Enabled:         false
+  AcrossEmptyLines: false
+  AcrossComments:  false
+  AlignCompound:   false
+  PadOperators:    false
+AlignConsecutiveDeclarations:
+  Enabled:         false
+  AcrossEmptyLines: false
+  AcrossComments:  false
+  AlignCompound:   false
+  PadOperators:    false
+AlignConsecutiveMacros:
+  Enabled:         false
+  AcrossEmptyLines: false
+  AcrossComments:  false
+  AlignCompound:   false
+  PadOperators:    false
 AlignEscapedNewlines: Left
 AlignOperands:   Align
-AlignTrailingComments: true
+AlignTrailingComments:
+  Kind:            Always
+  OverEmptyLines:  0
 AllowAllArgumentsOnNextLine: true
-AllowAllConstructorInitializersOnNextLine: true
 AllowAllParametersOfDeclarationOnNextLine: true
-AllowShortEnumsOnASingleLine: true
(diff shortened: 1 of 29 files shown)


micahsnyder:
Thank you for trying this update.  I just approved running the pipeline.  Looks like there are some failures though. 

If I remember correctly the interfaces changed for some of the newer versions of the actions so they have to be updated to work with the new version.  Are you able to work on this?

VoltrexKeyva:
> Thank you for trying this update.  I just approved running the pipeline.  Looks like there are some failures though. 
> 
> If I remember correctly the interfaces changed for some of the newer versions of the actions so they have to be updated to work with the new version.  Are you able to work on this? 

Sure, I'll work on this.

VoltrexKeyva:
Not sure why but some Clang-Format CIs are complaining about formatting issues even though running the formatter doesn't change any of those files it errors on.

micahsnyder:
It's entirely likely that clang-format 16 does something slightly different than clang-format 14.  They are constantly breaking behavior with clang-format. I'll see if I can figure out an alteration to those files that makes both versions happy.

micahsnyder:
@VoltrexKeyva Can we revert to using clang-format version 14?

I couldn't find clang-format-16 from any linux distros (checked alpine, ubuntu, opensuse, fedora).  
I tested with clang-format-15 and it tried to make some very dubious changes to the source (misaligning variable assignment `=`)

I think clang-format-14 is pretty stable.

VoltrexKeyva:
> I couldn't find clang-format-16 from any linux distros (checked alpine, ubuntu, opensuse, fedora).

You can check https://apt.llvm.org/ for instructions on how to get LLVM 16.

> I tested with clang-format-15 and it tried to make some very dubious changes to the source (misaligning variable assignment `=`)

I believe there are breaking changes in the configuration part of Clang-Format's styles, here are the changes that occurred:

- Clang-Format 15: https://releases.llvm.org/15.0.0/tools/clang/docs/ReleaseNotes.html#clang-format
- Clang-Format 16: https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#clang-format

It's most likely caused by this change in Clang-Format 15:

> **Important change**: Renamed `IndentRequires` to `IndentRequiresClause` and changed the default for all styles from `false` to `true`

The default for all styles being changed from `false` to `true` most likely being the culprit.

micahsnyder:
I tried pushing up the `.clang-format` file changes generated by clang-format-16 on my box, and for some reason the clang-format job fails because it doesn't understand the options in the new `.clang-format` file. 
Doesn't seem like the job is actually using clang-format-16. Very odd.

VoltrexKeyva:
> I tried pushing up the `.clang-format` file changes generated by clang-format-16 on my box, and for some reason the clang-format job fails because it doesn't understand the options in the new `.clang-format` file. 
> Doesn't seem like the job is actually using clang-format-16. Very odd. 

After looking through the source code of the `jidicula/clang-format-action`, it seems like it tries to install Clang-Format 16 from the official Ubuntu repositories, but they don't include it.

Maybe we could try [`DoozyX/clang-format-lint-action`](https://github.com/DoozyX/clang-format-lint-action)?

micahsnyder:
Sure definitely could try switching to DoozyX's action.  

We'll need to set up the `.clang-format-ignore` in place of these exclusion rules https://github.com/Cisco-Talos/clamav/blob/main/.github/workflows/clang-format.yml#L23-L57

micahsnyder:
Closing because it was never quite solved and is inactive.  Feel free to submit a new PR if you want to work on it again in the future.

PR #906, opened 2023-04-26
Clam 2315 remove rc suffix 2316 news

micahsnyder:


ragusaa (approved):

ragusaa (approved):

micahsnyder:
I dropped the cargo-vendor change from the notes because of issues in testing https://github.com/Cisco-Talos/clamav/pull/898.  So we're going with a subset of that change that simply gets the build to work again and without modifying the Cargo.lock file when generating the tarball.  So that's https://github.com/Cisco-Talos/clamav/pull/907.

PR #911, opened 2023-04-28
Fix possible crash in HTML CSS image extraction

micahsnyder:
When processing UTF-8 HTML code, the image extraction logic may panic if the string contains a multi-byte grapheme that includes a '(', ')', whitespace, or one of the other characters used to split the text when searching for the base64 image content.

The panic is because the `split_at()` method will panic if you try to split in the middle of a unicode grapheme.

This commit fixes the issue by processing the HTML string one grapheme at a time instead of one character (byte) at a time. The `grapheme_indices()` method is used to get the correct position of the start of each grapheme for splitting the string.

### DIFF
--- a/libclamav_rust/src/css_image_extract.rs
+++ b/libclamav_rust/src/css_image_extract.rs
@@ -25,6 +25,7 @@ use std::{ffi::CStr, mem::ManuallyDrop, os::raw::c_char};
 use base64::{engine::general_purpose as base64_engine_standard, Engine as _};
 use log::{debug, error, warn};
 use thiserror::Error;
+use unicode_segmentation::UnicodeSegmentation;
 
 use crate::sys;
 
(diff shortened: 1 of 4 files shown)


shutton (approved):
Rust bits look fine.

micahsnyder:
Force-push is just a rebase. No code change.

libclamav_rust/src/css_image_extract.rs
```diff
@@ -90,16 +91,17 @@ impl<'a> CssImageExtractor<'a> {
...
                     // Found nested left-paren.
                     depth += 1;
-                } else if c == ')' {
+                } else if c == ")" {
                     if depth > 1 {
                         // Found nested right-paren.
                         depth -= 1;
                     } else {
                         // Found right-paren.
+                        println!("Character: {c} at {pos}");
```
ragusaa, line 104: Did you mean to remove this?

ragusaa (changes requested):

libclamav_rust/src/css_image_extract.rs
micahsnyder: Oh lol whoops. Will fix

ragusaa (approved):

PR #917, opened 2023-05-02
clamonacc - use modify and open exec marks

m-sola:


clamonacc/fanotif/fanotif.c
```diff
@@ -83,14 +83,14 @@ cl_error_t onas_setup_fanotif(struct onas_context **ctx)
     (*ctx)->fan_mask = fan_mask;
 
     if (optget((*ctx)->clamdopts, "OnAccessPrevention")->enabled && !optget((*ctx)->clamdopts, "OnAccessMountPath")->enabled) {
-        logg(LOGG_DEBUG, "ClamFanotif: kernel-level blocking feature enabled ... preventing malicious files access attempts\n");
-        (*ctx)->fan_mask |= FAN_ACCESS_PERM | FAN_OPEN_PERM;
+        logg(LOGG_DEBUG, "ClamFanotif: kernel-level blocking feature enabled ... preventing malicious file execution attempts\n");
```
micahsnyder, line 86: If we are scanning all file access, but only execution or modification are blocked with prevention, we may get a few confused users who see they are able to read files before the scan completes.  

I suggest putting some of this wording into the comments for `OnAccessPrevention` in both:
- `clamd.conf.sample`
- `clamd.conf.5.in`

micahsnyder:
Since this PR is specific to `OnAccessPrevention` and notable for blocking execution, I thought I'd test how that works. 

The NOTES in the manpage are particularly interesting https://man7.org/linux/man-pages/man2/fanotify_mark.2.html 

```
NOTES
   FAN_OPEN_EXEC and FAN_OPEN_EXEC_PERM
       When using either FAN_OPEN_EXEC or FAN_OPEN_EXEC_PERM within the
       mask, events of these types will be returned only when the direct
       execution of a program occurs.  More specifically, this means
       that events of these types will be generated for files that are
       opened using [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html), [execveat(2)](https://man7.org/linux/man-pages/man2/execveat.2.html), or [uselib(2)](https://man7.org/linux/man-pages/man2/uselib.2.html).  Events of
       these types will not be raised in the situation where an
       interpreter is passed (or reads) a file for interpretation.

       Additionally, if a mark has also been placed on the Linux dynamic
       linker, a user should also expect to receive an event for it when
       an ELF object has been successfully opened using [execve(2)](https://man7.org/linux/man-pages/man2/execve.2.html) or
       [execveat(2)](https://man7.org/linux/man-pages/man2/execveat.2.html).

       For example, if the following ELF binary were to be invoked and a
       FAN_OPEN_EXEC mark has been placed on /:

           $ /bin/echo foo

       The listening application in this case would receive
       FAN_OPEN_EXEC events for both the ELF binary and interpreter,
       respectively:

           /bin/echo
           /lib64/ld-linux-x86-64.so.2
```

So that had me wondering if it would work correctly for shell scripts and python scripts.

I set install clamav with this PR to `/usr/local` and set up this configuration:
```
❯ clamconf -n
Checking configuration files in /usr/local/etc

Config file: clamd.conf
-----------------------
LogFile = "/tmp/clamd.log"
LogClean = "yes"
LocalSocket = "/tmp/clamd.socket"
User = "clamav"
OnAccessIncludePath = "/home/micah/tmp"
OnAccessExcludeUname = "clamav"
OnAccessPrevention = "yes"

freshclam.conf not found

clamav-milter.conf not found

Software settings
-----------------
Version: 1.1.0
Optional features supported: MEMPOOL AUTOIT_EA06 BZIP2 LIBXML2 PCRE2 ICONV JSON RAR

Database information
--------------------
Database directory: /usr/local/share/clamav
daily.cvd: version 26897, sigs: 2033621, built on Fri May  5 00:23:19 2023
main.cvd: version 62, sigs: 6647427, built on Thu Sep 16 05:32:42 2021
bytecode.cvd: version 334, sigs: 91, built on Wed Feb 22 13:33:21 2023
Total number of signatures: 8681139

Platform information
--------------------
uname: Linux 5.15.90.1-microsoft-standard-WSL2 #1 SMP Fri Jan 27 02:56:13 UTC 2023 x86_64
OS: Linux, ARCH: x86_64, CPU: x86_64
Full OS version: Ubuntu 20.04.6 LTS
zlib version: 1.2.13 (1.2.13), compile flags: a9
platform id: 0x0a21b4b40800000000090400

Build information
-----------------
GNU C: 9.4.0 (9.4.0)
sizeof(void*) = 8
Engine flevel: 180, dconf: 180
```

As you can see, I enabled prevention and have it monitoring `/home/micah/tmp`.
Then I made this `test.py` file, set it to `+x`, and tried to run it
```
❯ cat tmp/test.py
#!/usr/bin/env python3

print("hello!~\n")

❯ cd tmp/
❯ ./test.py
```

The scan appears to complete, as I see this in the `clamd` log file:
```
/home/micah/tmp/test.py: OK
```
But running `./test.py` never completed.  No output, and it's still hung.  I killed `clamonacc` and then the script ran. 

I switched to the `main` branch and tried again and that didn't happen.  But it also scanned this many times:
```
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
/home/micah/tmp/test.py: OK
```
😅

micahsnyder (changes requested):
Not sure what the solution is, but something wrong with prevention and executing scripts. I could use help verifying my test results.

micahsnyder:
Closing merge request for now. 
In a team meeting a bit ago, @m-sola mentioned that she ran into the same issue with prevention in testing and didn't have a solution.

PR #927, opened 2023-05-25
oss-fuzz - fixing overread/write when unpacking pe files

m-sola:
Added a CLI_CONTAINS buffer wrap check to ensure we aren't reading and writing beyond the exe buffer

### DIFF
--- a/libclamav/wwunpack.c
+++ b/libclamav/wwunpack.c
@@ -231,19 +231,25 @@ cl_error_t wwunpack(uint8_t *exe, uint32_t exesz, uint8_t *wwsect, struct cli_ex
     }
 
     if (CL_SUCCESS == error) {
-        if (pe + 6 > exesz || pe + 7 > exesz || pe + 0x28 > exesz ||
-            pe + 0x50 > exesz || pe + 0x14 > exesz)
-            return CL_EFORMAT;
 
         exe[pe + 6] = (uint8_t)scount;
         exe[pe + 7] = (uint8_t)(scount >> 8);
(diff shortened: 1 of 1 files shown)


libclamav/wwunpack.c
```diff
@@ -237,11 +237,16 @@ cl_error_t wwunpack(uint8_t *exe, uint32_t exesz, uint8_t *wwsect, struct cli_ex
...
         exe[pe + 7] = (uint8_t)(scount >> 8);
+        
         if (!CLI_ISCONTAINED(wwsect, sects[scount].rsz, wwsect + 0x295, 4))
             cli_dbgmsg("WWPack: unpack memory address out of bounds.\n");
         else
             cli_writeint32(&exe[pe + 0x28], cli_readint32(wwsect + 0x295) + sects[scount].rva + 0x299);
-        cli_writeint32(&exe[pe + 0x50], cli_readint32(&exe[pe + 0x50]) - sects[scount].vsz);
+
+        if (!CLI_ISCONTAINED(exe, exesz, exe + pe + 0x50, 4))
+            cli_dbgmsg("WWPack: unpack memory address out of bounds.\n");
```
micahsnyder, line 247: I think this, and the previous one on line 242 should fail out with `CL_EFORMAT`.
There is also an earlier check on line 234 that we can remove because it tries to fix the same thing, but isn't quite right.

I attached a suggested patch to in the CLAM-2306 JIRA comments.

micahsnyder (changes requested):

micahsnyder (approved):
Update looks good to me.  And after finally resolving the jenkins `py.xml` issue, it passed nicely through the test pipelines.

PR #931, opened 2023-05-29
harmonize socket/pid file path to those used in Docker

computersalat:


micahsnyder (commented):
I would be happy to make a few of these adjustments manually before merging.

micahsnyder:
Rebased with main branch, and squashed those extra commits.

micahsnyder (approved):

Issue #923, opened 2023-05-16
Win.Downloader.LNKAgent-10001628-0 causing clamd to crash


LinuxMagicSupportTeam:
Describe the bug
----------------
The new siganture Win.Downloader.LNKAgent-10001628-0 is causing clamd to crash even a database test being passed by freshclam.

It's been broght up on other online forums as well.

How to reproduce the problem
----------------------------
Logs:
 freshclam[720]: Tue May 16 01:00:20 2023 -> Received signal: wake up
 freshclam[720]: Tue May 16 01:00:20 2023 -> ClamAV update process started at Tue May 16 01:00:20 2023
 freshclam[720]: Tue May 16 01:00:20 2023 -> daily database available for update (local version: 26907, remote version: 26908)
 freshclam[720]: Tue May 16 01:00:21 2023 -> Testing database: '/var/lib/clamav/tmp.05d7654ad9/clamav-815c289ec8e7bb456a352c9aadb35fcb.tmp-daily.cld' ...
 freshclam[1170615]: Tue May 16 01:00:24 2023 -> ~[LibClamAV] Don't know how to create filter for: Win.Downloader.LNKAgent-10001628-0
 freshclam[1170615]: Tue May 16 01:00:24 2023 -> ~[LibClamAV] cli_ac_addpatt: cannot use filter for trie
 freshclam[720]: Tue May 16 01:00:24 2023 -> Database test passed.
 freshclam[720]: Tue May 16 01:00:24 2023 -> daily.cld updated (version: 26908, sigs: 2034816, f-level: 90, builder: raynman)
 freshclam[720]: Tue May 16 01:00:25 2023 -> main.cvd database is up-to-date (version: 62, sigs: 6647427, f-level: 90, builder: sigmgr)
 freshclam[720]: Tue May 16 01:00:25 2023 -> bytecode.cvd database is up-to-date (version: 334, sigs: 91, f-level: 90, builder: anvilleg)
 freshclam[720]: Tue May 16 01:00:25 2023 -> Clamd successfully notified about the update.
 clamd[768]: Tue May 16 01:00:25 2023 -> Reading databases from /var/lib/clamav
 clamd[768]: LibClamAV Warning: Don't know how to create filter for: Win.Downloader.LNKAgent-10001628-0
 clamd[768]: LibClamAV Warning: cli_ac_addpatt: cannot use filter for trie  <---- clamd dies

micahsnyder:
[...]

PR #934, opened 2023-05-30
CLAM-2353 Abort signature load for short signature patterns

micahsnyder:
If a signature has a pattern that is too short will fail to load the siganture but does not cause the entire load process to abort. This is bad for two reasons:
1) It is not immediately apparent that the signature is bad, and so it could be published accidentally.
2) The signature is partially loaded by the time the bad pattern is observed and that may cause a crash later.

Because of (1), it is not worth it to try to unload the first part of the signature. Instead, we should just abort the signature load.

Fixes: https://github.com/Cisco-Talos/clamav/issues/923

### DIFF
--- a/libclamav/matcher-ac.c
+++ b/libclamav/matcher-ac.c
@@ -2951,6 +2951,7 @@ cl_error_t cli_ac_addsig(struct cli_matcher *root, const char *virname, const ch
             cli_warnmsg("cli_ac_addsig: cannot use filter for trie\n");
             MPOOL_FREE(root->mempool, root->filter);
             root->filter = NULL;
+            return CL_EMALFDB;
         }
 
         /* TODO: should this affect maxpatlen? */
--- a/libclamav/matcher-bm.c
+++ b/libclamav/matcher-bm.c
@@ -72,6 +72,7 @@ cl_error_t cli_bm_addpatt(struct cli_matcher *root, struct cli_bm_patt *pattern,
             cli_warnmsg("cli_bm_addpatt: cannot use filter for trie\n");
             MPOOL_FREE(root->mempool, root->filter);
             root->filter = NULL;
+            return CL_EMALFDB;
         }
         /* TODO: should this affect maxpatlen? */
     }
--- a/libclamav/readdb.c
+++ b/libclamav/readdb.c
@@ -4875,9 +4875,11 @@ cl_error_t cli_load(const char *filename, struct cl_engine *engine, unsigned int
     if (fs)
         fclose(fs);
 
-    if (engine->cb_sigload_progress) {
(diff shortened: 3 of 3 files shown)


m-sola (changes requested):
Testing current main showed:

```:~/clamav/test/bad_sig$ ../../bin/clamscan -d Win.Downloader.LNKAgent-10001628-0.ldb ./s00234-023-03120-1.pdf
LibClamAV Warning: Don't know how to create filter for: Win.Downloader.LNKAgent-10001628-0
LibClamAV Warning: cli_ac_addsig: cannot use filter for trie
Compiling:   0s, ETA:   0s [========================>]       40/40 tasks

AddressSanitizer:DEADLYSIGNAL
=================================================================
==18764==ERROR: AddressSanitizer: SEGV on unknown address 0x7fda0a808fff (pc 0x7fd910cad7ec bp 0x7fff45018270 sp 0x7fff45018200 T0)
==18764==The signal is caused by a READ memory access.```

After changes:

```../../bin/clamscan -d Win.Downloader.LNKAgent-10001628-0.ldb ./s00234-023-03120-1.pdf
LibClamAV Warning: Don't know how to create filter for: Win.Downloader.LNKAgent-10001628-0
LibClamAV Warning: cli_ac_addsig: cannot use filter for trie
LibClamAV Error: cli_add_content_match_pattern: Problem adding signature (1).
LibClamAV Error: cli_add_content_match_pattern: Problem adding signature (1b).
LibClamAV Error: cli_loadldb: failed to parse subsignature 2 in Win.Downloader.LNKAgent-10001628-0
LibClamAV Error: Problem parsing database at line 1
LibClamAV Error: Can't load Win.Downloader.LNKAgent-10001628-0.ldb: Malformed database
ERROR: Malformed database     [                         ]        0/3 sigs

----------- SCAN SUMMARY -----------
Known viruses: 0
Engine version: 1.2.0-devel-20230601```

Fix works great, no issues there.

And this is certainly an edge case, but does it make sense to *not* print the loading bar when we have a malformed db?

micahsnyder:
@m-sola good idea.  I just pushed up a commit that adds a check if success before trying to print that loading progress bar.  This way it doesn't print overtop the error message.

While looking at that I also found that the same logic of adding a filter (and failing on a pattern < 2 bytes) exists for the Boyer-Moore matcher as well.   I don't have a test case to demonstrate it, but it definitely needs to fail out if a signature fails to load.  So I added a `return CL_EMALFDB;` there as well.

m-sola (approved):
Updated printout looks great!

micahsnyder:
Marking as a cherry-pick candidate because while this isn't super critical, it's safest to backport the fix for 1.0 LTS.

Issue #809, opened 2023-01-12
Raise artificial 4GB limit for MaxScanSize to allow scanning larger archives


scddev:
As comment in clamd.conf suggests:

```
# Value of 0 disables the limit
```

The result is a limit of 4GB:
```
Thu Jan 12 07:12:28 2023 -> Limits: Global size limit set to 4294967295 bytes.
Thu Jan 12 07:12:28 2023 -> Limits: File size limit set to 4294967295 bytes.
```

This is caused by a change dating back to 2011: https://github.com/Cisco-Talos/clamav/commit/6f4847aacfd153f0a88612789cac0deb233ee669

When reverting this change, also limits can be disabled again:
```
WARNING: Thu Jan 12 07:09:58 2023 -> Limits: Global size limit protection disabled.
WARNING: Thu Jan 12 07:09:58 2023 -> Limits: File size limit protection disabled.
```

micahsnyder:
@scddev You cannot disable the file size or scan size limit. That is expected behavior.  But by settings to `0` which maxes out the limit. That's as close as we can get to disabling it. 

Further, the real internal limit for MaxFileSize, specifically, is 2GB and not the 4GB configuration limit.
See: https://github.com/Cisco-Talos/clamav/issues/344
And: https://github.com/Cisco-Talos/clamav/blob/main/etc/clamd.conf.sample#L536

But perhaps we should modify the conf wording to say something like this instead:
```
# This option sets the maximum amount of data to be scanned for each input
# file. Archives and other containers are recursively extracted and scanned
# up to this value.
# Value of 0 will set the limit to 4GB which is the maximum option.
# Note: disabling this limit or setting it too high may result in severe damage
# to the system.
# Default: 400M
#MaxScanSize 1000M

# Files larger than this limit won't be scanned. Affects the input file itself
# as well as files contained inside it (when the input file is an archive, a
# document or some other kind of container).
# Value of 0 will set the limit to 2GB which is the maximum option.
# Note: disabling this limit or setting it too high may result in severe damage
# to the system.
# Technical design limitations prevent ClamAV from scanning files greater than
# 2 GB at this time.
# Default: 100M
#MaxFileSize 400M
```

scddev:
@micahsnyder I am fine with the technical limit for MaxFileSize and adapting the documentation, my main concern is in regards of the MaxScanSize which does not relate to any technical issue and just prevents a scan of larger files (archives).

[...]

PR #945, opened 2023-06-07
Raise artificial 4G limit for MaxScanSize

matthias-fratz-bsz:
Allows MaxScanSize (only) to be set higher than 4GB on 32-bit systems or using suffixes. (On 64-bit it can already be set beyond 32-bit limits by not using suffixes.) Intended as a fix for #809 

Introduces a new Suffix "G" to specify gibibytes for all size options. This isn't terribly useful because there only 1G, 2G and 3G really work, and even "4G" triggers the "too high" message... but it is more consistent.

This code makes 3 assumptions:
- It is ok to use C99 functions (`strtoll`) and featuers (`long long`)
- We don't want to break the ability to set the other limits higher than 4G on 64-bit systems
- Limit values are stored in `unsigned long int` downstream, so they can actually be 64-bit on 64-bit systems.
That last one may not be true, but if it isn't fulfilled then we can already cause breakage with something like `MaxFileSize=10000000000` (nominally 10GB).

Not sure how to test this. It builds on AMD64, but I don't have equipment to test this on i386 or more obscure architectures (esp. ILP64 ones)...

### DIFF
--- a/libclamav/wwunpack.c
+++ b/libclamav/wwunpack.c
@@ -232,19 +232,23 @@ cl_error_t wwunpack(uint8_t *exe, uint32_t exesz, uint8_t *wwsect, struct cli_ex
 
     if (CL_SUCCESS == error) {
 
-        exe[pe + 6] = (uint8_t)scount;
-        exe[pe + 7] = (uint8_t)(scount >> 8);
-        
-        if (!CLI_ISCONTAINED(wwsect, sects[scount].rsz, wwsect + 0x295, 4)) {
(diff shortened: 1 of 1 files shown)


ragusaa (approved):

micahsnyder:
What.... I didn't close it..

micahsnyder:
Well I can fix this... but apparently not by re-opening this PR. As I seem to have accidentally deleted the commits from @matthias-fratz-bsz's branch.

Here's a replacement PR with the same commit: https://github.com/Cisco-Talos/clamav/pull/986

PR #949, opened 2023-06-10
Prevent -f file disclosure

0xdreadnaught:
If a user is give sudo access to clamscan, they can point `--file-list` at a file owned by the sudo user, usually root, and they can parse the errors to read the file. for example `sudo clamscan -f /etc/shadow`. This can be easily resolved by adding line number tracking with a corresponding update to the error output.

Changing the error output to a line number will clearly indicate where the problem is to a legitimate user while giving nothing to an attacker. PSRT decided this wasn't something they wanted to patch, so here's a patch.

I don't believe this will interfere with other parts of ClamAV.

Before: 
image
After: 
image

### DIFF
--- a/clamscan/manager.c
+++ b/clamscan/manager.c
@@ -978,8 +978,12 @@ static int scan_files(struct cl_engine *engine, const struct optstruct *opts, st
             ret = scanstdin(engine, options);
         } else if (LSTAT(file, &sb) == -1) {
             /* Can't access the file */
-            perror(file);
-            logg(LOGG_WARNING, "%s: Can't access file\n", file);
+            perror("Error");  // change the argument to a constant string
+            if(optget(opts, "file-list")->enabled){
+                logg(LOGG_WARNING, "ERROR: Unable to access file on line %d\n", get_filelist_line_number());
+            } else {
+                logg(LOGG_WARNING, "ERROR: Unable to access a file\n");
+            }
             ret = 2;
         } else {
             /* Can access the file. Now have to identify what type of file it is */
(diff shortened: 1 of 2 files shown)


micahsnyder:
> `--leave-temps` doesn't leave files that the malicious user can access. They are all root owned. image

That's a good point. 

> While they do disclose small amounts of information, mainly filenames, they are nothing in comparison to full on file reads. I do agree this is the tip of the iceberg, but it's also the biggest piece of it.

True it will be a subset for whatever is next.  But to be clear I don't think anyone should be running `clamscan` or `clamd` as root.  If another remote code execution vuln comes along, that means the attacker gets to run as root.  I would much rather say "Don't do this!" - than go down a rabbit-hole trying to prevent disclosing sensitive parts of documents, archives, and other files. 

> As for the user experience, I'm not sure if it would be too much overhead but we could add an `fopen` check before printing the error. If it's a valid path then it discloses the path in the error. If it's not a valid path, then it prints the line number.

This would be better, but again... not sure how much I care if I'm recommending against scanning with sudo.  

Pretend `clamscan` is a really lousy document reader.  Would you give users `sudo` privileges to run it on other peoples files?

0xdreadnaught:
You clearly already have your mind set so feel free to close the PR. Your official documents cover sudo setup. They even specifically state the reason behind why it is "not recommended". But it would be nice for Clam to update their docs to "not supported" so it's clear what your level of support really is.

Personally I think it's foolish/lazy to reject a patch because it requires poor configuration.

micahsnyder:
I was on PTO last week, hence this delay.  I waited to think about it and let you think about it before I reply.  After consideration I don't have anything else to add.  

It's not just a matter of not supporting running as `sudo`.  Like, giving the scanner root level privileges is more dangerous and what we wish to discourage.   But your specific use case is more unique.  

ClamAV is not intended to keep the files contents confidential from the user/account running ClamAV.  In fact, it's quite the opposite.  We print out all sorts of stuff through the debug log to help the user understand more about the structure and contents of the file and intend to add more of that kind of thing.  Using `clamscan` for file analysis, even without the temp files,  is critical to developing signatures, or debugging issues with the scanning process.  

So yeah my mind is pretty set on rejecting this kind of change.  If anything, I want to see more of the opposite where it prints more about file contents.

PR #952, opened 2023-06-13
Abort signature load for short signature patterns

micahsnyder:
Backport of https://github.com/Cisco-Talos/clamav/pull/934

---

If a signature has a pattern that is too short will fail to load the signature but does not cause the entire load process to abort. This is bad for two reasons:
1) It is not immediately apparent that the signature is bad, and so it could be published accidentally.
2) The signature is partially loaded by the time the bad pattern is observed and that may cause a crash later.

Because of (1), it is not worth it to try to unload the first part of the signature. Instead, we should just abort the signature load.

Fixes: https://github.com/Cisco-Talos/clamav/issues/923

We should also abort loading if the filter pattern for the boyer-moore matcher is shorter than 2 bytes.

Also, do not print the final "Loading" progress bar if an error occurred.


micahsnyder:
Internal testing went well. Will merge.

PR #955, opened 2023-06-15
Allow to use client authentication with certificate in freshclam

jedrzejj:
This PR adds the possibility to define certificate and private key for client authentication in freshclam when using private repo. Certificate and key paths are read from environment variables and passed on to libcurl, similarly to the CA bundle.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,6 +9,18 @@ ClamAV 1.2.0 includes the following improvements and changes:
 
 ### Major changes
 
+- Added ability for Freshclam to use a client certificate PEM file and a
+  private key PEM file for authentication to a private mirror by setting the
+  following environment variables:
+  - `FRESHCLAM_CLIENT_CERT`: May be set to the path of a file (PEM) containing
+    the client certificate.
+  - `FRESHCLAM_CLIENT_KEY`: May be set to the path of a file (PEM) containing
+    the client private key.
+  - `FRESHCLAM_CLIENT_KEY_PASSWD`: May be set to a password for the client key
+    PEM file, if it is password protected.
+
+  Patch courtesy of jedrzej.
+
 ## Other improvements
 
 ### Bug fixes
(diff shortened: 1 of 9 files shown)


micahsnyder:
Apologies for the delay in reviewing this.  This change look very straightforward. 

Some issues that I've taken on with the commit I just pushed:
- I found some missing documentation.  
- I found no reason why this feature couldn't work on mac and windows.
- I renamed the variables to prefix with `FRESHCLAM_` instead of `CURL_` because they do not piggyback on existing `curl` environment variables and are unique to clamav.  I specifically I chose `FRESHCLAM_` over `CLAM_` or `CLAMAV_` because it may make sense to someday support certificate-based authentication between `clamd` and other scan clients -- and I figure I don't want confusion with the names in that hypothetical future. 
- I was looking at the libcurl documentation and noticed that it supports password protected key files, so I added one more variable to support that.

I believe this works, and could be tested like so, if your `freshclam.conf` file is set up to talk to a private mirror that enforces client certificate authentication...
Linux/Unix:
```bash
FRESHCLAM_CLIENT_CERT="/home/micah/client_certificate.pem" FRESHCLAM_CLIENT_KEY="/home/micah/client_private_key.pem"  FRESHCLAM_CLIENT_KEY_PASSWD="testtest" ./install/bin/freshclam
```
Windows powershell:
```powershell
$env:FRESHCLAM_CLIENT_CERT="C:\Users\micah\.ssh\client_certificate.pem" & $env:FRESHCLAM_CLIENT_KEY="C:\Users\micah\.ssh\client_private_key.pem" & .\install\freshclam.exe
```

I don't have such a server and am uncertain how to quickly set one up.

micahsnyder (approved):
Had a bit of a hiccup this morning with our Jenkins. It's fixed now and this ran through okay, not that I expected any differently. Perhaps we could set up a test some time that makes use of this feature, but in the meantime I'm going to trust that it works okay base don the limited testing I did, and happy that it isn't causing any problems.

Issue #564, opened 2022-04-29
ClamAV is unusable on FIPS-enabled Linux systems due to MD5 use


qralston:
Describe the bug
----------------

ClamAV is completely unusable on certain Linux distributions when FIPS mode is enabled.

The Linux system affected include at least RHEL8 and RHEL9 (beta).

How to reproduce the problem
----------------------------

Build and install ClamAV on a RHEL8 or RHEL9 beta system without FIPS mode enabled. Observe that all ClamAV functionality works.

Reboot the host into FIPS mode (by passing `fips=1` to the kernel).

Observe that any attempt to load any of the databases fails:

```
$ clamscan /usr/share/dict/words
LibClamAV Error: Can't load /var/lib/clamav/main.cvd: Can't allocate memory
LibClamAV Error: cli_loaddbdir(): error loading database /var/lib/clamav/main.cvd
ERROR: Can't allocate memory
…
```

In this case, the _Can't allocate memory_ error is somewhat of a red herring. The true issue is that when FIPS mode is active, non–FIPS-approved hashing algorithms are disabled, and that includes MD5, which ClamAV uses extensively internally.

This was a known issue, and was already reported back on 2019-10-29 in [BZ#12424](https://bugzilla.clamav.net/show_bug.cgi?id=12424):

> When FIPS mode is active -- it disables features not approved for cryptographic use.  This includes use of the MD5 hashing algorithm, which ClamAV uses for both verdict caching in clamd as well as for many, if not most, hash-based signatures.
> 
> You're observing use of MD5 failing when freshclam test-loads a database containing MD5 signatures.  I suspect if you downloaded the databases manually and ran clamd, it would also fail pretty fast when it collects an MD5 hash to use in caching, whitelisting, or for comparison with MD5 signatures. 
> 
> We have a branch where we've experimented with switching to use SHA256 hashes. It it is a rather large change and requires more investigation before I can really make any promises about a specific direction or timeline for this effort.

This bug was not carried over to GitHub, so this issue is a continuation of the Bugzilla bug.

The bottom line: in 2022, no software should be using anything other than the SHA2 (SHA-256, SHA-384, SHA-512) and SHA3 family of hashing algorithms. Any use of MD5 and/or SHA-1 should be replaced with SHA2 or later, as both MD5 and SHA-1 are considered by cryptographers to be broken.

An increasing number of security hardening standards (e.g., DISA STIGs) require FIPS mode, so this will only affect more ClamAV users over time.

[...]

PR #959, opened 2023-06-19
fix:https://github.com/Cisco-Talos/clamav/issues/564

barrydegraaff:
Allow to use ClamAV on FIPS enabled OpenSSL

### DIFF
--- a/libclamav/crypto.c
+++ b/libclamav/crypto.c
@@ -170,7 +170,7 @@ unsigned char *cl_hash_data(const char *alg, const void *buf, size_t len, unsign
     size_t cur;
     int winres = 0;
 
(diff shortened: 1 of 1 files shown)


qralston:
This merge request should be rejected; see [my comment on issue 564](https://github.com/Cisco-Talos/clamav/issues/564#issuecomment-1600003995).

micahsnyder:
This change would make it "work" but is a cheat.  Sorry @barrydegraaff

PaulSD:
In addition to the issues discussed in #564, also note that this PR introduces a memory leak, since the return value of `EVP_get_digestbyname()` does not need to be freed, but the return value of `EVP_MD_fetch()` does need to be freed.

PR #961, opened 2023-06-20
Fix destination directory for freshclam databases

jedrzejj:
Fix a bug in `freshclam` DB installation, if DESTDIR is defined on `make install` invocation

### DIFF
--- a/freshclam/CMakeLists.txt
+++ b/freshclam/CMakeLists.txt
@@ -47,7 +47,7 @@ else()
 endif()
 
 # Install an empty database directory
-INSTALL(CODE "FILE(MAKE_DIRECTORY \${ENV}\${CMAKE_INSTALL_PREFIX}/\${DATABASE_DIRECTORY})" COMPONENT programs)
+INSTALL(CODE "FILE(MAKE_DIRECTORY \"\$ENV{DESTDIR}\${CMAKE_INSTALL_PREFIX}/\${DATABASE_DIRECTORY}\")" COMPONENT programs)
 
 # Now we rename freshclam-bin executable to freshclam using target properties
 set_target_properties( freshclam-bin


micahsnyder (changes requested):
I agree that this is not working correctly.  In fact, I don't think it's working at all, regardless of whether `make DESTDIR=some/path install` is used, or not.

However, I don't think this solution is correct either. 

Ideally what we want is for it to make the empty database directory in one of several places...

If `DATABASE_DIRECTORY` is a relative path (doesn't start with `/` or `C:\`) then install to: 
`${CMAKE_INSTALL_PREFIX}/${DATABASE_DIRECTORY}`

If  `DATABASE_DIRECTORY` is an absolute path then install to: 
`${DATABASE_DIRECTORY}`

The two most common use cases on Linux/Unix are:
- If installing to CMAKE_INSTALL_PREFIX `/usr/local` and DATABASE_DIRECTORY is the default `share/clamav`, then it would install to `/usr/local/share/clamav`. 
- If installing to CMAKE_INSTALL_PREFIX `/usr` and DATABASE_DIRECTORY is the default `/var/lib/clamav`, then it would install to `/var/lib/clamav`. 

If installing with `make DESTDIR=some/path install`, then the same as above, except it should be prefixed with DESTDIR, like:
`${PREFIX}/${CMAKE_INSTALL_PREFIX}/${DATABASE_DIRECTORY}`
or:
`${PREFIX}/${DATABASE_DIRECTORY}`

So those two uses cases when using `make DESTDIR=/build` would end up like this:
`/build/usr/local/share/clamav`
or and this:
 `/build/var/lib/clamav`

I just tried testing with the `main` branch and with this PR and it didn't seem to make the empty database directory at install time in either case, regardless if using `DESTDIR` or not.  So it seems entirely broken to me.  Maybe I'm doing something wrong?

micahsnyder:
Closing because PR didn't work as intended, and is inactive.

PR #962, opened 2023-06-23
Systemd timer for freshclam updates

nils-werner:
This PR implements a systemd unit that **updates freshclam once**, without sending freshclam into the background.

It comes with a systemd timer which, when activated, will call this unit periodically. This takes the "burden of timing the updates" off of freshclam and puts it onto systemd. The timer can then easily activated, audited and the logs inspected:

```
sudo systemctl enable --now clamav-freshclam-once.timer
sudo systemctl list-timers
sudo systemctl status clamav-freshclam-once.timer
sudo systemctl status clamav-freshclam-once.service
journalctl -u clamav-freshclam-once.service
```

And if you want a different update interval you can edit the timer unit file using

```
sudo systemctl edit clamav-freshclam-once.timer
```

### DIFF
--- a/freshclam/CMakeLists.txt
+++ b/freshclam/CMakeLists.txt
@@ -61,4 +61,18 @@ if(SYSTEMD_FOUND)
         FILES ${CMAKE_CURRENT_BINARY_DIR}/clamav-freshclam.service
         DESTINATION ${SYSTEMD_UNIT_DIR}
         COMPONENT programs)
+    configure_file(
+        ${CMAKE_CURRENT_SOURCE_DIR}/clamav-freshclam-once.service.in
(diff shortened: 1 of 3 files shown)


micahsnyder (approved):
This change makes sense to me. And the code looks good as far as I can tell. 

I tested on an ubuntu 22.04 VM with 
```
systemctl enable clamav-freshclam-once.timer
systemctl start clamav-freshclam-once
systemctl status clamav-freshclam-once
```

It seemed to work okay. 

Would you be interested in helping write some documentation for setting enabling the systemd freshclam, clamd, and maybe clamonacc services?  
The documentation page would be here https://github.com/Cisco-Talos/clamav-documentation/blob/main/src/manual/Usage/Services.md 

This source backs https://docs.clamav.net/manual/Usage/Services.html

A related PR: https://github.com/Cisco-Talos/clamav/pull/859

PR #968, opened 2023-06-28
Export symbols for calculating image fuzzy hash

micahsnyder:
These symbols are used by an internal python tool for generating signatures:
- fuzzy_hash_calculate_image
- ffierror_fmt

`ffierror_fmt` is required to free the error structure passed back in case of an error.

Since version 1.1.0 started using libclamav.map again, we need to explicitly export these symbols.


micahsnyder:
Decided against backporting. Doesn't make sense for 1.0, since the libclamav.map file isn't used. And 1.1 isn't LTS so probs not worth it.  We can just use 1.2 for the internal python tool.

PR #970, opened 2023-07-07
CLAM-2298 CMake: link openssl 3.x with clamsubmit on macOS

micahsnyder:
When switching to openssl 3.x, linking with clamsubmit fails with undefined openssl symbols. The error message from Xcode is crazy obtuse:

ld: initializer '_OPENSSL_cpuid_setup' is >4GB from start of image in 'anon' from /Users/clamav_jenkins_svc/clamav-mussels-cookbook/test/install-x86_64/lib/libcrypto.a(libcrypto-lib-x86_64cpuid.o)
clang: error: linker command failed with exit code 1 (use -v to see invocation)

Anyhow... It seems that we must explicitly link clamsubmit with openssl now in order for this to work.

### DIFF
--- a/clamsubmit/CMakeLists.txt
+++ b/clamsubmit/CMakeLists.txt
@@ -34,6 +34,8 @@ target_link_libraries( clamsubmit
     PRIVATE
         ClamAV::libclamav
         ClamAV::common
+        OpenSSL::SSL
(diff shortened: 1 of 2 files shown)


rsundriyal:
Looks good

PR #977, opened 2023-07-20
Clam-2379 Correction to PE wwunpack overread guard

micahsnyder:
The prior fix for the wwunpack overread in commit
89cd0df3d7d4c2163d79056b5c0ec55c4992c8f4 was a little too late, but also removed an earlier, smaller guard for a write.

This commit just moves the larger guard a little earlier to protect against both.

Resolves: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60655

Note: This bug never made it into a release, so not a vuln.


ragusaa:
Code looks good, verifying the test case and will approve.

ragusaa (approved):

PR #981, opened 2023-07-27
Clam-2380: OLE2: Fix bounds check on OLE2 encryption info check

micahsnyder:
The checks for the encryption info cspName and encryption verifier don't have the size of the overall file available for the check and may overflow.

This commit passes in the size of the file to the `initialize_encryption_key()` function and does all size checks within that function instead of doing the overall size check before that function.

Resolves: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=60563

We determined that this possible overflow read is not a vulnerability.

### DIFF
--- a/libclamav/ole2_extract.c
+++ b/libclamav/ole2_extract.c
@@ -84,7 +84,7 @@ typedef struct ole2_header_tag {
     unsigned char clsid[16];
     uint16_t minor_version __attribute__((packed));
     uint16_t dll_version __attribute__((packed));
-    int16_t byte_order __attribute__((packed)); /* -2=intel */
+    int16_t byte_order __attribute__((packed));             /* -2=intel */
 
     uint16_t log2_big_block_size __attribute__((packed));   /* usually 9 (2^9 = 512) */
     uint32_t log2_small_block_size __attribute__((packed)); /* usually 6 (2^6 = 64) */
@@ -134,7 +134,7 @@ typedef struct ole2_header_tag {
  * https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/60fe8611-66c3-496b-b70d-a504c94c9ace
  */
 typedef struct property_tag {
-    char name[64]; /* in unicode */
+    char name[64];       /* in unicode */
     uint16_t name_size __attribute__((packed));
     unsigned char type;  /* 1=dir 2=file 5=root */
     unsigned char color; /* black or red */
@@ -2089,7 +2089,7 @@ static cl_error_t generate_key_aes(const char *const password, encryption_key_t 
 
     tmp = verifier->salt_size;
     if (verifier->salt_size > sizeof(verifier->salt)) {
-        cli_warnmsg("ole2: Invalid salt length '0x%x'\n", verifier->salt_size);
+        cli_dbgmsg("ole2: Invalid salt length '0x%x'\n", verifier->salt_size);
         tmp = sizeof(verifier->salt);
     }
     memcpy(buffer, verifier->salt, tmp);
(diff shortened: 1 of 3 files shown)


micahsnyder:
Good idea.  You're right it'll just annoy users.  I can switch them to Debug.

ragusaa:
Was just thinking if there was a problem parsing, rather than a specific warning about the salt, to just have something like 
LibClamAV Debug: ole2: Invalid salt length '0x20202020'
LibClamAV Debug: ole2: Invalid encrypted verifier hash length 0x20202020
LibClamAV Warning: Unable to parse the encryption verifier.

libclamav/ole2_extract.c
```diff
@@ -2350,68 +2358,77 @@ static bool initialize_encryption_key(const encryption_info_stream_standard_t *h
...
      * Find the end, and we'll have the index of the EncryptionVerifier.
      * The CPSName string *should* always be either
      * 'Microsoft Enhanced RSA and AES Cryptographic Provider'
      * or
      * 'Microsoft Enhanced RSA and AES Cryptographic Provider (Prototype)'
-     *
      */
-    for (idx = 0; idx < CSP_NAME_LENGTH(headerPtr) - 1; idx += 2) {
-        if (((uint16_t *)&(headerPtr->encryptionInfo.cspName[idx]))[0] == 0) {
+    encryptionInfo_CSPName = (uint16_t *)(encryptionInfoStreamPtr + sizeof(encryption_info_t));
```
ragusaa, line 2397: I believe this line should be 
encryptionInfo_CSPName = (uint16_t *)(encryptionInfoStreamPtr + sizeof(encryption_info_stream_standard_t));

This accounts for the version_major, version_minor, flags, and size at the beginning of the encryption_info_stream_standard_t structure.

libclamav/ole2_extract.c
```diff
@@ -2350,68 +2358,77 @@ static bool initialize_encryption_key(const encryption_info_stream_standard_t *h
...
+    for (idx = 0; idx * sizeof(uint16_t) < remainingBytes; idx++) {
+        if (encryptionInfo_CSPName[idx] == 0) {
             break;
         }
     }
 
-    idx += 2;
-    if ((sizeof(headerPtr->encryptionInfo.cspName) - idx) <= sizeof(encryption_verifier_t)) {
+    encryptionVerifierPtr = (uint8_t*)encryptionInfo_CSPName + (idx + 1) * sizeof(uint16_t);
+    remainingBytes -= (idx + sizeof(uint16_t));
```
ragusaa, line 2412: Should this be  
'remainingBytes -= (idx * sizeof(uint16_t);'

ragusaa (changes requested):
See code comments in PR

libclamav/ole2_extract.c
micahsnyder: OMG. Yes.

micahsnyder: Fixed

micahsnyder:
> Was just thinking if there was a problem parsing, rather than a specific warning about the salt, to just have something like
> LibClamAV Debug: ole2: Invalid salt length '0x20202020' 
> LibClamAV Debug: ole2: Invalid encrypted verifier hash length 0x20202020 
> LibClamAV Warning: Unable to parse the encryption verifier.

I think people may still complain about the warning with no idea what to do if it turns out to be common. I think I'd rather leave it off.

libclamav/ole2_extract.c
micahsnyder: Oh crap, you're right. Fixed.

ragusaa (approved):

PR #986, opened 2023-08-03
Raise artificial 4G limit for MaxScanSize

micahsnyder:
This limit is internally "long long", so >=64-bit even on 32-bit platforms. Also fixes a related issue where limits could have been set to negative values on 64-bit platforms where setting a "long long" (64-bit signed) can overflow if assigned from an "unsigned long" (64-bit unsigned).

Resolves: https://github.com/Cisco-Talos/clamav/issues/809

### DIFF
--- a/common/optparser.c
+++ b/common/optparser.c
@@ -63,7 +63,7 @@
 #define MAXCMDOPTS 150
 
 #define MATCH_NUMBER "^[0-9]+$"
-#define MATCH_SIZE   "^[0-9]+[KM]?$"
+#define MATCH_SIZE   "^[0-9]+[KMG]?$"
 #define MATCH_BOOL   "^(yes|true|1|no|false|0)$"
 
 #define FLAG_MULTIPLE 1 /* option can be used multiple times */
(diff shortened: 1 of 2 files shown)


micahsnyder:
@ragusaa approved https://github.com/Cisco-Talos/clamav/pull/945 which *was* the same as this one, except this is rebased to have the latest commits from main.  

I want to see this run through Jenkins before I merge it -- as the tests never ran on the previous PR.

micahsnyder:
Some internal tests that expected specifying LogFileMaxSize to fail when using `1g` or `1G` now pass with this change (yay!).  I'll update the tests and re-run this through Jenkins again.

micahsnyder:
Thanks so much for the contribution @matthias-fratz-bsz

micahsnyder:
@McPatate  file-size is different from scan-size.  file-size is still capped at ~2GB, and is the reason you're seeing that message.  This PR only raised the limit for scan-size.

McPatate:
@micahsnyder, just to clarify, does this mean that the maximum file size ClamAV can scan is 2GB - 1? In other words, ClamAV will only scan larger files if they are archives that, when uncompressed, exceed the 2GB limit and are thus subject to the MaxScanSize constraint. Is my understanding correct?

If so, are there any plans to raise the limit here as well?

McPatate:
Understood, thank you for taking the time to detail what is going on!

Issue #990, opened 2023-08-08
Unresolved external symbol(s) when trying to link clamav applications


driverxdw:
Describe the bug
----------------
environment:
host: `x64-win10`
compiler: `msvc(vs2022)`
clamav: `v1.1.0`

I used `vcpkg + cmake` to compile clamav on windows according to the clamav documentation, but finally reported an error when linking each application clamav application like this:
```
PS C:\Users\xdw\Desktop\clamav-1.1.0\build> cmake --build . -j16
MSBuild version 17.6.3+07e294721 for .NET Framework

  bytecode_runtime.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclamav\bytecode_runtime.dir\Debug\bytecode_runtime.lib
  regex.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclamav\regex.dir\Debug\regex.lib
  win32_compat_obj.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\win32\compat\win32_compat_obj.dir\Debug\win32_compat_obj.lib
  yara.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclamav\yara.dir\Debug\yara.lib
  lzma_sdk.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclamav\lzma_sdk.dir\Debug\lzma_sdk.lib
  win32_compat.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\win32\compat\Debug\libwin32_compat.lib
  Auto build dll exports
  Auto build dll exports
  clammspack.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclammspack\Debug\libclammspack.dll
  clamunrar.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclamunrar\Debug\libclamunrar.dll
  Auto build dll exports
  clamunrar_iface.vcxproj -> C:\Users\xdw\Desktop\clamav-1.1.0\build\libclamunrar_iface\Debug\libclamunrar_iface.dll
  Auto build dll exports
     Creating library C:/Users/xdw/Desktop/clamav-1.1.0/build/libclamav/Debug/clamav.lib and object C:/Users/xdw/Desktop/clamav-1.1.0/build/libclamav/Debug/clamav.exp
[...]

PR #994, opened 2023-08-09
CMake-Rust [Windows]: Fix issue detecting native static libs 1.1.1

micahsnyder:
Cherry-pick of https://github.com/Cisco-Talos/clamav/pull/992

---

Some change in Cargo/Rust version 1.70 or 1.71 appears to have broken the build on Windows because we are incorrectly attempting to check the native static libraries by compiling an empty file (/dev/null) which does not exist on Windows.

A simple fix is to make an empty file of our own and use that instead.

Fixes: https://github.com/Cisco-Talos/clamav/issues/990


micahsnyder:
Test pipeline is in progress but got past the important parts. So backport here looks good.

PR #997, opened 2023-08-11
Fix build issue in oss-fuzz environment (1.1.1)

micahsnyder:
There is an error building clam with the rust nightly compiler:

error[E0635]: unknown feature `proc_macro_span_shrink`
  --> /rust/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.56/src/lib.rs:92:30
   |
92 |     feature(proc_macro_span, proc_macro_span_shrink)
   |                              ^^^^^^^^^^^^^^^^^^^^^^

We need this to work because oss-fuzz is apparently using rust nightly in their image.

Bumping the proc-macro2 crate version in Cargo.lock resolves the error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -823,9 +823,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.56"
+version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


micahsnyder:
fuzz-regression pipeline looking good again.

PR #1001, opened 2023-08-17
Update scan limits documented in clamscan and clamav-milter manpages

micahsnyder:
Commit d1656ee increased the default scan limits. We updated them in the clamd manpages but forgot to update them in the clamscan and clamav-milter manpages.

Note that I'm also removing the "max: <4 GB" annotations because of the change in commit 2962509 that removed the artificial 4 GB limit on sizes that most notably increases the max-scansize to make it easy to scan very large archives comprised of more than 4 GB worth of stuff.

There is no specific limit now on these size options.


rsundriyal (approved):
Looks good.

PR #1002, opened 2023-08-18
Windows: json-c 0.17 compatibility with ssize_t type definition

micahsnyder:
json-c 0.17 defines the ssize_t type using a typedef on Windows. We have been setting ssize_t for Windows to a different type using a #define instead of a typedef.

We should have been using a typedef since it is a type. However, we must also match the exact type they're setting it to or else the compiler will baulk because the types are different.

Note: in C11, it's fine to use typedef the same type more than once, so long as you're defining it the same every time.


micahsnyder:
Verbally approved by a teammate that trusts me too much.

Also, I'm happy with the test results.

micahsnyder:
Probably should have backported this for 1.1 and 1.0. I'll throw together the PR's real quick. Hopefully no conflict. 

I'm unsure if the same issue exists with 0.103 with the Visual Studio build.  It ... probably does.  I'll investigate.

PR #1004, opened 2023-08-18
Clam 2226 zip filenames

micahsnyder:
- ZIP: Always parse file names

   Having the filename is useful for certain callbacks, and will likely be
   more useful in the future if we can start comparing detected filetypes
   with file extensions.

   E.g. if filetype is just "binary" or "text" we may be able to do better
   by trusting a ".js" extension to determine the type.
   Or else if detected file type is "pe" but the extension is ".png" we may
   want to say it's suspicious.

- Adjust example callback program to disable metadata option

   The CL_SCAN_GENERAL_COLLECT_METADATA is no longer required for the Zip
   parser to record filenames for embedded files, and described in the
   previous commit.

   This program can be used to demonstrate that it is behaving as desired.

### DIFF
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -624,8 +624,8 @@ static unsigned int parse_local_file_header(
     uint32_t csize, usize;
     unsigned int size_of_fileheader_and_data = 0;
 
-    uint32_t nsize;
-    const char *src;
+    uint32_t nsize  = 0;
+    const char *src = NULL;
 
     if (!(local_header = fmap_need_off(map, loff, SIZEOF_LOCAL_HEADER))) {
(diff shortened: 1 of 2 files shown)


libclamav/unzip.c
```diff
@@ -624,6 +624,9 @@ static unsigned int parse_local_file_header(
     uint32_t csize, usize;
     unsigned int size_of_fileheader_and_data = 0;
 
+    uint32_t nsize;
```
shutton, line 627: Would suggest initializing these to 0/NULL for a little future-proofing.

shutton (approved):

libclamav/unzip.c
micahsnyder: Fixed

PR #1008, opened 2023-08-22
Update to UnRAR v6.2.10

micahsnyder:


libclamunrar/extract.cpp
```diff
@@ -646,9 +646,9 @@ bool CmdExtract::ExtractCurrentFile(Archive &Arc,size_t HeaderSize,bool &Repeat)
       wcsncpyz(DestFileName,Cmd->DllDestName,ASIZE(DestFileName));
 #endif
 
-    if (ExtrFile && Command!='P' && !Cmd->Test && !Cmd->AbsoluteLinks &&
-        ConvertSymlinkPaths)
-      ExtrFile=LinksToDirs(DestFileName,Cmd->ExtrPath,LastCheckedSymlink);
+    // if (ExtrFile && Command!='P' && !Cmd->Test && !Cmd->AbsoluteLinks &&
```
shutton, line 649: Someone once told me, "don't commit commented-out code -- that's what SCM is for"

shutton (approved):
Thank goodness for "hide whitespace"

Seems fine. Probably no need to keep the old code around. If you do, I'd add a comment about what it *used* to do, and why you took it out.

libclamunrar/extract.cpp
micahsnyder: Haha that's a good point.  I say the same thing sometimes. Okay, will just delete it.

micahsnyder: Fixed.

micahsnyder:
I'm still trying to figure out why GitHub's Mac pipeline is failing to use C++11 (and failing the the build).  
Works fine on our Mac arm64 universal binary static-dependencies `.pkg` build, and also fine on our Mac x64 building just for x64...

micahsnyder:
Silly me.  It was `unrar_iface` that was failing to build, because it includes `rar.hpp` which requires C++11 features.  Fixed with https://github.com/Cisco-Talos/clamav/pull/1008/commits/a8fdb2b3770b84caf19c2474bc98a3631f3a6a36

Now just to watch the test pipelines...

micahsnyder:
Testing looking good after that last fix.

Issue #1030, opened 2023-09-21
Heuristics.Limits.Exceeded.MaxFileSize not emitted for 2GB+ files


fawind:
Describe the bug
----------------

ClamAV is currently unable to scan files larger then 2GB (https://github.com/Cisco-Talos/clamav/issues/344#issuecomment-954236634).

Given that, I would expect ClamAV to emit the `Heuristics.Limits.Exceeded.MaxFileSize` warning when encountering a file larger then 2GB and configured with `--alert-exceeds-max=true`.

However, the MaxFileSize warning seems to be emitted based on the value configured through the `--max-filesize` option. I.e. when running with `--max-filesize=4000M`, ClamAV will not emit a MaxFileSize warning, even when encountering and skipping files larger then 2GB.

You can workaround this by setting `--max-filesize` to 2GB, which makes ClamAV emit warnings for 2GB+ files. However this is not obvious and very easy to misconfigure, resulting in files silently getting skipped without emitting the respective warning.

Ideally, ClamAV would always emit the MaxFileSize warning when it skips scanning bytes of files due to their size. Alternatively, it might also make sense to restrict the `--max-filesize` option to not allow setting values larger then 2GB or at least emit a warning if users misconfigure this.

How to reproduce the problem
----------------------------

Given `my-big-file` with size 2.5GB, the following command does not emit the MaxFileSize warning:

```bash
$ clamscan -d ./db --max-filesize=4000M --alert-exceeds-max=true my-big-file
Loading:     0s, ETA:   0s [========================>]        1/1 sigs
Compiling:   0s, ETA:   0s [========================>]       10/10 tasks

/tmp/clamav-limits/my-big-file: OK

----------- SCAN SUMMARY -----------
Known viruses: 1
Engine version: 1.2.0
Scanned directories: 0
Scanned files: 1
Infected files: 0
Data scanned: 0.00 MB
Data read: 2500.00 MB (ratio 0.00:1)
Time: 4.383 sec (0 m 4 s)
```

Note that a workaround is to set `--max-filesize` to 2GB, where the warning is emitted:

```bash
$ clamscan -d ./db --max-filesize=2048M --alert-exceeds-max=true my-big-file
Loading:     0s, ETA:   0s [========================>]        1/1 sigs
Compiling:   0s, ETA:   0s [========================>]       10/10 tasks

/tmp/clamav-limits/my-big-file: Heuristics.Limits.Exceeded.MaxFileSize FOUND

[...]

PR #1032, opened 2023-09-23
Fix alert-exceeds-max feature for files > 2GB and < max-filesize

micahsnyder:
The --alert-exceeds-max feature should alert for all files larger than 2GB because 2GB is the internal limit for individual files. This isn't working correctly because the `goto done;` exit condition after recording the exceeds-max heuristic skips over the logic that reports the alert.

This fix moves the ">2GB" check up to the location where the max-filesize engine option is set by clamd or clamscan. If max-filesize > 2GB - 1 is requested, then max-filesize is set to 2GB - 1.

Additionally, a warning is printed if max-filesize > 2GB is requested (with an exception for when it's maxed out by setting --max-filesize=0).

Resolves: https://github.com/Cisco-Talos/clamav/issues/1030

### DIFF
--- a/libclamav/others.c
+++ b/libclamav/others.c
@@ -623,7 +623,20 @@ cl_error_t cl_engine_set_num(struct cl_engine *engine, enum cl_engine_field fiel
             engine->maxscansize = num;
             break;
         case CL_ENGINE_MAX_FILESIZE:
-            engine->maxfilesize = num;
+            /* We have a limit of around 2GB (INT_MAX - 2). Enforce it here.
+             *
+             * TODO: Large file support is large-ly untested. Remove this restriction and test with a large set of large files of various types.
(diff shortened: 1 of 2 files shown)


micahsnyder:
Note: I'm actually a little bit on the fence about the warning when setting max-filesize > 2GB.  I expect it will draw a lot of attention.  However, a lot of people mistakenly think the limit is 4GB or even unlimited because there is presently no warning.

ragusaa:
> Note: I'm actually a little bit on the fence about the warning when setting max-filesize > 2GB. I expect it will draw a lot of attention. However, a lot of people mistakenly think the limit is 4GB or even unlimited because there is presently no warning.

I like having a warning that a file is not being scanned due to a size limit, so that the user is aware.

ragusaa (approved):

Issue #1028, opened 2023-09-19
ClamD fails silently if temporary folder does not exist


electrosonics:
Describe the bug
----------------
ClamD fails silently if temporary folder does not exist
----------------

While testing the lastest version of ClamD with ClamDScan.exe, I was not getting any positive results for test files. ClamDscan reported that it took 0.005 second to scan a 20 megabyte file. I traced the issue to the ClamD.conf file TemporaryDirectory entry. It contained a typo. After I corrected the typo and restarted the ClamD service, the test file took a lot longer to scan and detections were now positive.

It would be nice if ClamDScan could verify that the TemporaryDirectory path is valid and report the error if it is not. This would have saved a lot of time debugging this issue.

How to reproduce the problem
----------------------------

To reproduce this error, edit the ClamD.conf file and change the TemporaryDirectory to an non-existent folder, restart the ClamD service and use ClamDScan to test a file with a known issue (or test a large file and note that the scan time in non-zero).

Config file: clamd.conf
-----------------------
LogFile = "C:\Program Files (x86)\[snip]\Antivirus\ClamAV\clamd.log"
LogFileUnlock = "yes"
LogTime = "yes"
LogVerbose = "yes"
ExtendedDetectionInfo = "yes"
TemporaryDirectory = "C:\Program Files (x86)\[snip]\Scratch"
DatabaseDirectory = "C:\Program Files (x86)\[snip]\Antivirus\ClamAV\db"
TCPSocket = "3310"
TCPAddr = "127.0.0.1"
MaxConnectionQueueLength = "30"
StreamMaxLength = "52428800"
MaxThreads = "100"
ReadTimeout = "60"
MaxQueue = "200"
IdleTimeout = "60"
SelfCheck = "1800"
ExitOnOOM = "yes"
Debug = "yes"

Config file: freshclam.conf
---------------------------
LogTime = "yes"
DatabaseDirectory = "C:\Program Files (x86)\[snip]\Antivirus\ClamAV\db"
UpdateLogFile = "C:\Program Files (x86)\[snip]\Antivirus\ClamAV\freshclam.log"
DatabaseMirror = "database.clamav.net"

clamav-milter.conf not found

Software settings
-----------------
Version: 1.2.0
Optional features supported: MEMPOOL AUTOIT_EA06 BZIP2 LIBXML2 PCRE2 JSON RAR

Database information
--------------------
Database directory: C:\Program Files (x86)\[snip]\Antivirus\ClamAV\db
[...]

PR #1037, opened 2023-09-27
Adds logging and changes return status for bad temp dir: Issue #1028

Kiggins26:
Resolves: https://github.com/Cisco-Talos/clamav/issues/1028

### DIFF
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -568,6 +568,13 @@ int main(int argc, char **argv)
                 ret = 1;
                 break;
             }
+
+            STATBUF sb;
+            if (CLAMSTAT(opt->strarg, &sb) != 0 && !S_ISDIR(sb.st_mode)) {
+                logg(LOGG_ERROR, "Current configuration of TemporaryDirectory: %s does not exist, or is not valid \n", opt->strarg);
+                ret = 1;
+                break;
+            }
         }
 
         cl_engine_set_clcb_hash(engine, hash_callback);
--- a/clamd/session.c
+++ b/clamd/session.c
@@ -591,8 +591,9 @@ int execute_or_dispatch_command(client_conn_t *conn, enum commands cmd, const ch
         }
         case COMMAND_INSTREAM: {
             int rc = cli_gentempfd(optget(conn->opts, "TemporaryDirectory")->strarg, &conn->filename, &conn->scanfd);
-            if (rc != CL_SUCCESS)
-                return rc;
+            if (rc != CL_SUCCESS) {
+                return 1;
+            }
             conn->quota = optget(conn->opts, "StreamMaxLength")->numarg;
             conn->mode  = MODE_STREAM;
             return 0;


clamd/clamd.c
```diff
@@ -568,6 +568,12 @@ int main(int argc, char **argv)
                 ret = 1;
                 break;
             }
+
+            if (stat(CL_ENGINE_TMPDIR, &sb) != 0 && !S_ISDIR(sb.st_mode)) {
+                logg(LOGG_ERROR, "%s is not a valid dir\n", CL_ENGINE_TMPDIR);
```
micahsnyder, line 573: It would be useful to explain why we care about this directory, in case it isn't apparent. 
Something like:
`"The temporary directory is configured to be '%s' but this directory does not exist or is not a valid directory.\n"`

clamd/session.c
```diff
@@ -591,8 +591,10 @@ int execute_or_dispatch_command(client_conn_t *conn, enum commands cmd, const ch
         }
         case COMMAND_INSTREAM: {
             int rc = cli_gentempfd(optget(conn->opts, "TemporaryDirectory")->strarg, &conn->filename, &conn->scanfd);
-            if (rc != CL_SUCCESS)
-                return rc;
+            if (rc != CL_SUCCESS) {
+                logg(LOGG_ERROR, "Failed to generate file in TemporaryDirectory, filename: %s \n", &conn->filename);
```
micahsnyder, line 595: There should already be a log message when `cli_gentempfd` fails, but I suppose an additional error is okay. 

https://github.com/Cisco-Talos/clamav/blob/7d26bc8f6a801bb35ff7f2b154f33817e5863ea6/libclamav/others_common.c#L1249-L1322

micahsnyder (changes requested):
A little more detail explaining why we care of the directory is invalid/doesn't exist would be good.  Otherwise this seems like the right fix to me.

clamd/session.c
Kiggins26: o7, resolving comment then

clamd/clamd.c
Kiggins26: updated the log to `Current configuration of TemporaryDirectory: %s does not exist, or is not valid`

clamd/clamd.c
```diff
@@ -568,6 +568,12 @@ int main(int argc, char **argv)
                 ret = 1;
                 break;
             }
+
+            if (stat(CL_ENGINE_TMPDIR, &sb) != 0 && !S_ISDIR(sb.st_mode)) {
+                logg(LOGG_ERROR, "Current configuration of TemporaryDirectory: %s does not exist, or is not valid", CL_ENGINE_TMPDIR);
```
micahsnyder, line 573: I just noticed the log message is missing the `\n` new line at the end.

clamd/clamd.c
```diff
@@ -568,6 +568,12 @@
...
```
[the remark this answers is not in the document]
micahsnyder: Oh this is silly of me to not notice this earlier.  The string variable for the temp directory is `opt->strarg`, not `CL_ENGINE_TMPDIR`.

clamd/clamd.c
```diff
@@ -568,6 +568,12 @@ int main(int argc, char **argv)
                 ret = 1;
                 break;
             }
+
+            if (stat(CL_ENGINE_TMPDIR, &sb) != 0 && !S_ISDIR(sb.st_mode)) {
```
micahsnyder, line 572: In addition to swithcing to `opt->strarg`, we should probably use `CLAMSTAT` instead of `stat`. 
```suggestion
            if (CLAMSTAT(opt->strarg, &sb) != 0 && !S_ISDIR(sb.st_mode)) {
```

clamd/clamd.c
micahsnyder, line 573:
```suggestion
                logg(LOGG_ERROR, "Current configuration of TemporaryDirectory: %s does not exist, or is not valid\n", opt->strarg);
```

clamd/clamd.c
micahsnyder: You will need to also declare the `sb` variable at the top of this block, like:
```
        STATBUF sb;
```

micahsnyder (changes requested):

micahsnyder (approved):
Looks good to me. Thanks for the help @Kiggins26!

Issue #344, opened 2021-10-28
Scanning files bigger then 2GB


kwon-kihong:
**Describe the bug**
-----
The file with problem has over 2GB of data always scan skipped 

Data scanned: 0.00 MB

**ClamAV Version**
------
0.103.3

**How to reproduce the problem**
------
```
dd if=/dev/zero of=2G.test bs=1 count=0 seek=2G
clamscan -v -a --stdout -d /app/clamav_defs ./2G.test --max-filesize 4000M --max-scansize 4000M

LibClamAV Warning: **************************************************
LibClamAV Warning: ***  The virus database is older than 7 days!  ***
LibClamAV Warning: ***   Please update it as soon as possible.    ***
LibClamAV Warning: **************************************************
Scanning /app/2G.test
/app/2G.test: OK

----------- SCAN SUMMARY -----------
Known viruses: 8543862
Engine version: 0.103.3
Scanned directories: 0
Scanned files: 1
Infected files: 0
Data scanned: 0.00 MB
Data read: 2048.00 MB (ratio 0.00:1)
Time: 22.835 sec (0 m 22 s)
Start Date: 2021:10:28 03:16:06
End Date:   2021:10:28 03:16:29
```

jhawkins1:
We are using CLAMAV and have a requirement for handling files larger than 4GB (MPEGs, ISOs, ZIP, etc...).   We are looking at potentially making the required changes, and associated testing, to CLAMAV.   Assuming the Project contribution guidelines are followed, would the project accept patches to support this new functionality and include into the mainline?   Is anyone already working on this that we could assist or collaborate ?

sviscapi:
Dear all,

@jhawkins1 : that would really be appreciated :) We also have a use case involving large files (up to 50GB) for long term preservation:

https://www.programmevitam.fr/pages/english/pres_english/
https://github.com/ProgrammeVitam/vitam

So far I've been told ClamAV won't currently handle files larger than 4GB. Is that correct ?

Best regards,

Samuel, for Conseil Départemental de l'Hérault
https://herault.fr/

micahsnyder:
No one is working on this right now. 

[...]

PR #1041, opened 2023-09-30
Fix alert-exceeds-max feature for files > 2GB and < max-filesize (1.0.4)

micahsnyder:
Backport of https://github.com/Cisco-Talos/clamav/pull/1032

---

The --alert-exceeds-max feature should alert for all files larger than 2GB because 2GB is the internal limit for individual files. This isn't working correctly because the `goto done;` exit condition after recording the exceeds-max heuristic skips over the logic that reports the alert.

This fix moves the ">2GB" check up to the location where the max-filesize engine option is set by clamd or clamscan. If max-filesize > 2GB - 1 is requested, then max-filesize is set to 2GB - 1.

Additionally, a warning is printed if max-filesize > 2GB is requested (with an exception for when it's maxed out by setting --max-filesize=0).

Resolves: https://github.com/Cisco-Talos/clamav/issues/1030

### DIFF
--- a/libclamav/others.c
+++ b/libclamav/others.c
@@ -622,7 +622,20 @@ cl_error_t cl_engine_set_num(struct cl_engine *engine, enum cl_engine_field fiel
             engine->maxscansize = num;
             break;
         case CL_ENGINE_MAX_FILESIZE:
-            engine->maxfilesize = num;
+            /* We have a limit of around 2GB (INT_MAX - 2). Enforce it here.
+             *
+             * TODO: Large file support is large-ly untested. Remove this restriction and test with a large set of large files of various types.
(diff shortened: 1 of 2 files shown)


Raj-2024:
Why is there a limit of 2 Gb ? How can we get ClamAV to scan files bigger than 2 GB up to 4GB ?

micahsnyder:
> Why is there a limit of 2 Gb ? How can we get ClamAV to scan files bigger than 2 GB up to 4GB ?

@Raj-2024 The ClamAV source code has a lot of older modules that were not designed with large files in mind. It would require a careful audit to identify and upgrade all areas in ClamAV to safely handle files larger than 2GB.

The ticket requesting this change to support files larger than 2GB is here: https://github.com/Cisco-Talos/clamav/issues/344

Issue #793, opened 2022-12-14
clamav-clamonacc 1.0.0 stuck on 100% cpu Usage if non-existent OnAccessIncludePath Directory is given


amsnek:
clamav version: 1.0.0 

when using a non existant directory in clamd.conf:
```
OnAccessIncludePath /foobar
```

clamav-clamonacc will be stuck on 100% cpu usage, even though running in --verbose shows no errors / running seemingly normally (scanning other existing OnAccessIncludePath correctly)

expected behavior: 
WARN that directory does not exist and continuing normally

ragusaa:
Hi,

Sorry for the delay.  I am unable to reproduce this.  Could you tell me what platform you are running on, and could you share your clamd.conf file?

Thanks,
Andy

amsnek:
Hello @ragusaa 

No Problem! This happens on rhel7 and rhel8 for me.

clamd.conf
```
ExtendedDetectionInfo yes
LocalSocket /run/clamav/clamd.ctl
PidFile /var/run/clamd.pid
LocalSocketGroup clamav
LocalSocketMode 660
FixStaleSocket yes
FollowDirectorySymlinks no
FollowFileSymlinks no
CrossFilesystems no
ConcurrentDatabaseReload no
User clamav
OnAccessPrevention no
OnAccessExtraScanning yes
ExcludePath ^/proc/
ExcludePath ^/sys/
ExcludePath ^/dev/
OnAccessExcludePath /usr/lib64
OnAccessExcludePath /usr/lib
OnAccessExcludePath /var/log
OnAccessExcludePath /var/lib
OnAccessExcludePath /var/log/audit
OnAccessIncludePath /foobar
OnAccessExcludeRootUID no
OnAccessExcludeUname clamav
DatabaseDirectory /var/lib/clamav
OnAccessRetryAttempts 1
MaxThreads 64
OnAccessMaxThreads 24
MaxQueue 192
```

output from clamconf
```
Checking configuration files in /etc/clamav

Config file: clamd.conf
-----------------------
ExtendedDetectionInfo = "yes"
PidFile = "/var/run/clamd.pid"
DatabaseDirectory = "/var/lib/clamav"
LocalSocket = "/run/clamav/clamd.ctl"
LocalSocketGroup = "clamav"
LocalSocketMode = "660"
MaxThreads = "64"
MaxQueue = "192"
ExcludePath = "^/proc/", "^/sys/", "^/dev/"
CrossFilesystems disabled
[...]

PR #1047, opened 2023-10-06
Fixed clamonacc infinite loop when watch directory doesn't exist

ragusaa:
Fixes: https://github.com/Cisco-Talos/clamav/issues/793

### DIFF
--- a/clamonacc/inotif/inotif.c
+++ b/clamonacc/inotif/inotif.c
@@ -598,8 +598,8 @@ void *onas_ddd_th(void *arg)
                         }
                     }
                 }
-                pt = (struct optstruct *)pt->nextarg;
             }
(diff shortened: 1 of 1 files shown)


micahsnyder (approved):
In testing this does indeed fix the bug and scanning of other valid include-paths do work after this change.

There is unfortunately no error or warning message when the OnAccessIncludePath does not exist.  But no adverse effect either, now. 

I played with it a tiny bit to try and put in a warning message.  I was able to, but it didn't feel like the right spot and I wasn't sure if there would be adverse effects to that... so I think I wanna accept this as-is.

opoplawski:
Is this going to get back-ported to the 1.0 LTS branch?  It would be appreciated.

Issue #819, opened 2023-01-26
Feature request: Extraction of attachments inside OneNote documents


LSchuepbach:
Describe the bug
----------------

I'd wish that attached files inside OneNote `.one` documents would be correctly recognized and extracted when using the `--leave-temps` option.

As mentioned in [Bleeping Computer](https://www.bleepingcomputer.com/news/security/hackers-now-use-microsoft-onenote-attachments-to-spread-malware/), OneNote is becoming a new way of spreading malwares.

How to reproduce the problem
----------------------------

Using 1.0.0 in docker, the .one file is considered as binary, --leave-temps leaves nothing.

[https://blog.didierstevens.com/2023/01/22/analyzing-malicious-onenote-documents/](https://blog.didierstevens.com/2023/01/22/analyzing-malicious-onenote-documents/) gives a sample, link to the document by Microsoft explaining the structure and a python script that recognizes the boundary and extracts the attachments.

I have access to a couple more samples if someone is interested.

SecT0uch:
@LSchuepbach Small world :wave:

PR #1048, opened 2023-10-08
Support for extracting attachments from OneNote section files

micahsnyder:
Resolves: https://github.com/Cisco-Talos/clamav/issues/819

Also includes rudimentary support in libclamav Rust code for getting slices from FMap's and for interacting with libclamav's context structure.


micahsnyder:
I added 3 test tiles to demonstrate extraction of the clam.exe program from OneNote section files exported by OneNote 2007, OneNote 2010, and a recent Office364 OneNote export.  If extraction were to fail, it would fail the libclamav unit tests as well as some clamd and clamscan tests. 

I have temporarily set the `onenote_parser` module to pull from https://github.com/micahsnyder/onenote.rs/tree/CLAM-2329-new-from-slice pending this PR: https://github.com/msiemens/onenote.rs/issues/12

I think it's ready for review.

libclamav/scanners.c
```diff
@@ -979,6 +979,190 @@ static cl_error_t cli_scanegg(cli_ctx *ctx)
     return status;
 }
 
+// static cl_error_t cli_scanonenote(cli_ctx *ctx)
```
ragusaa, line 982: Do you still need this?

libclamav_rust/build.rs
```diff
@@ -141,7 +145,9 @@ fn execute_bindgen() -> Result<(), &'static str> {
         // are probably not the same when generated as when compiled.
         .layout_tests(false)
         // Enable bindgen to find generated headers in the build directory, too.
-        .clang_arg(build_include_path);
+        .clang_arg(build_include_path)
+        // Hack to include 3rd party C libary headers
+        .clang_arg("-I/home/micah/.mussels/install/host-static/include");
```
ragusaa, line 150: What is the plan for removing this hardcoded path?

ragusaa (changes requested):
Do we need to address the hardcoded path before merging?

libclamav_rust/build.rs
micahsnyder: Ooops!  Good catch!  This shouldn't stay in the PR. I will remove it.

I needed it to make `MAINTAINER_MODE` work.  It is currently broken.  The real fix (later, see CLAM-2451) will be to provide 3rd party library include paths from cmake to build.rs `MAINTEINER_MODE` is enabled so that bindgen can find them.

libclamav/scanners.c
micahsnyder: Urgh.  I do not. Forgot all about it.  Will remove.

micahsnyder:
Ok fixed to remove hardcoded path and dead code.

ragusaa (approved):

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+
+            let remaining = if let Some(pos) = find_bytes(remaining_vec, file_data_store_object) {
+                let (_, remaining) = remaining_vec.split_at(pos);
+                // Found file data store object.
+                remaining
+            } else {
+                return None;
+            };
+
+            let data_length_bytes = &remaining[16..20];
```
shutton, line 288: Another bytes-to-int transform

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+            let data_length = ((data_length_bytes[3] as u64) << 24)
+                | ((data_length_bytes[2] as u64) << 16)
+                | ((data_length_bytes[1] as u64) << 8)
+                | (data_length_bytes[0] as u64);
+
+            let data: &[u8] = &remaining[mem::size_of::<FileDataHeader>()
```
shutton, line 294: If it makes the code more readable, you could define:
``` rust
const SIZE_OF_FILE_DATA_HEADER: usize = mem::size_of::<FileDataHeader>;
```
near the top of the file.  `mem::size_of` is a const fn, so you can do that.  It doesn't need to be runtime.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+    fn next(&mut self) -> Option<ExtractedFile> {
+        // Find the next embedded file
+        if self.remaining.is_some() {
+            // Data stored in a slice.
+            if let Some(file) = self.next_file() {
+                // Decode the base64 encoded image
+                Some(file)
+            } else {
+                None
+            }
```
shutton, line 334: I think you can boil this whole thing down to:

```suggestion
            self.next_file()
```

You were just unwrapping and re-wrapping the `Option<>`

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+        } else if self.remaining_vec.is_some() {
+            // Data stored in a Vec.
+            if let Some(file) = self.next_file_vec() {
```
shutton, line 337: Ditto

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+                // Decode the base64 encoded image
+                Some(file)
+            } else {
+                None
+            }
+        } else if self.embedded_files.len() > 0 {
+            // Data stored in a Vec.
```
shutton, line 344: Ditto

libclamav_rust/src/fmap.rs
```diff
@@ -0,0 +1,102 @@
...
+    UninitializedPtr(&'static str),
+}
+
+#[derive(PartialEq, Eq, Hash, Debug)]
+pub struct FMap {
+    fmap_ptr: *mut sys::cl_fmap_t,
+}
+
+impl TryFrom<*mut sys::cl_fmap_t> for FMap {
+    type Error = &'static str;
```
shutton, line 52: It's generally a good idea to utilize an error type that implements `std::error::Error`, as it guarantees certain behaviors that will be expected upstream.  `&str` kinda works because it implements `Display`, which is one of `std::error::Error`'s bounds, but it doens't implement the entire trait.

libclamav_rust/src/fmap.rs
```diff
@@ -0,0 +1,102 @@
...
+    NotContained(usize, usize, usize),
+
+    #[error("FMap pointer not initialized: {0}")]
+    UninitializedPtr(&'static str),
+}
+
+#[derive(PartialEq, Eq, Hash, Debug)]
+pub struct FMap {
+    fmap_ptr: *mut sys::cl_fmap_t,
+}
```
shutton, line 49: If this is all this struct will ever contain, it's probably going to save you a few keystrokes to just make it a tuple struct:

```suggestion
pub struct FMap(*mut sys::cl_fmap_t);
```

libclamav_rust/src/fmap.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn try_from(value: *mut sys::cl_fmap_t) -> Result<Self, Self::Error> {
+        if value.is_null() {
+            return Err::<Self, &str>("Attempted to create Rust FMap interface from NULL pointer");
+        }
+
+        Ok(FMap { fmap_ptr: value })
+    }
+}
+
+impl<'a> FMap {
```
shutton, line 63: You shouldn't need a lifetime here -- the struct doesn't contain any references.

libclamav_rust/src/scanners.rs
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    let ret = unsafe { cli_magic_scan_buff(ptr as *const c_void, len, ctx, name_ptr, 0) };
+
+    if ret != cl_error_t_CL_SUCCESS {
+        debug!("cli_magic_scan_buff returned error: {}", ret);
+    }
+
+    // Okay now safe to drop the name CString.
+    drop(unsafe { CString::from_raw(name_ptr) });
```
shutton, line 68: An alternative:
```suggestion
    let _ = unsafe { CString::from_raw(name_ptr) };
```
That gives the compiler control over when it gets dropped, which is exactly the same here.

libclamav_rust/src/scanners.rs
```diff
@@ -0,0 +1,124 @@
...
+    ret
+}
+
+/// Scan a OneNote file for attachments
+#[no_mangle]
+pub unsafe extern "C" fn scan_onenote(ctx: *mut cli_ctx) -> cl_error_t {
+    let fmap = match ctx::current_fmap(ctx) {
+        Ok(fmap) => fmap,
+        Err(e) => {
+            warn!("Error getting FMap from ctx: {}", e);
```
shutton, line 79:
```suggestion
            warn!("Error getting FMap from ctx: {e}");
```

libclamav_rust/src/scanners.rs
```diff
@@ -0,0 +1,124 @@
...
+            return cl_error_t_CL_ERROR;
+        }
+    };
+
+    let file_bytes = match fmap.need_off(0, fmap.len()) {
+        Ok(bytes) => bytes,
+        Err(err) => {
+            error!(
+                "Failed to get file bytes for fmap of size {}: {}",
```
shutton, line 88: `err` implements `Display`
```suggestion
                "Failed to get file bytes for fmap of size {}: {err}",
```

libclamav_rust/src/scanners.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    let one = OneDump::from_bytes(file_bytes, fmap.name());
+
+    let one = match one {
+        Ok(x) => x,
+        Err(err) => {
+            error!("Failed to parse OneNote file: {}", err.to_string());
+            return cl_error_t_CL_ERROR;
+        }
+    };
```
shutton, line 104: I tried to make this prettier, but this was the best I could manage, since this function doesn't return a `Result<>`

```suggestion
    let one = match OneDump::from_bytes(file_bytes, fmap.name()) {
        Ok(one) => one,
        Err(err) => {
            error!("Failed to parse OneNote file: {}", err.to_string());
            return cl_error_t_CL_ERROR;
        }
    };
```

libclamav_rust/src/scanners.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    one.into_iter().enumerate().all(|(_, attachment)| {
```
shutton, line 106: Do you still need `.enumerate()`? You're throwing the index out.

Also, you're calling `.all()`, but always returning `true`.  I think you probably wanted a `.for_each()` here.

libclamav_rust/src/util.rs
```diff
@@ -46,3 +48,23 @@ pub fn file_from_fd_or_handle(fd: i32) -> File {
...
     compile_error!("implemented only for unix and windows targets")
 }
+
+/// Get a string from a pointer
+///
+/// # Safety
+///
+/// The caller is responsible for making sure the lifetime of the pointer
+/// exceeds the lifetime of the output string.
+pub fn optional_cstring(ptr: *const i8) -> Option<&'static str> {
```
shutton, line 58: This is kind of a weird semantic. You're doing the Result->Option conversion for the caller, and denying them the error (if they want it). They could do that themselves with `.ok()`

Maybe deserves a second look.  This could be genericized a bit as `cstr_from_ptr`, and the calls might read a little better.

shutton (changes requested):
I didn't flag all the spots where you could probably pull out some `const`'s in `onenote.rs` for all those hardcoded hex strings.

libclamav_rust/Cargo.toml
```diff
@@ -20,6 +20,7 @@ base64 = "0.21.0"
 sha1 = "0.10.5"
 unicode-segmentation = "1.10.1"
 bindgen = "0.65"
+onenote_parser = { git = "https://github.com/micahsnyder/onenote.rs.git", branch = "CLAM-2329-new-from-slice" }
```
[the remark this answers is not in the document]
micahsnyder: It's either that or we have it depend on my fork for the next release. I'm not sure if that would upset anyone.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+use std::path::PathBuf;
+
+use log::{debug, error};
+use thiserror::Error;
+
+use onenote_parser;
+
+/// OneDumpError enumerates all possible errors returned by this library.
+#[derive(Error, Debug)]
+pub enum OneDumpError {
```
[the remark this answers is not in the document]
micahsnyder: Love it. Will do that.  I'll also rename all of the "OneDump"'s to "OneNote".

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+        .position(|window| window == needle)
+}
+
+pub struct ExtractedFile {
+    pub name: Option<String>,
+    pub data: Vec<u8>,
+}
+
+#[derive(Default)]
+pub struct OneDump<'a> {
```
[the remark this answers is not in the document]
micahsnyder: Will change OneDump to OneNote.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+    embedded_files: Vec<ExtractedFile>,
+    remaining_vec: Option<Vec<u8>>,
+    remaining: Option<&'a [u8]>,
+}
+
+// https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-onestore/8806fd18-6735-4874-b111-227b83eaac26
+#[repr(packed)]
+struct FileDataHeader {
```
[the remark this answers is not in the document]
micahsnyder: I like it.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+    _guid_header: [u8; 16],
+    _cb_length: u64,
+    _unused: u32,
+    _reserved: u64,
+}
+
+impl<'a> OneDump<'a> {
+    pub fn from_bytes(data: &'a [u8], filename: &str) -> Result<OneDump<'a>, OneDumpError> {
+        debug!(
+            "Inspecting OneNote file for attachments from in-memory buffer of size {}-bytes named {}\n",
```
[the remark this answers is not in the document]
micahsnyder: It turns out I can't make this change if changing `filename` from `&str` to `&Path` because `Path` does not implement `Display`.  So I'll have to do `filename.to_string_loss()`.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+
+        fn parse_section_buffer(
+            data: &[u8],
+            filename: &str,
+        ) -> Result<Vec<ExtractedFile>, OneDumpError> {
+            let mut embedded_files: Vec<ExtractedFile> = vec![];
+            let mut parser = onenote_parser::Parser::new();
+
+            if let Ok(section) = parser.parse_section_buffer(data, &filename) {
+                // file appears to be OneStore 2.8 `.one` file.
```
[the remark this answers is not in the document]
micahsnyder: I've updated my PR that adds the `parse_section_buffer()` method so it takes a `&Path` and then make this change.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+
+        // Try to parse the section buffer using the onenote_parser crate.
+        // Attempt to catch panics in case the parser encounter unexpected issues.
+        let result_result = panic::catch_unwind(|| -> Result<Vec<ExtractedFile>, OneDumpError> {
+            let embedded_files = parse_section_buffer(data, filename)?;
+            Ok(embedded_files)
+        });
+
+        // Check if it panicked. If no panic, grab the parse result.
+        let result = match result_result {
```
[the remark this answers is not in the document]
micahsnyder: Love it.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+            Ok(result) => result,
+            Err(_) => return Err(OneDumpError::Parse),
+        };
+
+        if let Ok(embedded_files) = result {
+            // Successfully parsed the OneNote file with the onenote_parser crate.
+            Ok(OneDump {
+                embedded_files: embedded_files,
+                remaining_vec: None,
```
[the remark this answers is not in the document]
micahsnyder: Awesome good idea

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+                remaining: None,
+            })
+        } else {
+            debug!("Unable to parse OneNote file with onenote_parser crate. Trying a different method known to work with older office 2010 OneNote files to extract attachments.");
+
+            let embedded_files: Vec<ExtractedFile> = vec![];
+
+            const ONE_MAGIC: &[u8] = &[
```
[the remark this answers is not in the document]
micahsnyder: Neat!! I'll grab that.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+
+            Ok(OneDump {
+                embedded_files: embedded_files,
+                remaining_vec: None,
+                remaining: Some(data),
+            })
+        }
+    }
+
+    pub fn find_next_file(&'a mut self, remaining_in: &'a [u8]) -> Option<&'a [u8]> {
```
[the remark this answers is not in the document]
micahsnyder: Turns out I don't think I need this function at all. I think it was experimentation dead code.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+            // Found file data store object.
+            remaining
+        } else {
+            // No occurence of "url"
+            // No more 'url's.
+            return None;
+        };
+
+        let data_length_bytes = &remaining[16..4];
+        let data_length = ((data_length_bytes[1] as u64) << 24)
```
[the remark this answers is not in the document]
micahsnyder: Nice! I had a feeling there was a better way for this. Thanks!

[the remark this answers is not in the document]
micahsnyder: Actually, this change causes a panic in testing on that unwrap.

```

4:
4: [DEBUG]: stderr: thread '<unnamed>' panicked at 'range end index 5 out of range for slice of length 4', libclamav_rust/src/onenote.rs:206:50
4: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
4: fatal runtime error: failed to initiate panic, error 5
4: Aborted
4:
4:
4: --------------------------- Captured stdout teardown ---------------------------
4:
4: [INFO]: Removed tmp directory: /home/micah/workspace/clamav-micah-3/build/unit_tests/TC-j2jj_sey
4: ===================== 6 failed, 46 passed in 3.41 seconds ======================
1/1 Test #4: clamscan .........................***Failed    3.90 sec
```

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+            let remaining = if let Some(pos) = find_bytes(remaining_in, file_data_store_object) {
+                let (_, remaining) = remaining_in.split_at(pos);
+                // Found file data store object.
+                remaining
+            } else {
+                return None;
+            };
+
+            let data_length_bytes = &remaining[16..20];
+            let data_length = ((data_length_bytes[3] as u64) << 24)
```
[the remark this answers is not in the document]
micahsnyder: I'm worried this one will panic as well.

libclamav_rust/src/onenote.rs
micahsnyder: Yeah tis a bit better. Thanks

libclamav_rust/src/onenote.rs
micahsnyder: Hahahaha so much better. Silly me.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+        self.remaining = remaining;
+
+        if let Some(file_data) = file_data {
+            Some(ExtractedFile {
+                name: None,
+                data: file_data,
+            })
+        } else {
+            None
+        }
```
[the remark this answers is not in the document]
micahsnyder: Oh that's nice.

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+#[repr(packed)]
+struct FileDataHeader {
+    _guid_header: [u8; 16],
+    _cb_length: u64,
+    _unused: u32,
+    _reserved: u64,
+}
+
+impl<'a> OneDump<'a> {
+    pub fn from_bytes(data: &'a [u8], filename: &str) -> Result<OneDump<'a>, OneDumpError> {
```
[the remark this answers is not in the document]
micahsnyder: Good point.  I've switched it to `&Path` and changed it in the PR for the onenote_parser crate as well.

libclamav_rust/src/fmap.rs
micahsnyder: At least for now.  Someday I imagine we will want to change it from a wrapper to actually implement the fmap stuff in Rust. 

So TBH not sure if a tuple is better or worse for the future.

libclamav_rust/src/onenote.rs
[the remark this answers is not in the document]
shutton: It gets even simpler if `OneDumpError::Parse` takes a value (as a tuple) that happens to be the error type you're mapping.  Then you can drop the closure syntax completely.

libclamav_rust/src/onenote.rs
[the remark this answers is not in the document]
shutton: I might've botched the translation, though where'd it get index `5` from?

libclamav_rust/src/onenote.rs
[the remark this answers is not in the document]
micahsnyder: I admit it would be nice to get a better explanation for a panic than just "Parse error" but the argument passed back is simple `Box<dyn Any + Send, Global>` and I don't know how to get meaningful information from that.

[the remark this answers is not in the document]
shutton: Blame my inability to figure out the context within the [lousy] Github comment viewer.  I didn't realize that was the error from a panic trap.

> Unlike with normal errors, this value doesn’t implement the [Error](https://doc.rust-lang.org/std/error/trait.Error.html) trait.

Bleh...

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,355 @@
...
+
+        Some(file_data)
+    }
+
+    pub fn next_file(&mut self) -> Option<ExtractedFile> {
+        debug!("Looking to extract file from OneNote section...");
+
+        let mut file_data: Option<Vec<u8>> = None;
+
+        let remaining = if let Some(remaining_in) = self.remaining {
```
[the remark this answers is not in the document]
micahsnyder: Hm doesn't quite work.  I don't see a cleaner option and I'm decently happy with the current code here.

libclamav_rust/src/fmap.rs
micahsnyder: If I do
```rust
impl TryFrom<*mut sys::cl_fmap_t> for FMap {
    type Error = dyn std::error::Error;
```
then I get
```
error[E0277]: the size for values of type `(dyn std::error::Error + 'static)` cannot be known at compilation time
   --> libclamav_rust/src/fmap.rs:54:48
    |
54  |     fn try_from(value: *mut sys::cl_fmap_t) -> Result<Self, Self::Error> {
    |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `Sized` is not implemented for `(dyn std::error::Error + 'static)`
note: required by a bound in `Result`
   --> /home/micah/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:502:20
    |
502 | pub enum Result<T, E> {
    |                    ^ required by this bound in `Result`

```

libclamav_rust/src/fmap.rs
micahsnyder: I think I need it to tell the compiler that the `FMap` must outlive the slice returned by `need_off`

libclamav_rust/src/util.rs
micahsnyder: Good catch on the enumerate, but regarding `all()`, no we return `false` if there is an error.

libclamav_rust/src/scanners.rs
micahsnyder: Good catch on the enumerate, but regarding `all()`, no we return `false` if there is an error.

libclamav_rust/src/fmap.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    pub fn name(&self) -> &'static str {
+        let name_result = unsafe { str_from_ptr((*self.fmap_ptr).name) };
+        match name_result {
+            Ok(Some(name)) => name,
+            _ => "",
+        }
+    }
+}
```
shutton, line 110:
```suggestion
    pub fn name(&self) -> &'static str {
        unsafe {
            str_from_ptr((*self.fmap_ptr).name)
                .unwrap_or(Some("<invalid-utf8>"))
                .unwrap_or("<unnamed>")
        }
    }
}
```

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,287 @@
...
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+ *  MA 02110-1301, USA.
+ */
+
+use std::mem;
+use std::panic;
+use std::path::Path;
+use std::path::PathBuf;
+
```
shutton, line 27:
```suggestion
use std::{
    mem, panic,
    path::{Path, PathBuf},
};
```

libclamav_rust/src/fmap.rs
shutton:
```suggestion
#[derive(Debug, thiserror::Error)]
pub enum FmapError {
    #[error("Attempted to create Rust FMap interface from NULL pointer")]
    Null,
}
```

libclamav_rust/src/fmap.rs
```diff
@@ -0,0 +1,110 @@
...
+        unsafe { (*self.fmap_ptr).len == 0 }
+    }
+
+    pub fn name(&self) -> &'static str {
+        let name_result = unsafe { str_from_ptr((*self.fmap_ptr).name) };
+        match name_result {
+            Ok(Some(name)) => name,
+            _ => "",
+        }
+    }
```
shutton, line 109:
```suggestion
        unsafe {
            str_from_ptr((*self.fmap_ptr).name)
                .unwrap_or(Some("<invalid-utf8>"))
                .unwrap_or("<unnamed>")
        }
    }
```

shutton (changes requested):

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,287 @@
...
+                embedded_files,
+                ..Default::default()
+            })
+        } else {
+            debug!("Unable to parse OneNote file with onenote_parser crate. Trying a different method known to work with older office 2010 OneNote files to extract attachments.");
+
+            let embedded_files: Vec<ExtractedFile> = vec![];
+
+            let file_magic = data.get(0..16).ok_or(Error::Format)?;
+            if file_magic != ONE_MAGIC {
```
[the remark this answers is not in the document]
micahsnyder: The reason the file magic is checked here is that the `onenote_parser` may handle new/different types of onenote documents now, or in the future, and should have it's own checks to verify that the file magic is correct for whichever types of onenote files it supports.  It's possible they will check the same magic sequence, or more, when we run `onenote_parser::parse_section_buffer()`

But our own parser, which is initiated in this spot, only knows about onenote documents starting with this `ONE_MAGIC` sequence.

I'll add a comment saying something along those lines.

libclamav_rust/build.rs
```diff
@@ -135,7 +139,7 @@ fn execute_bindgen() -> Result<(), &'static str> {
         // Silence code-style warnings for generated bindings.
         .raw_line("#![allow(non_snake_case, non_camel_case_types, non_upper_case_globals)]")
         // Make the bindings pretty.
-        .rustfmt_bindings(true)
+        .formatter(bindgen::Formatter::Rustfmt)
         // Disable the layout tests.
```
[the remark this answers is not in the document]
micahsnyder: `.rustfmt_bindings(true)` is deprecated, hence switching to `.formatter(bindgen::Formatter::Rustfmt)`

libclamav_rust/src/fmap.rs
micahsnyder: Seems nice. 👍

libclamav_rust/src/onenote.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// Open a OneNote document given the document was provided as a slice of bytes.
+    pub fn next_file(&mut self) -> Option<ExtractedFile> {
+        debug!("Looking to extract file from OneNote section...");
+
+        let mut file_data: Option<Vec<u8>> = None;
+
+        let remaining = if let Some(remaining_in) = self.remaining {
+            let remaining = if let Some(pos) = find_bytes(remaining_in, FILE_DATA_STORE_OBJECT) {
+                let (_, remaining) = remaining_in.split_at(pos);
```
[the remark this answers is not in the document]
micahsnyder: Huh? I don't follow

libclamav_rust/src/onenote.rs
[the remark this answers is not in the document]
micahsnyder: I guess I mean to say I don't think it is translated correctly. But, I'm fairly happy with the existing code.  Unless you have another suggestion I'll just leave as-is.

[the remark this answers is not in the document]
shutton: Discussed OOB...

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -34,14 +34,14 @@ use image::{imageops::FilterType::Lanczos3, DynamicImage, ImageBuffer, Luma, Pix
 use log::{debug, error, warn};
 use num_traits::{NumCast, ToPrimitive, Zero};
 use rustdct::DctPlanner;
-use thiserror::Error;
+use thiserror;
```
shutton, line 37: You shouldn't need this.
```suggestion
```

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -68,6 +68,9 @@ pub enum FuzzyHashError {
 
     #[error("{0} parmeter is NULL")]
     NullParam(&'static str),
+
+    #[error("{0} hash must be 16 characters in length")]
+    InvalidHashLength(&'static str),
```
shutton, line 73: Are all hashes going to be 16 bytes?

Also, did you mean bytes? `char` is a different beast outside of C.

```suggestion
    #[error("{0} hash must be {1} bytes in length")]
    InvalidHashLength(&'static str, usize),
```

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -81,18 +84,18 @@ pub enum FuzzyHash {
...
 
 impl TryFrom<&str> for ImageFuzzyHash {
-    type Error = &'static str;
+    type Error = Error;
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
         if value.len() != 16 {
-            return Err("Image fuzzy hash must be 16 characters in length");
+            return Err(Error::InvalidHashLength("ImageFuzzyHash"));
         }
```
shutton, line 92: If you want to avoid repeating yourself...
```suggestion
            return Err(Error::InvalidHashLength(std::any::type_name::<Self>()));
```

libclamav_rust/src/fmap.rs
```diff
@@ -0,0 +1,113 @@
...
+            debug!(
+                "need_off at {:?} len {:?} for fmap size {:?} returned NULL",
+                at, len, fmap_size
+            );
+            return Err(Error::NotContained(at, len, fmap_size));
+        }
+
+        let slice: &[u8] = unsafe { std::slice::from_raw_parts(ptr, len) };
+
+        debug!(
```
shutton, line 88: This might need to drop to `trace!`.  I'm assuming this will get very noisy.

libclamav_rust/src/ctx.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    let recursion_stack_size = unsafe { *ctx }.recursion_stack_size as usize;
+    let recursion_level = unsafe { *ctx }.recursion_level as usize;
+
+    let recursion_stack =
+        unsafe { slice::from_raw_parts((*ctx).recursion_stack, recursion_stack_size) };
+
+    let current_level = recursion_stack[recursion_level];
+
+    current_level.fmap.try_into().map_err(|_| Error::BadMap)
```
shutton, line 99: ...no cookie.

shutton (changes requested):

libclamav_rust/src/fuzzy_hash.rs
micahsnyder: That message is a copy-paste but yeah I may as well improve while I'm touching it. bytes is better, and customizing the hash length is better.

micahsnyder: Actually, characters makes sense in this context because it's talking about a hex string -> bytes conversion.  Each char of a hex string is just a nibble.

libclamav_rust/Cargo.toml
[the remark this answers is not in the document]
micahsnyder: Done: https://github.com/Cisco-Talos/onenote.rs/tree/CLAM-2329-new-from-slice
Also changed in the Cargo.toml with last push

libclamav_rust/src/onenote.rs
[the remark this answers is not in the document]
micahsnyder: Fixed as per our discussion, with:


```diff
 -            let data_length_bytes = &remaining[16..20];
-            let data_length = ((data_length_bytes[3] as u64) << 24)
-                | ((data_length_bytes[2] as u64) << 16)
-                | ((data_length_bytes[1] as u64) << 8)
-                | (data_length_bytes[0] as u64);
+            let data_length =  if let Some(x) = remaining.get(16..20) {
+                u32::from_le_bytes(x.try_into().unwrap()) as u64
+            } else {
+                return None;
+            };

-            let data: &[u8] = &remaining
-                [SIZE_OF_FILE_DATA_HEADER..SIZE_OF_FILE_DATA_HEADER + data_length as usize];
+            let data: &[u8] = remaining.get(SIZE_OF_FILE_DATA_HEADER..SIZE_OF_FILE_DATA_HEADER + data_length as usize)?; 
```

You also mentioned the `bytes` crate, which could have been an interesting option.  But I'll pass for now.

libclamav_rust/src/fmap.rs
micahsnyder: This could indeed get noisy.  But our logging implementation doesn't implement `trace`-level, as that level is also not an option for `clamscan` or `clamd` logging.  Perhaps I should simply remove this.

libclamav_rust/src/fuzzy_hash.rs
shutton: ah...

shutton (approved):

libclamav_rust/src/ctx.rs
micahsnyder: Okay now 🍪 ?

libclamav_rust/src/fmap.rs
micahsnyder: Removed.

libclamav/scanners.c
```diff
@@ -4591,6 +4591,11 @@ cl_error_t cli_magic_scan(cli_ctx *ctx, cli_file_t type)
                 ret = cli_scanegg(ctx);
             break;
 
+        case CL_TYPE_ONENOTE:
+            if (SCAN_PARSE_ARCHIVE && (DCONF_ARCH & ARCH_CONF_EGG))
```
micahsnyder, line 4595: I found an issue. This should be ARCH_CONF_ONENOTE (which doesn't exist, have to edit dconf.h to add it).

micahsnyder: Fixed.

ragusaa (approved):

libclamav_rust/Cargo.toml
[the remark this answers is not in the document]
micahsnyder: No movement on https://github.com/msiemens/onenote.rs/pull/13

We ended up using a fork after all, but here: https://github.com/Cisco-Talos/onenote.rs/tree/CLAM-2329-new-from-slice
https://github.com/Cisco-Talos/clamav/blob/main/libclamav_rust/Cargo.toml#L23

After resolving some issues with `cargo vendor`, using a git repo as the source seems to work o.k.

PR #1051, opened 2023-10-11
Fix link error with LLD 17

yasuhirokimura:
Developers of FreeBSD base system are currently working to upgrade its LLVM/Clang/LLDB/LLD to 17. As a part of it they tried building all ports in FreeBSD ports collections to check if build of them succeeds with LLVM/Clang/LLD 17. As a result there are some ports that fail to be built with it and unfortunately `security/clamav` is one of them. The build of it fails with link error as following.

```
  ld: error: version script assignment of 'CLAMAV_PRIVATE' to symbol 'cli_cvdunpack' failed: symbol not defined
  ld: error: version script assignment of 'CLAMAV_PRIVATE' to symbol 'cli_dbgmsg_internal' failed: symbol not defined
  ld: error: version script assignment of 'CLAMAV_PRIVATE' to symbol 'init_domainlist' failed: symbol not defined
  ld: error: version script assignment of 'CLAMAV_PRIVATE' to symbol 'init_whitelist' failed: symbol not defined
  ld: error: version script assignment of 'CLAMAV_PRIVATE' to symbol 'cli_parse_add' failed: symbol not defined
  ld: error: version script assignment of 'CLAMAV_PRIVATE' to symbol 'cli_bytecode_context_clear' failed: symbol not defined
  cc: error: linker command failed with exit code 1 (use -v to see invocation)
```

According to the investigation of ClamAV's source code, `cli_cvdunpack` is a static function so it isn't visible to external consumers. And other mentioned symbols aren't found anywhere. So fix link error by removing all of them from linker version script.

### DIFF
--- a/libclamav/libclamav.map
+++ b/libclamav/libclamav.map
@@ -95,7 +95,6 @@ CLAMAV_PRIVATE {
     cli_strlcat;
     cli_strlcpy;
     cli_strntoul;
-    cli_cvdunpack;
     cli_regcomp;
     cli_regexec;
     cli_regfree;
@@ -112,7 +111,6 @@ CLAMAV_PRIVATE {
     cli_ole2_extract;
     cli_errmsg;
     cli_debug_flag;
-    cli_dbgmsg_internal;
     cli_vba_readdir;
     cli_vba_inflate;
     cli_ppt_vba_read;
@@ -161,8 +159,6 @@ CLAMAV_PRIVATE {
     regex_list_match;
     cli_hashset_destroy;
     phishing_init;
-    init_domainlist;
-    init_whitelist;
     phishing_done;
     blobCreate;
     blobAddData;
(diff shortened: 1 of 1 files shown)


micahsnyder:
Thanks @yasuhirokimura these changes make sense to me. Will run through the test pipelines and merge if all goes well. 

I imagine we should backport this change to other versions as well.

micahsnyder:
The code change makes sense to me.  Functions that either don't exist (or are static and happen to not be used) should not be exported in the libclamav.map file.

We can backport this as far as 1.0.x. And probably should, because 1.0.x is LTS and will be used for a long a couple more years.

We could backport to 0.103.x as well if we remove the change to `cli_cvdunpack` because that function is used outside libclamav in 0.103.  However, I don't think it's worth the effort.  0.103 has less than 1 year of life left and systems running clang-17 should be able to use the newer clamav.

I tried testing with clang-17/clang++-17/lld-17 on ubuntu:22.04 to verify the failure and fix.  I did not see any failure, so I'm unsure why it's failing on FreeBSD.  But the change still makes sense so I'm not going to fuss about it.

micahsnyder (approved):

PR #1057, opened 2023-10-13
Windows: update DLL names for openssl and pthreads to support new versions

micahsnyder:
Goal is primarily to switch to openssl 3.x.


micahsnyder:
Build worked and runs ok.  Manually checked the zips and installer to be sure, though automation tested the zips as well.

PR #1061, opened 2023-10-18
Add Haiku (BeOS-like) support

Luca1991:
Hello,
recently I ported ClamAV (and all its modules except MILTER) to Haiku (a BeOS-like operating system).
I tried to do things in the cleanest way possible and to keeps patches at a minimum, in the hope to upstream these changes.
All three proposed changes are wrapped in if/elseif statements to avoid the possibility of introducing regressions.

I really hope that this PR will be accepted, so future ClamAV versions can be compiled without patches.

Thank you very much,
Luca

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -612,7 +612,7 @@ endforeach()
 
 # Always use '-fPIC'/'-fPIE' option if OS is not Haiku.
 if(NOT HAIKU)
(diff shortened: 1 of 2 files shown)


micahsnyder (changes requested):
Outside of this super minor change it looks good to me, testing all worked fine.

micahsnyder (approved):

micahsnyder:
Thanks for the PR @Luca1991 👍

PR #1079, opened 2023-11-03
Fix PDF decryption issue for some empty password files

micahsnyder:
Some PDF's with an empty password can't be decrypted. Investigation found that the problem is a strlen check to prevent an overflow rather than passing down the actual length of the allocated field.

Specifically, the UE buffer may have NULL values in it, so a strlen check will claim the field is shorter than it is and then later checks fail because the length is the wrong size.

While at it, I improved code comments on the function reading dictionary key-value strings and switched a flag use a bool rather than an int.


ragusaa:
Code looks fine.  Verified more files were extracted with the PR and with MAIN.

Issue #1082, opened 2023-11-10
Maybe Bug: LibClamAV Warning: file_bytes is not valid unicode: invalid utf-8 sequence of 1 bytes from index ...


snailcatcher:
Describe the bug
----------------

Hi there,
I'm using clamAV in docker (clamav/clamav:stable_base) in combination with a asp.net 7 app using [nClam](https://github.com/tekmaven/nClam). Most of the time everything is working fine and as i except it to do. But sometimes I get strange encoding issues. The problem is i just get them in our production environment and I am not able to reproduce this behavior in any way. 

How to reproduce the problem
----------------------------

Unfortunately I am not able to reproduce the issue. 

```
/ # clamconf -n
Checking configuration files in /etc/clamav

Config file: clamd.conf
-----------------------
LogFile = "/var/log/clamav/clamd.log"
LogTime = "yes"
PidFile = "/tmp/clamd.pid"
LocalSocket = "/tmp/clamd.sock"
TCPSocket = "3310"
User = "clamav"

Config file: freshclam.conf
---------------------------
PidFile = "/tmp/freshclam.pid"
UpdateLogFile = "/var/log/clamav/freshclam.log"
DatabaseMirror = "database.clamav.net"

Config file: clamav-milter.conf
-------------------------------
LogFile = "/var/log/clamav/milter.log"
LogTime = "yes"
PidFile = "/tmp/clamav-milter.pid"
User = "clamav"
ClamdSocket = "unix:/tmp/clamd.sock", "unix:/tmp/clamd.sock", "unix:/tmp/clamd.sock", "unix:/tmp/clamd.sock", "unix:/tmp/clamd.sock", "unix:/tmp/clamd.sock"
MilterSocket = "inet:7357"

Software settings
-----------------
Version: 1.2.1
Optional features supported: MEMPOOL AUTOIT_EA06 BZIP2 LIBXML2 PCRE2 ICONV JSON RAR 

Database information
--------------------
Database directory: /var/lib/clamav
bytecode.cvd: version 334, sigs: 91, built on Wed Feb 22 21:33:21 2023
daily.cld: version 27088, sigs: 2046165, built on Thu Nov  9 08:39:37 2023
main.cvd: version 62, sigs: 6647427, built on Thu Sep 16 12:32:42 2021
Total number of signatures: 8693683

Platform information
--------------------
[...]

PR #1084, opened 2023-11-14
Clam 2484: Fix warning when scanning some HTML files

micahsnyder:
HTML files with <style> blocks containing non-utf8 sequences are causing warnings when processing them to extract base64 encoded images.

To resolve this, we can use the to_string_lossy() method that may allocate and sanitize a copy of the content if the non-utf8 characters are encountered.

Resolves: https://github.com/Cisco-Talos/clamav/issues/1082


shutton (approved):
Much cleaner

PR #1090, opened 2023-11-17
typos (no functional changes)

RainRat:
typos only (no functional changes)


micahsnyder (approved):
All looks good.  I would note the unrar and libmspack code is third party and the changes there will get overwritten next time we update those sources from their upstream projects.  But I have no issue merging this as-is.  Not worth the effort to revert those specific files first.

PR #1091, opened 2023-11-17
Update regex_suffix.c

RainRat:
cli_regex2suffix incorrect memset use


micahsnyder (approved):
Looks good, thanks for the fix!

Issue #1087, opened 2023-11-16
clamav\libclamav\libmspack.c -> mspack_fmap_open potential null pointer usage


RainRat:
I notice that mspack_fmap_open contains two

`memset(mspack_handle, 0, (sizeof(*mspack_handle)));`

One of which is before the check for a null pointer.

Potential fix:

Remove the memset before the check for null pointer.

PR #1099, opened 2023-11-24
Update libmspack.c

RainRat:
Remove redundant memset. See https://github.com/Cisco-Talos/clamav/issues/1087


micahsnyder (approved):
This change makes sense to me. Thanks for helping cleaning it up.

PR #1116, opened 2023-12-15
Docs: add ARM64 cross-compile instructions for Windows and Linux

micahsnyder:


INSTALL-cross-linux-arm64.md
```diff
@@ -0,0 +1,228 @@
...
+sudo apt install -y g++-aarch64-linux-gnu
+rustup target add aarch64-unknown-linux-gnu
+```
+
+## Install build dependencies
+
+If you have a sysroot for your `aarch64-unknown-linux-gnu` target platform with the required dependencies installed, skip this step. Else, do these things to install arm64 (aarch64) versions of the ClamAV library dependencies on the local host.
+
+```bash
+sudo dpkg --add-architecture arm64
```
ragusaa, line 22: suggest copying instructions from that link, in case that page goes away.

ragusaa: Also, maybe a note about changing 'focal' to 'jammy'.

ragusaa:
I followed the instructions for not having a sysroot, and am getting link errors for stuff that's in the curses library.  
/usr/lib/gcc-cross/aarch64-linux-gnu/11/../../../../aarch64-linux-gnu/bin/ld: CMakeFiles/clamdtop.dir/clamdtop.c.o: in function `resize':
clamdtop.c:(.text+0xc0): undefined reference to `stdscr'

micahsnyder:
@ragusaa Ah right... On some distros they split up ncurses into libncurses and libtinfo.  `stdscr` is one they moved to the other library: https://stackoverflow.com/a/11269810

I think you can change the `CURSES_LIBRARY` variable in that `CMAKE_TOOLCHAIN_ARM64.cmake` to have both, like this:
```cmake
set(CURSES_LIBRARY "/usr/lib/aarch64-linux-gnu/libncurses.a;/usr/lib/aarch64-linux-gnu/libtinfo.a")
```

You should not need to separately `sudo apt-get install -y libtinfo:arm64`. I think the ncurses package should have brought it along: https://packages.ubuntu.com/jammy/libncurses-dev

INSTALL-cross-linux-arm64.md
```diff
@@ -0,0 +1,228 @@
...
+set(OPENSSL_SSL_LIBRARY     "/usr/lib/aarch64-linux-gnu/libssl.so")
+
+set(LIBXML2_INCLUDE_DIR     "/usr/include/libxml2")
+set(LIBXML2_LIBRARY         "/usr/lib/aarch64-linux-gnu/libxml2.so")
+
+set(PCRE2_INCLUDE_DIR       "/usr/include/")
+set(PCRE2_LIBRARY           "/usr/lib/aarch64-linux-gnu/libpcre2-8.so")
+
+set(CURSES_INCLUDE_DIR      "/usr/include/")
+set(CURSES_LIBRARY          "/usr/lib/aarch64-linux-gnu/libncurses.a")
```
ragusaa, line 180: I received link errors for stuff in 'ncurses'.  Changing to the shared object resolved them.  I believe we also need 'tinfo' for static linking.

micahsnyder: Ha looks like you found the same thing about libtinfo that I commented on a minute prior.

micahsnyder: Good to know that using the `.so` also resolves it.

ragusaa (changes requested):
see comments

ragusaa (approved):

Issue #1013, opened 2023-08-29
Clamav 1.2.0 - clamconf result question


teoberi:
Describe the bug
----------------
When I run the command
`./clamconf`
the result contains:

> Checking configuration files in /opt/clamav/etc
> Config file: clamd.conf
> ...
> MaxScanTime disabled
> !!! MaxScanSize: UNKNOWN INTERNAL TYPE !!!
> MaxFileSize = "104857600"
> ...

I specify that I have not modified the respective option in clamd.conf, it is default.

teoberi:
The message appears in the **clamconf.c** file twice, lines 116 and 187, as a default option for switch instructions.
Isn't the case missing:
`case CLOPT_TYPE_SIZE64:`

teoberi:
Maybe adding the missing case to the clamconf.c file would be the solution to solving the problem.
After line 103
`                case CLOPT_TYPE_SIZE:`
will be added
`                case CLOPT_TYPE_SIZE64:`
After line 171
`                case CLOPT_TYPE_SIZE:`
will be added
`                case CLOPT_TYPE_SIZE64:`
The printf message in line 172 will also be modified.

rsundriyal:
Hi @teoberi, Thanks for sharing the issue. We have created a ticket to track the issue. Will update you on its fix.

dohoangkhiem:
Hello, 

Any update on this? We have the config "MaxScanSize 500M" in our clamd.conf and running "clamconf -n" also gave us the error 

```
!!! MaxScanSize: UNKNOWN INTERNAL TYPE !!!
MaxFileSize = "209715200"
```

That makes us unsure that our config is really appied and effective, or due to that error clamd will revert to the default MaxScanSize (100M)?

teoberi:
Have you tried the changes [here](https://github.com/Cisco-Talos/clamav/issues/1013#issuecomment-1700623971)?
It works for me.

PR #1121, opened 2023-12-20
Clamav - clamconf result question

teoberi:
Fix issue https://github.com/Cisco-Talos/clamav/issues/1013 - Clamav 1.2.0 - clamconf result question


micahsnyder (approved):
Thanks @teoberi this looks great to me.

micahsnyder:
I confirmed this output before the fix:
```
Config file: clamd.conf
-----------------------
TCPSocket = "3310"
!!! MaxScanSize: UNKNOWN INTERNAL TYPE !!!
```
and this output after:
```
Config file: clamd.conf
-----------------------
TCPSocket = "3310"
MaxScanSize = "5368709120"
```

PR #1135, opened 2024-01-09
Merge clamav-sys

shutton:
Bare merge of `clamav-sys` into `clamav`.  All commits that were imported from `clamav-sys` have been prefaced with `clamav-sys:` to reduce confusion about their origin.

### DIFF
--- a/clamav-sys/.gitignore
+++ b/clamav-sys/.gitignore
@@ -1,0 +1,2 @@
+/target
+Cargo.lock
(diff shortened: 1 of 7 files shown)


micahsnyder:
this strategy for bringing in clamav-sys in a way that it will stay up to date with clamav itself makes sense to me. 
the commit message prefix "clamav-sys:" also seems good, and while it is a lot of commits to bring in, I like how it maintains history of the changes and gives credit to the contributors. 

👍

val-ms:

PR #1136, opened 2024-01-09
Centralize bindgen

shutton:
Centralize all use of bindgen (which exposes libclamav symbols to Rust code) into the clamav-sys module.  libclamav_rust is no longer responsible for exporting internal symbols, and is focused only on exporting C-compatible symbols referencing code implemented in Rust.


val-ms:

PR #1141, opened 2024-01-15
Clam 2197 Fix PDF metadata decryption issues + Add support for checking empty owner password

micahsnyder:
1. Fix PDF metadata decryption issues.
    The encrypted metadata may be stored in a `<>`-block containing hex bytes (not raw binary).
    Strip off the <> and decode the hex to binary before attempting to decrypt it.

2. PDF: Add support for checking empty owner password.
    Specifically for algorithm 6 (/R 6). I have not attempted the same for older algorithms.
    Use the O and OE strings to test if an empty owner password will decrypt the file.
    
    The primary difference with checking the user password is the U-string is also concatenated when making the hashes to verify the key.

### DIFF
--- a/libclamav/pdf.c
+++ b/libclamav/pdf.c
@@ -2832,7 +2832,19 @@ static void dbg_printhex(const char *msg, const char *hex, unsigned len)
     }
 }
 
-static void compute_hash_r6(const char *password, size_t pwlen, const unsigned char salt[16], unsigned char hash[32])
+/**
+ * @brief Compute the hash of the password concatenated with the validation salt and (for owner-password checks) the U string.
+ *
+ * Some details and comments for how to compute this hash comes from the PyPDF project:
+ * https://github.com/py-pdf/pypdf/blob/3.17.4/pypdf/_encryption.py#L568
+ *
+ * @param password  The password to hash.
+ * @param pwlen     The length of the password.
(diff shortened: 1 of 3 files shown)


libclamav/pdf.c
```diff
@@ -2879,184 +2903,325 @@ static void compute_hash_r6(const char *password, size_t pwlen, const unsigned c
...
+                (const unsigned char *)(O + 40), // owner key salt
+                hash,
+                U);
+
+            if (OE_len != 32) {
+                cli_dbgmsg("check_owner_password: OE length is not 32: %zu\n", OE_len);
+                noisy_warnmsg("check_owner_password: OE length is not 32: %zu\n", OE_len);
             } else {
                 pdf->keylen = 32;
                 pdf->key    = cli_malloc(pdf->keylen);
```
ragusaa, line 2979: probably don't need cli_malloc here, since keylen is fixed.

micahsnyder: I forgot to reply to this one.  There are 3 other locations where the pdf->key is allocated.  For one of those,  the size is not 32.  I confirmed that for that one it's always less than 32, so... we could probably make it a part of the pdf buffer and not malloc it here.  But I'm okay with keeping it as-is for now.

PR #1145, opened 2024-01-18
Refactor and extend FindRust.cmake

shutton:
Export additional build context to Rust-based builds, and eliminate some of the duplicated logic when bulding a macOS univeral binary.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -189,7 +189,11 @@ function(cargo_vendor)
     # This will allow us to package vendored dependencies in source tarballs
     # for online builds when we run `cpack --config CPackSourceConfig.cmake`
     message(STATUS "Running `cargo vendor` to collect dependencies for ${ARGS_TARGET}. This may take a while if the local crates.io index needs to be updated ...")
+
     make_directory(${CMAKE_SOURCE_DIR}/.cargo)
+
+    get_cargo_local_dir_env(_cargo_local_dir_env)
+
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E env "CARGO_TARGET_DIR=${ARGS_BINARY_DIRECTORY}" ${cargo_EXECUTABLE} vendor "${CMAKE_SOURCE_DIR}/.cargo/vendor"
         WORKING_DIRECTORY "${ARGS_SOURCE_DIRECTORY}"
@@ -221,6 +225,26 @@ directory = \".cargo/vendor\"
     endif()
 endfunction()
 
+# Export the project name (including upstream project) and maintainer mode
+# status for use by build scripts
+function(get_cargo_extra_env OUTPUT)
+    set(_result "PROJECT_NAME=${PROJECT_NAME}")
+    list(APPEND _result "CMAKE_PROJECT_NAME=${CMAKE_PROJECT_NAME}")
+    list(APPEND _result "MAINTAINER_MODE=${MAINTAINER_MODE}")
+    # Propagate any additional Rust compiler flags
+    list(APPEND _result "RUSTFLAGS=${RUSTFLAGS}")
+    set("${OUTPUT}" "${_result}" PARENT_SCOPE)
+endfunction()
+
+# From within another function, set environment variables pointing to local
+# directories as specified by the arguments to that function
+function(get_cargo_local_dir_env OUTPUT)
+    # Specify the destination directory for build products
+    set(_env "CARGO_TARGET_DIR=${ARGS_BINARY_DIRECTORY}")
+    list(APPEND _env "CARGO_INCLUDE_DIRECTORIES=\"${ARGS_INCLUDE_DIRECTORIES}\"")
+    set(${OUTPUT} "${_env}" PARENT_SCOPE)
+endfunction()
+
 function(add_rust_executable)
     set(options)
     set(oneValueArgs TARGET SOURCE_DIRECTORY BINARY_DIRECTORY)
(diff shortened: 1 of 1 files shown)


micahsnyder:
@shutton looks like your PR branch was started from an older commit and conflicts with the FindRust.cmake from the latest in `main`.  Perhaps your fork is out of sync.  You'll have to rebase with the upstream main and resolve the conflicts.

micahsnyder:
Ha!  You fixed it a literal second after I wrote that. Nice.

cmake/FindRust.cmake
```diff
@@ -142,6 +142,10 @@ if(NOT DEFINED CARGO_HOME)
     endif()
 endif()
 
+#
+# Environment for most invocations of `cargo` 
+#
```
micahsnyder, line 147: I'm not really sure what this code block means to say

cmake/FindRust.cmake
```diff
@@ -190,6 +194,7 @@ function(cargo_vendor)
     # for online builds when we run `cpack --config CPackSourceConfig.cmake`
     message(STATUS "Running `cargo vendor` to collect dependencies for ${ARGS_TARGET}. This may take a while if the local crates.io index needs to be updated ...")
     make_directory(${CMAKE_SOURCE_DIR}/.cargo)
+    GetCargoLocalDirEnv(_cargo_local_dir_env)
```
micahsnyder, line 197: Two things:
1. It would be nice to add some new lines between these 4 function calls.  
2. Do we need to call `GetCargoExtraEnv(_cargo_extra_env)` as well?  Probably not, but I thought I should ask.

cmake/FindRust.cmake
```diff
@@ -221,6 +226,24 @@ directory = \".cargo/vendor\"
...
+function(GetCargoExtraEnv result)
+    set(_result "PROJECT_NAME=${PROJECT_NAME}")
+    list(APPEND _result "CMAKE_PROJECT_NAME=${CMAKE_PROJECT_NAME}")
+    list(APPEND _result "MAINTAINER_MODE=${MAINTAINER_MODE}")
+    # Propagate any additional Rust compiler flags
+    list(APPEND _result "RUSTFLAGS=${RUSTFLAGS}")
+    set("${result}" "${_result}" PARENT_SCOPE)
+endfunction()
+
+function(GetCargoLocalDirEnv result)
```
micahsnyder, line 240: Three things:
1. Naming convention for CMake functions seems to be snake_case. Can we switch this from CamelCase?

2. I wonder if this should be defined before the `cargo_vendor` function as it is first used there. 

3. This function could use a comment block

cmake/FindRust.cmake
```diff
@@ -190,6 +194,7 @@ function(cargo_vendor)
     # for online builds when we run `cpack --config CPackSourceConfig.cmake`
     message(STATUS "Running `cargo vendor` to collect dependencies for ${ARGS_TARGET}. This may take a while if the local crates.io index needs to be updated ...")
     make_directory(${CMAKE_SOURCE_DIR}/.cargo)
+    GetCargoLocalDirEnv(_cargo_local_dir_env)
     execute_process(
         COMMAND ${CMAKE_COMMAND} -E env "CARGO_TARGET_DIR=${ARGS_BINARY_DIRECTORY}" ${cargo_EXECUTABLE} vendor "${CMAKE_SOURCE_DIR}/.cargo/vendor"
```
micahsnyder, line 199: Do we need to add `"${_cargo_local_dir_env}"` here?  Or is `GetCargoExtraEnv` not required for this function?

cmake/FindRust.cmake
```diff
@@ -221,6 +226,24 @@ directory = \".cargo/vendor\"
     endif()
 endfunction()
 
+# Export the project name (including upstream project) and maintainer mode
+# status for use by build scripts
+function(GetCargoExtraEnv result)
```
micahsnyder, line 231: Naming convention for CMake functions seems to be snake_case. Can we switch this from CamelCase?

cmake/FindRust.cmake
```diff
@@ -288,39 +315,43 @@ function(add_rust_library)
...
+
+    if(RUST_COMPILER_TARGET STREQUAL "universal-apple-darwin")
+        foreach(arch IN LISTS CMAKE_OSX_ARCHITECTURES)
+            message("Making custom command for OSX arch ${arch}")
+            if(arch STREQUAL "arm64")
+                # Convert to rustc's designation
+                set(arch "aarch64")
+            endif()
+            set(this_output "${ARGS_BINARY_DIRECTORY}/${arch}-apple-darwin/${CARGO_BUILD_TYPE}/lib${ARGS_TARGET}.a")
+            list(APPEND lipo_inputs "${this_output}")
```
micahsnyder, line 329: This is neat.  I'd suggest a little whitespace plus a comment for reading clarity.

```suggestion

            # Record path of the compiled library for lipo to assemble the universal binary, later. 
            set(this_output "${ARGS_BINARY_DIRECTORY}/${arch}-apple-darwin/${CARGO_BUILD_TYPE}/lib${ARGS_TARGET}.a")
            list(APPEND lipo_inputs "${this_output}")

```

cmake/FindRust.cmake
```diff
@@ -288,39 +315,43 @@ function(add_rust_library)
...
+            add_custom_command(
+                OUTPUT "${this_output}"
+                COMMAND ${CMAKE_COMMAND} -E env "CARGO_CMD=build" "${_cargo_extra_env}" "${_cargo_local_dir_env}" ${cargo_EXECUTABLE} ${MY_CARGO_ARGS} --target=${arch}-apple-darwin
+                COMMAND_EXPAND_LISTS
+                WORKING_DIRECTORY "${ARGS_SOURCE_DIRECTORY}"
+                DEPENDS ${LIB_SOURCES}
+                COMMENT "Building ${ARGS_TARGET} for ${arch} in ${ARGS_BINARY_DIRECTORY} with:  ${cargo_EXECUTABLE} ${MY_CARGO_ARGS_STRING}")
+        endforeach()
+        # `lipo` will provide the Universal binary
```
micahsnyder, line 338:
```suggestion

        # `lipo` will provide the Universal binary
```

micahsnyder (changes requested):

cmake/FindRust.cmake
shutton: It's a comment that got detached from its code.

cmake/FindRust.cmake
shutton: Absolutely.  I think I picked this style up from another file I was using as reference.

cmake/FindRust.cmake
shutton: 1. Changed.
2. Doesn't seem to matter thanks to lazy evaluation, but there could be more tidying done
3. Added.

cmake/FindRust.cmake
shutton: 1. Done.
2. Tricky.  `execute_process()` doesn't support the `COMMAND_EXPAND_LISTS` directive, which is necessary to make the environment supplementation work properly. You also didn't reference the other variable here (the include dir), so I just left this one as-is.

cmake/FindRust.cmake
shutton: See above (execute_process doesn't support lists)

micahsnyder (approved):

PR #1149, opened 2024-01-22
Clam 2507 rc2 prep

micahsnyder:
Set the 1.3.0-rc2 version suffix
Rust: Update pinned dependency versions in Cargo.lock file
News: Update release notes for 1.3.0-rc2


shutton (approved):
LGTM

ragusaa (approved):

micahsnyder:
I rebased and also dropped the commit that updates the Cargo.lock file. The `home` package recently raised their MSRV (minimum supported Rust version) to 1.70, which is newer than some of our build systems.  I don't want to go upgrade all of those for a last second update.  We can keep the lock file as is for this release and then raise our MSRV during 1.4 dev.

PR #1150, opened 2024-01-23
Fix clamonacc infinite loop when a file has been deleted before scan

gsuehiro:
The `err` variable is used as an escape condition for the while loop. This is a pointer. 
Shouldn't `*err` be used?


micahsnyder (approved):
I tried to reproduce the issue and I wasn't able to trigger the infinite loop. 
But your fix makes sense to me. Very nice catch.  Thanks @gsuehiro.

PR #1154, opened 2024-01-24
Refactor bindgen

shutton:
This moves all of the bindgen functionality (exporting libclamav APIs for usage by Rust) into `clamav-sys`, and isolates the cbindgen functionality (exporting Rust APIs for usage by C) to the `libclamav_rust` crate.


val-ms:

PR #1165, opened 2024-02-05
Draft: ClamD: Add new zOPTSCAN command w. scan options like ALLMATCH

micahsnyder:
Added a new ClamD command which can pass options along with the actual scan command.  Until now, there has been no way to pass scan options from clamdscan to clamd, which either:
1. prevents the use of scan options 
2. means scan options must be set for all scans through clamd.conf instead of setting them per-scan with clamdscan
3. requires the use of a custom scan command for each option, like the old `ALLMATCHSCAN` command.

What inspired this work is the fact that `ALLMATCHSCAN` only works for filepaths and does not work in combination with INSTREAM, or the much preferred FDPASS scan commands.  Scanning filepaths is only valid if `clamd` can open the file.  If `clamd` is started as a system service, it should have restricted access, and so read-only access to the file be granted by the scanning client.  

In short, this enables the use of allmatch mode with fd-passing via `clamdscan --fdpass --allmatch`, which is certainly valuable and until now has not been possible.

In the future, I would like to fully replace the `clamd` socket protocol with something easier to extend (i.e. REST, gRPC, etc.). That will most certainly be a breaking change most people.  So until then, this change should enable us to pass ALLMATCH and other scan options from the scan clients to clamd. 

For now, `zOPTSCAN` scan options are just `ALLMATCH`.

And current scan commands are:
- `CONTSCAN`
- `FILDES`
- `INSTREAM`
- `MULTISCAN`

For example, you can sendstuff like this:
```
zOPTSCAN CONTSCAN /some/file/path\0
```
```
zOPTSCAN ALLMATCH CONTSCAN /some/file/path\0
```
```
zOPTSCAN MULTISCAN /some/file/path\0
```
```
zOPTSCAN ALLMATCH MULTISCAN /some/file/path\0
```
```
zOPTSCAN INSTREAM \0 <4-byte length><length-bytes data chunk> <4-byte length><length-bytes data chunk> <4-byte zero>
```
```
zOPTSCAN ALLMATCH INSTREAM \0 <4-byte length><length-bytes data chunk> <4-byte length><length-bytes data chunk> <4-byte zero>
```
etc. (I'm not sure how to represent the fdpass structure here in text 😅. 

Note that a space is presently required between the final option before the path or before the \0 for INSTREAM and FDPASS where the subsequent bytes appear are sent in the next message.

/This change SHOULD be backwards compatible with older scan clients (or else we should not merge it). 
/This presently fails a clamd test with a crash because of an invalid free of the `conn->options` pointer.  I am trying to track it down and feel the github diff viewer and additional testing may help.


ragusaa:
When I run 
./bin/clamdscan --allmatch --fdpass ~/PR-1165/samples/ --multiscan

I get errors.  Not the same every time, but things like 
/home/aragusa/PR-1165/samples/1393668558: BC.Unix.Exploit.Agent-1393645-1 FOUND
ERROR: Bogus session id from clamd
ERROR: Clamd closed the connection before scanning all files.

another time
ERROR: FD send failed: Broken pipe
LibClamAV Error: File tree walk aborted.

micahsnyder:
Good find, Andy. `--multiscan`) (MULTISCAN) and `--fdpass` (FILEDES) are indeed incompatible in this PR, and I believe they're also incompatible before this PR.  If we're ever going to fix it, we should fix it with the addition of this OPTSCAN command.  I will close this PR to work on it more before opening a new PR.

One solution may be to change MULTISCAN from a scan type to a a scan option (like ALLMATCH).
Another would be to make it so scanning file paths and file descriptors are always multithreaded and just alter clamdscan behavior to wait for scan results before sending the next if `--multiscan` is not used.  I like this second way better, since it means less duplicate code and less complexity.

I believe the current logic for MULTISCAN relies on sending a directory path and then clamd would do a multithreaded scan of the directory contents.  That isn't a valid design for a multithreaded scan passing file descriptors.  We would have to make it so clamdscan recurses the directory tree, sending all of the file descriptors to be scanned - presumably before listening for the results -- or using another thread of its own to send them while simultaneously awaiting results. 

I probably should've also taken the opportunity to rename the OPTSCAN commands describing what the subsequent data:
- "CONTSCAN" to "FILEPATH" or just "PATH"
- "FILEDES" to "FILEDESCRIPTOR" or just "DESCRIPTOR"
- "INSTREAM" to "FILESTREAM" or maybe "FILEDATA", "FILEDATASTREAM", "DATASTREAM", or just "STREAM"

Issue #1138, opened 2024-01-11
Possible misleading documentation for `clamdscan --fdpass`


vc-jl:
Hi there,

`man clamdscan` claims:

```
       --fdpass
              Pass the file descriptor permissions to clamd. This is useful 
              if clamd is running as a different user as it is faster than 
              streaming the file to clamd.  Only available if connected to 
              clamd via local(unix) socket.
```

This sounds as if "permissions" are sent, whatever that means. As I understand it, the actual file descriptor of the opened file is sent via socket to the daemon.

If I understand this correctly, the documentation is a bit misleading and could be more precise.

I could also provide a pull request if desired.

Thanks
Johannes

ragusaa:
Hi,

Thank you, that is an error in the documentation.  We will be happy to review a PR.

Thanks,
Andy

PR #1187, opened 2024-02-26
Docs: Clarify help text about --fdpass option

vc-jl:
My shot at a better help text for `--fdpass` as mentioned here: https://github.com/Cisco-Talos/clamav/issues/1138. I am happy to receive further suggestions on how to rephrase.

Thanks!


micahsnyder (approved):
This seems like a reasonable improvement.  We would also normally make the same change to the `clamdscan --help` output as well, though in this case I think it's a little too lengthy for `--help` output.  So I'm happy to merge this change as-is. 

Thanks @vc-jl

PR #1188, opened 2024-02-27
Stop parsing yara files when a syntax error s found.

ragusaa:
The yara parser can potentially overwrite heap buffers parsing invalid yara files. Exit on error to avoid this.  These overwrites are only observed when running with address sanitization and mpool disabled.

### DIFF
--- a/libclamav/yara_grammar.c
+++ b/libclamav/yara_grammar.c
@@ -3649,6 +3649,8 @@ yyerrlab:
       }
 # undef YYSYNTAX_ERROR
 #endif
+      /*Exit out, no reason to continue parsing, since we have already found errors.*/
+      goto yyreturn;
     }
 
 


ragusaa:
I am not sure if this is a viable solution, but it does prevent the ossfuzz errors.  This will require more testing.

micahsnyder (commented):
This fix feels wrong for a couple reasons.

1. this file is generated from the `yara_grammar.y` file, so changing this file won't have a lasting effect.  I am unsure where the corresponding code is in the `.y` file.

2. This code is derived from Yara project sources, here: https://github.com/VirusTotal/yara/blob/c009195c1e08da559600be94a78a14c2beb06d6b/libyara/grammar.c#L5325  I don't see a similar change here.  Our source is way behind theirs and I assume they've since fixed it.  We should really try to figure out how they fixed it, or else just work on upgrading to the latest sources and see if it resolves the issue.
 
My recommendation is to cancel this for now.

PR #1192, opened 2024-03-02
Clam 1796 lha lzh archive support

micahsnyder:
PR still requires some unit tests.


ragusaa:
ctest clamscan_valgrind fails.

libclamav_rust/build.rs
```diff
@@ -157,7 +159,10 @@ fn execute_bindgen() -> Result<(), &'static str> {
     }
 
     for &include_path in BINDGEN_INCLUDE_PATHS {
-        builder = builder.clang_arg(include_path);
+        builder = builder
+            .clang_arg(include_path)
+            // Hack to include 3rd party C libary headers
+            .clang_arg("-I/home/micah/.mussels/install/host-static/include");
```
ragusaa, line 165: Looks like you forgot to fix this?

ragusaa:
Code looks good.  Need to run tests.

libclamav_rust/build.rs
micahsnyder: Oh, crap. I absolutely need to remove this.  I think this was a workaround for issues with having openssl-1.1-dev installed to my system while building with openssl 3.  It's not relevant to this PR at all and should never have been committed.

micahsnyder: Fixed.

libclamav/ole2_extract.c
```diff
@@ -256,8 +256,8 @@ cli_ole2_get_property_name2(const char *name, int size)
     if ((name[0] == 0 && name[1] == 0) || size <= 0 || size > 128) {
         return NULL;
     }
-    CLI_MAX_MALLOC_OR_GOTO_DONE(newname, size * 7,
-                                cli_errmsg("OLE2 [cli_ole2_get_property_name2]: Unable to allocate memory for newname: %u\n", size * 7));
+    CLI_MAX_MALLOC_OR_GOTO_DONE(newname, size * 5,
```
ragusaa, line 259: Could you add a comment specifying the significance of the 5?

libclamav_rust/src/scanners.rs
```diff
@@ -127,3 +134,164 @@ pub unsafe extern "C" fn scan_onenote(ctx: *mut cli_ctx) -> cl_error_t {
...
+            }
+            Ok(false) => {
+                debug!("No more files in the archive.");
+                break;
+            }
+            Err(err) => {
+                // Error getting the next file.
+                // Use debug-level because may not actually be an LHA/LZH archive.
+                // LHA/LZH does not have particularly identifiable magic bytes.
+                debug!("An error occured when checking for the next file in this LHA or LZH archive: {err}");
```
ragusaa, line 290: Typo, 'occurred'

ragusaa:
I added some code to scanners.rs to dump all files extracted to disk, to verify that they are the same as what 'lha' writes.  I notice that when I scan the contents of lha.tar.gz, from https://jira-eng-sjc1.cisco.com/jira/browse/CLAM-2343, I do not get some of the files.  The first one I notice missing is the contents of alg6.lha.  Interestingly enough, if I scan only that file with clamscan, the contents are extracted.

libclamav/ole2_extract.c
micahsnyder: Good point.  I will do so or else make a #define with a descriptive name.

libclamav_rust/src/scanners.rs
micahsnyder: For the life of me I can never remember how to spell that.

ragusaa: I only noticed because I was searching for it in the code, then had to google it.

libclamav/ole2_extract.c
micahsnyder: Fixed.

libclamav_rust/src/scanners.rs
micahsnyder: Fixed.

ragusaa (approved):

micahsnyder:
I had to tweak the valgrind suppression rules a bit more, but it finally passed. I'm happy now.

PR #1198, opened 2024-03-06
CLAM-2216-ClamOnAccLoop Applied the fix that was not included in 0.103 version

Juho-J:
Applied the fix that was already done here: https://github.com/Cisco-Talos/clamav/pull/1047

This fix was originally aimed to fix the issue where a wrong configuration would lead to clamonacc process just running in a infinite loop and using 100% of the available cpu time that was allocated to the process.

Example of the incorrect configuration that triggers this bug:
´OnAccessIncludePath /var/lib/this_path_does_not_exist´

After this is changed to a correct path, clamonacc will not be stuck in the loop and will work as intended.

Sorry if I did not follow good practises when creating this PR. This is my first time creating a PR to a public project.

I did not test this as this was already accepted in the original PR.

### DIFF
--- a/clamonacc/inotif/inotif.c
+++ b/clamonacc/inotif/inotif.c
@@ -585,8 +585,8 @@ void *onas_ddd_th(void *arg)
                         }
                     }
                 }
-                pt = (struct optstruct *)pt->nextarg;
             }
+            pt = (struct optstruct *)pt->nextarg;
         }
     }
 


Juho-J:
Oh and also some background story:

Ubuntu 22 LTS release uses the version 0.103 as the latest package available. If this was not the case I would just upgrade ClamAV packages and that would fix this issue.

micahsnyder:
Thanks for the PR. 

Ubuntu 22 LTS is certainly behind with 0.103.  The 0.103 LTS release will reach end-of-life in September this year in terms of our team providing any fixes for the release: https://docs.clamav.net/faq/faq-eol.html#version-support-matrix

At this point we've only been publishing patch versions for 0.103 LTS when there is a major bug or security issue.  I don't know of any other bugs at this time that would necessitate another 0.103 patch version.  To me, this change alone isn't compelling enough to do so.  I think we will leave this PR open until then, or until 0.103 end-of-life.

If you want to upgrade to a newer release that already has the ClamOnAcc loop fix, you have two options:
1. compile clamav from source
2. use one of the `.deb` packages we provide on clamav.net/downloads.  More details [here](https://docs.clamav.net/manual/Installing.html#installing-with-an-installer).

micahsnyder (approved):

micahsnyder:
Rebased with `dev/0.103.12` to sign the commit, and fixed up the commit message.

Issue #1190, opened 2024-02-29
unclear to the user - freshclam --datadir only works with absolute paths and does not create directories recursively


TorstenEckern:
Describe the bug
----------------

The user is never notified that freshclam --datadir
- does not create directories recursively
- needs an absolute path to run properly.

How to reproduce the problem
----------------------------

I'm using the default clamav 0.103.11 apt package installed on Ubuntu 20.04 (tag: focal-20221130 -> linux/amd64) in a fresh docker container:
```
#### Build Stage
################################################################################
ARG UBUNTU_BASE_TAG=focal-20221130
FROM ubuntu:${UBUNTU_BASE_TAG} as builder
ENV DEBIAN_FRONTEND=noninteractive

# Setup base system
# -----------------
RUN apt-get update && apt-get install -y \
    clamav-freshclam \
    && rm -rf /var/lib/apt/lists/*
```
    
I run everything as root and directly after install of clamav-freshclam.
###########################
```
root@b5a47fb61fdc:/# freshclam -v --datadir=./foo/bar/clamav
Creating missing database directory: ./foo/bar/clamav
ERROR: Failed to create database directory: ./foo/bar/clamav
Manually prepare the database directory, or re-run freshclam with higher privileges.
ERROR: Initialization error!
```

Second experiment: 
#############################
```
root@b5a47fb61fdc:/# freshclam -v --datadir=./clamav
Creating missing database directory: ./clamav
Assigned ownership of database directory to user "clamav".
Thu Feb 29 11:37:25 2024 -> *Current working dir is ./clamav/
Thu Feb 29 11:37:25 2024 -> *Can't open freshclam.dat in /sandbox/packages/m3d-virus-scan/clamav
Thu Feb 29 11:37:25 2024 -> *It probably doesn't exist yet. That's ok.
Thu Feb 29 11:37:25 2024 -> *Failed to load freshclam.dat; will create a new freshclam.dat
Thu Feb 29 11:37:25 2024 -> *Creating new freshclam.dat
Thu Feb 29 11:37:25 2024 -> *Saved freshclam.dat
Thu Feb 29 11:37:25 2024 -> ClamAV update process started at Thu Feb 29 11:37:25 2024
Thu Feb 29 11:37:25 2024 -> !Can't change dir to ./clamav/
Thu Feb 29 11:37:25 2024 -> *Querying current.cvd.clamav.net
Thu Feb 29 11:37:25 2024 -> *TTL: 1197
[...]

PR #1199, opened 2024-03-06
Fix usage for --datadir option

ragusaa:
Make the usage clear to the user that the option specified by '--datadir' must be an absolute path to a directory that already exists, and is writeable by freshclam and readable by clamscan/clamd.

Resolves: https://github.com/Cisco-Talos/clamav/issues/1190

### DIFF
--- a/freshclam/freshclam.c
+++ b/freshclam/freshclam.c
@@ -191,6 +191,8 @@ static void help(void)
     printf("    --no-dns                             Force old non-DNS verification method\n");
     printf("    --checks=#n          -c #n           Number of checks per day, 1 <= n <= 50\n");
     printf("    --datadir=DIRECTORY                  Download new databases into DIRECTORY\n");
+    printf("                                         NOTE: DIRECTORY must already exist, be an absolute path, and");
+    printf("                                         be writeable by freshclam and readable by clamd/clamscan.");
     printf("    --daemon-notify[=/path/clamd.conf]   Send RELOAD command to clamd\n");
     printf("    --local-address=IP   -a IP           Bind to IP for HTTP downloads\n");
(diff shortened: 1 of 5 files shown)


freshclam/freshclam.c
```diff
@@ -191,6 +191,8 @@ static void help(void)
     printf("    --no-dns                             Force old non-DNS verification method\n");
     printf("    --checks=#n          -c #n           Number of checks per day, 1 <= n <= 50\n");
     printf("    --datadir=DIRECTORY                  Download new databases into DIRECTORY\n");
+    printf("                                         NOTE: DIRECTORY must already exist, be an absolute path, and");
+    printf("                                         be writeable by freshclam and readable by clamd/clamscan.");
```
micahsnyder, line 195: I'm good with this.  

Can you also add this to:
- the freshclam manpage:  https://github.com/Cisco-Talos/clamav/blob/main/docs/man/freshclam.1.in#L63-L64
- the freshclam.conf.sample:  https://github.com/Cisco-Talos/clamav/blob/main/etc/freshclam.conf.sample#L10-L13
- the freshclam.conf manpage:  https://github.com/Cisco-Talos/clamav/blob/main/docs/man/freshclam.conf.5.in#L64-L65

I also noticed that the comments in `freshclam.conf.sample` have more detail than in `freshclam.conf.5.in` manpage. If you can sync them (for this particular option), that would be nice.

micahsnyder (changes requested):

freshclam/freshclam.c
ragusaa: done

ragusaa: done

micahsnyder (approved):

PR #1217, opened 2024-03-26
Clam 2427 dead code

micahsnyder:
This PR removes a bunch of preprocessor checks around dead code for build configurations that are no longer supported (e.g. no bzip2, no pcre2, using pcre instead of pcre2, no libxml2, no json-c). 

This "resolves" some coverity issues found in said dead code, and should make the code easier to understand and maintain.


ragusaa:
The changes look fine, and the unit tests pass.

It would still be good to do a regression run after this is merged, since there are so many changes.

ragusaa (approved):

micahsnyder:
> The changes look fine, and the unit tests pass.
> 
> It would still be good to do a regression run after this is merged, since there are so many changes.

Yeah we'll definitely at least regression test the release candidate vs the 1.3.1 release. And it couldn't hurt to regression test sooner.

micahsnyder:
I just rebased and force-pushed.

PR #1220, opened 2024-03-29
Fix unicode conversion truncation bug

micahsnyder:
The in_iconv_u16() function resolves "alignment" issues where the length of the input string is not mod(4). The solution trims the extra bytes off the input string. If the input string is total less than 4 bytes, then those extra bytes are put in a 4-byte array and are converted. However, if the input string is longer, then those extra bytes are lost.

This fix saves the extra "unaligned" bytes in the 4-byte array and converts them afterwards so we don't accidentally lose 1 to 2 characters.


ragusaa (approved):

micahsnyder:
Just rebased with main. No changes.

PR #1233, opened 2024-04-12
Clam 2612 Fix DatabaseCustomURL CVD prune bug, and add 'valhalla' optional database

micahsnyder:
- Freshclam: fix issue DatabaseCustomURL CVD prune issue

   If using DatabaseCustomURL to download a CVD that Freshclam doesn't know about, i.e. one that is not in the hardcoded standard or optional database lists in freshclam.c, Freshclam will prune the database and then re-download it.

   This change makes it so we look for URL's with ".cvd" at the end and then take those into consideration when checking which  CVD's (or CLD's) should be pruned.

   Note that I didn't change the interface to fc_prune_database_directory(). That would have been cleaner, but would've changed the public API and I want to backport this fix.

- Add 'valhalla' to Freshclam's list of optional CVD's

### DIFF
--- a/freshclam/freshclam.c
+++ b/freshclam/freshclam.c
@@ -1029,7 +1029,7 @@ fc_error_t get_official_database_lists(
     uint32_t i;
 
     const char *hardcodedStandardDatabaseList[] = {"daily", "main", "bytecode"};
-    const char *hardcodedOptionalDatabaseList[] = {"safebrowsing", "test"};
+    const char *hardcodedOptionalDatabaseList[] = {"safebrowsing", "test", "valhalla"};
 
     if ((NULL == standardDatabases) || (NULL == nStandardDatabases) || (NULL == optionalDatabases) || (NULL == nOptionalDatabases)) {
         mprintf(LOGG_ERROR, "get_official_database_lists: Invalid arguments.\n");
--- a/freshclam/freshclam.c
+++ b/freshclam/freshclam.c
@@ -1455,6 +1455,10 @@ fc_error_t perform_database_update(
     uint32_t nUpdated      = 0;
     uint32_t nTotalUpdated = 0;
 
+    uint32_t i;
+    char **doNotPruneDatabaseList = NULL;
+    uint32_t nDoNotPruneDatabases = 0;
+
     STATBUF statbuf;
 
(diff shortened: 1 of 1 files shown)


freshclam/freshclam.c
```diff
@@ -1475,7 +1475,38 @@ fc_error_t perform_database_update(
...
+        // include the URL databases in the prune process
+        char **newDatabaseList = (char **)malloc(sizeof(char *) * (nDatabases + nUrlDatabases));
+        if (NULL == newDatabaseList) {
+            logg(LOGG_ERROR, "perform_database_update: Can't allocate memory for newDatabaseList\n");
+            status = FC_EMEM;
+            goto done;
+        }
+
+        for (i = 0; i < nDatabases; i++) {
+            newDatabaseList[i] = strdup(databaseList[i]);
```
ragusaa, line 1489: Should you check for NULL in case strdup fails?

freshclam/freshclam.c
```diff
@@ -1475,7 +1475,38 @@ fc_error_t perform_database_update(
...
+        }
+        uint32_t nNewDatabases = nDatabases;
+
+        for (i = 0; i < nUrlDatabases; i++) {
+            // only append the URL databases that end with '.cvd'
+            if (strlen(urlDatabaseList[i]) > 4 && 0 == strcasecmp(urlDatabaseList[i] + strlen(urlDatabaseList[i]) - 4, ".cvd")) {
+                const char* startOfFilename = strrchr(urlDatabaseList[i], '/') + 1;
```
ragusaa, line 1496: Where do the urls come from?  Is it possible that they would be malformed and not contain a '/'?

ragusaa:
Testing passed

freshclam/freshclam.c
micahsnyder: The URL's come from the freshclam.conf `DatabaseCustomURL` option. It would be a misconfiguration. Could happen.

ragusaa (approved):

micahsnyder:
Oops. There's a windows build issue:
> `freshclam.c.obj : error LNK2019: unresolved external symbol strndup referenced in function perform_database_update`

I'll fix it.

Issue #1249, opened 2024-04-18
clamav 1.3.1 seven tests fail


Elielsilverio:
I download and build clamav 1.3.1 with success, but on run ctest, failed

/postfix/clamav-1.3.1/build # ctest
Test project /postfix/clamav-1.3.1/build
      Start  1: libclamav
 1/11 Test  #1: libclamav ........................***Failed   13.73 sec
      Start  2: libclamav_valgrind
 2/11 Test  #2: libclamav_valgrind ...............***Failed  160.04 sec
      Start  3: libclamav_rust
 3/11 Test  #3: libclamav_rust ...................   Passed   10.24 sec
      Start  4: clamscan
 4/11 Test  #4: clamscan .........................***Failed   11.04 sec
      Start  5: clamscan_valgrind
 5/11 Test  #5: clamscan_valgrind ................***Failed  339.20 sec
      Start  6: clamd
 6/11 Test  #6: clamd ............................***Failed   26.24 sec
      Start  7: clamd_valgrind
 7/11 Test  #7: clamd_valgrind ...................***Failed   93.12 sec
      Start  8: freshclam
 8/11 Test  #8: freshclam ........................   Passed    2.50 sec
      Start  9: freshclam_valgrind
 9/11 Test  #9: freshclam_valgrind ...............***Failed   61.37 sec
      Start 10: sigtool
10/11 Test #10: sigtool ..........................   Passed    1.06 sec
      Start 11: sigtool_valgrind
11/11 Test #11: sigtool_valgrind .................   Passed   30.54 sec

36% tests passed, 7 tests failed out of 11

Total Test time (real) = 749.09 sec

The following tests FAILED:
          1 - libclamav (Failed)
          2 - libclamav_valgrind (Failed)
          4 - clamscan (Failed)
          5 - clamscan_valgrind (Failed)
          6 - clamd (Failed)
          7 - clamd_valgrind (Failed)
          9 - freshclam_valgrind (Failed)
Errors while running CTest
Output from these tests are in: /postfix/clamav-1.3.1/build/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.

/postfix/clamav-1.3.1/unit_tests/libclamav_test.py:47: AssertionError
----------------------------- Captured stdout call -----------------------------

[...]

PR #1258, opened 2024-04-29
For RAR load, check LD_LIBRARY_PATH before checking install path

micahsnyder:
ClamAV initalization's rarload() function tries to load libclamunrar_iface from the install path before checking under LD_LIBRARY_PATH.
This means the unit tests will use the wrong unrar library if testing on a system where ClamAV is already installed.
In the event there is an ABI break between versions, this will cause a bunch of tests to fail.

This commit fixes the issue by checking for libclamunrar_iface under LD_LIBRARY_PATH *first* before checking in the install lib directory.

Note in the previous version we were also checking LD_LIBRARY_PATH on Windows, which is not a thing. I removed this.

Fixes: https://github.com/Cisco-Talos/clamav/issues/1249

Also removed check for WARN_DLOPEN_FAIL define, which was not used, and mistakenly set for the unrar library build target.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -94,13 +94,13 @@ jobs:
       - uses: actions/checkout@v1
 
       - name: Install Build Tools
-        run: brew install bison flex
+        run: brew install bison flex pipx
 
       - name: Install Dependencies
-        run: brew install bzip2 check curl-openssl json-c libxml2 ncurses openssl@1.1 pcre2 zlib
+        run: brew install bzip2 check curl json-c libxml2 ncurses openssl@3 pcre2 zlib
 
       - name: Install pytest for easier to read test results
-        run: python3 -m pip install pytest
+        run: pipx install pytest
 
       - uses: lukka/get-cmake@v3.21.2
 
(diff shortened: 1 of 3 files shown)


TheRaynMan (approved):
Looks good to me, thanks for also fixing the web link for LoadLibraryA()

micahsnyder:
@TheRaynMan  could use re-review.  I tided it up further while trying to figure out why there were test failures in Jenkins on Windows. (didn't figure it out, seemed fine to me). 

Will continue to investigate.

micahsnyder:
Apparently GitHub actions on macOS fails today as well for a couple different reasons. I'm fixing them in this PR as well.

micahsnyder:
UGH finally found the correct path for brew library installs after github's change to the macOS runners: `/opt/homebrew/include` and `/opt/homebrew/lib`.

I also fixed the windows build issues which had to do with the length of the path/commands in our Jenkins and not an actual bug in our code afaict.  I solved that by relocating the build directory to be *next* to the source dir and to use a short directory name.  Previously had been longer and in a subdirectory of the source directory. 

I should be done fussing with this now.

TheRaynMan (approved):

PR #1260, opened 2024-04-30
Adding param to define test pipelines path

rsundriyal:
Adding parameter to set test pipelines path.


micahsnyder (approved):
Makes sense to me. We'll have to merge this, then relocate the test-pipelines multi-branch pipeline, and then rebase any clamav change before testing so that the paths match up.  Which is fine.

PR #1263, opened 2024-05-07
LZH: check CRC after reading file data

micahsnyder:
The checksum should be verified after reading file data, not before.


rsundriyal (approved):
Looks good.

PR #1293, opened 2024-06-27
cli_check_mydoom_log: Avoid unaligned access.

sebastianas:
fmap_need_off_once() may return an unaligned pointer. This in return leads to an unaligned access during the load of the uint32_t variables loading to failures on architectures not supporting unaligned access.

This was reported to the Debian BTS as #1073128.

[bigeasy: Commit message, reworked the patch a bit].

Link: https://bugs.debian.org/1073128

### DIFF
--- a/libclamav/special.c
+++ b/libclamav/special.c
@@ -48,7 +48,8 @@
 
 int cli_check_mydoom_log(cli_ctx *ctx)
 {
-    const uint32_t *record;
+    uint32_t record[16];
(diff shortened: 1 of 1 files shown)


sebastianas:
This just built in Debian and it also fixed a long standing issue on sparc64 which existed since the 1.0.0 version.

LocutusOfBorg:
thanks for the nice and quick fix!

micahsnyder (approved):

PR #1294, opened 2024-06-27
Add a dependency for the version map.

sebastianas:
Add a dependency  on that version map so that it is re-run once it changes.

This has been missed in the initial pull request.


micahsnyder:
The code change looks good to me. The GitHub builds all passed as well. I would not have been surprised it caused some issue on Windows... but it seems happy enough. I would like to rebase it to add Jenkins CI changes to work with changes to our CI infra and see it pass there. But I'm optimistic it'll pass just fine.

Thank you for the PR @sebastianas

PR #1301, opened 2024-07-03
cmake/FindRust.cmake: inherit RUSTFLAGS from the environment

liushuyu:
This pull request changes the `FindRust.cmake` script so the `RUSTFLAGS` set from the environment won't be overridden.

This is needed because some downstream distributors (e.g. Linux distros) will need to append several Rust flags to get reproducible builds and/or make the project build on certain architectures.


micahsnyder:
Thanks @liushuyu this makes sense to me. Will test it soon as our internal CI system is working right. (everything seems to be broken right now, very frustrating!)

micahsnyder (approved):
Looks good to me.

Issue #1300, opened 2024-07-03
test_pe_cert_trust fails after 2024-07-28


bmwiedemann:
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds) (sponsored by the NLnet NGI0 fund), I found that
our clamav 1.3.1 package fails a test after 2024-07-28.

How to reproduce the problem
----------------------------

run the tests with the clock set to 2024-07-29, e.g. on openSUSE I do
```bash
osc checkout openSUSE:Factory/clamav && cd $_
osc build --vm-type=kvm --noservice --clean --build-opt=--vm-custom-opt="-rtc base=2024-07-29T00:00:00" standard
```

Attachments
-----------

```
FAILED clamscan/assorted_test.py::TC::test_pe_cert_trust - AssertionError: as...
```

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +16 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.
See https://reproducible-builds.org/ for why this matters.

Kangie:
Let's see.

>  self.step_name('Test that clam can trust an EXE based on an authenticode certificate check.')

Your test output is truncated, but looking at the test I assume we're failing to validate the certificate because it's not valid in 15 years.

The `test.exe` that this function evaluates is checked into the repo here: https://github.com/Cisco-Talos/clamav/blob/main/unit_tests/input/pe_allmatch/test.exe

I suspect that getting a signed bin that's valid for 15 years is likely to be more trouble than you excluding this test for reproducible builds :)

@micahsnyder is there any context that you can share around this?

n.b. I'm sure that you will uncover many legitimate bugs through this process, and please don't let me discourage you, but I feel like any users still running current versions of ClamAV in 15+ years have bigger problems than this given the lack of signature updates (etc).

bmwiedemann:
The certificate likely expires in 3 weeks, so something needs to be done about it  soon anyway.

In [other projects](https://github.com/eclipse/paho.mqtt.python/pull/854), we use a CA/certs that expire in 1000y. Or does it have to be signed by some official authority? Maybe there is a way to use a custom test-CA for that test?

Kangie:
> The certificate likely expires in 3 weeks, so something needs to be done about it soon anyway.

Oops, I did miss that!

micahsnyder:
[...]

PR #1305, opened 2024-07-10
Fix unit test caused by expiring signing certificate

micahsnyder:
The clamscan test "assorted_test.py::TC::test_pe_cert_trust" is about to fail because the "test.exe" test file was signed with a cert set to expire after only 2 years, and it has been 23 months.

While attempting to generate a new one that will last 73000 days (200 years), I discovered that any signing certificate set to expire after 2038 will fail the trust-check because the `ca.not_after` variable is maxed out `time_t` incapable of expressing a higher number. To fix this, I've upgraded the variables to `uint64_t`.

I also had to replace a bunch of generated signatures to match the new "test.exe".

Finally, I noticed that "ca.not_before" was being set to the token[8] instead of token[9], which presumably mean the "NotBefore" field for Trusted and Revoked Certificates was non-functional, as it was treating the "CertSign" boolean as the "NotBefore" value.

Fixes: https://github.com/Cisco-Talos/clamav/issues/1300


libclamav/crtmgr.h
```diff
@@ -66,8 +66,8 @@ typedef struct cli_crt_t {
     BIGNUM *n;
     BIGNUM *e;
     BIGNUM *sig;
-    time_t not_before;
-    time_t not_after;
+    uint64_t not_before;
+    uint64_t not_after;
```
bmwiedemann, line 70: time_t always was defined as signed int, which is why it rolls over at 0x7fffffff (for 32bit) - was the change to unsigned intentional here? The `>> 1` below seems to still account for the sign bit.

micahsnyder: Fair I guess the `>> 1` wouldn't be necessary since switching to unsigned.  I don't think there's any point in it being signed. (?)

bmwiedemann: There could be compatibility reasons. And there is really no point in being unsigned as we don't need the extra bit.

micahsnyder: Fair doesn't hurt to use `int64_t`. I just force-pushed the change.

ragusaa (approved):

bmwiedemann (approved):

bmwiedemann:
Hi, what is missing to get this merged?

micahsnyder:
> Hi, what is missing to get this merged?

We've had some issues with our internal Jenkins test pipelines that has slowed down some reviews on Github. I think this is good to go, though.

micahsnyder:
Rebased with upstream main branch to get CI fixes.

rsundriyal (approved):

bmwiedemann:
Is there a plan to make a release with this? Patching binary files in not so easy with some build systems.

peteanning:
@micahsnyder we are using 1.0.6 LTS building from source (now failing) will #1305 be released as a patch for 1.0.6

micahsnyder:
@peteanning Sorry for the late response. We will publish 1.0.7 (and 1.3.2) with the fix soon, alongside or shortly after 1.4.0.

PR #1307, opened 2024-07-11
CI: Fix recent CI failures

liushuyu:
This pull request fixes the recent CI failures with `clang-format,` and Rust builds on Windows.

### DIFF
--- a/cmake/FindRust.cmake
+++ b/cmake/FindRust.cmake
@@ -430,6 +430,8 @@ foreach(LINE ${LINE_LIST})
     string(REPLACE "native-static-libs: " "" LINE "${LINE}")
     string(REGEX REPLACE "  " "" LINE "${LINE}")
     string(REGEX REPLACE " " ";" LINE "${LINE}")
+    # remove linker flags
+    list(FILTER LINE EXCLUDE REGEX "/.*")
 
     if(LINE)
         message(STATUS "Rust's native static libs: ${LINE}")
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -56,9 +56,9 @@ jobs:
           - check: "win32/compat"
             exclude: ""
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Run clang-format style check for C/C++ programs.
-        uses: jidicula/clang-format-action@v4.4.1
+        uses: jidicula/clang-format-action@v4.13.0
         with:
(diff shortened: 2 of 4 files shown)


micahsnyder:
Heh I also put in something like this, though your linker-flag-fig may be better https://github.com/Cisco-Talos/clamav/pull/1306

I also have a similar change for the clang-format workflow in https://github.com/Cisco-Talos/clamav/pull/1280/files

micahsnyder:
Oh but thank you for working on this as well.  I appreciate your time.

liushuyu:
> Heh I also put in something like this, though your linker-flag-fig may be better #1306
> 
> I also have a similar change for the clang-format workflow in https://github.com/Cisco-Talos/clamav/pull/1280/files

Ah sorry, I did not check through all the pull requests. Although I can update this pull requests to get other GitHub Actions  updated (e.g. `vcpkg` stuff)

cmake/FindRust.cmake
```diff
@@ -430,6 +430,8 @@ foreach(LINE ${LINE_LIST})
     string(REPLACE "native-static-libs: " "" LINE "${LINE}")
     string(REGEX REPLACE "  " "" LINE "${LINE}")
     string(REGEX REPLACE " " ";" LINE "${LINE}")
+    # remove linker flags
+    list(FILTER LINE EXCLUDE REGEX "/.*")
```
micahsnyder, line 434: I think the LINE is for all the flags.  This works great if the flag we're filtering is at the end (which it has been from what I can tell), but if any real libs were to appear afterwards, I think this would also remove them. 

What do you think about this?
```
    list(FILTER LINE EXCLUDE REGEX "/[a-zA-Z:]*;?")
```

liushuyu: Oh, it's filtering `LINE` as a list. In CMake, a list is just a semi-colon separated string.

liushuyu: Here is the example: https://godbolt.org/z/bK1M5aabY

micahsnyder: Oh, that's awesome. ❤️

micahsnyder:
> Although I can update this pull requests to get other GitHub Actions updated (e.g. `vcpkg` stuff)

If you're up for it, it would be nice.  Thankfully those aren't broken (yet) but it would always be nice if they were working with the latest versions.

micahsnyder (approved):

Issue #1175, opened 2024-02-15
Clamd can't start if clamd.conf contains inline comments


jmlrt:
Describe the bug
----------------

Clamd start is failing when some comment are added on the same line as a config value in the `clamd.conf` file.

How to reproduce the problem
----------------------------

Edit `clamd.conf` and add `MaxRecursion 20 # set to 20` for example.

```
$ grep MaxRecursion /etc/clamav/clamd.conf
MaxRecursion 20 # set to 20
$ clamd
ERROR: Incorrect argument format for option MaxRecursion
ERROR: Can't open/parse the config file /etc/clamav/clamd.conf
```

Move the inline comment to its own line:

```
$ grep -B1 MaxRecursion /etc/clamav/clamd.conf
# set to 20
MaxRecursion 20
$ clamd
$ echo $?
0
$ ps -ef | grep clamd | grep -v grep
  121 clamav    0:16 {clamd} /run/rosetta/rosetta /usr/sbin/clamd clamd
```

Attachments
-----------

If the issue is reproducible only when scanning a specific file, attach it to the ticket.

micahsnyder:
I agree, it would be nice if clamav's config files would support inline comments.

userwiths:
[Here is an attempt at a solution. ](https://github.com/Cisco-Talos/clamav/pull/1308)I've tested it locally by manually editing the `clamd.conf` file and I'm open to suggestions in case you see anything that should be improved/changed. Hope it helps.

## Cause of issue:

The first issue I see is that inline comments are not accepted due to failed [Regex checks](https://github.com/Cisco-Talos/clamav/blob/main/common/optparser.c#L65). This would be the case for `MATCH_NUMBER`, `MATCH_SIZE` and `MATCH_BOOL` because of the end line assertion `$` not allowing for anything (comment in this case) between the value and the end of the line. Because of it, lines with an inline comment fail the check that is done [here](https://github.com/Cisco-Talos/clamav/blob/main/common/optparser.c#L1175).

In case we fix the Regex expression, we would need to account for the inline comments in the following `switch`/`case` [branches](https://github.com/Cisco-Talos/clamav/blob/main/common/optparser.c#L1204) (I've accounted for them right before the switch so we don't have code repetition). The approach I took was to check if the value of the argument contains the `#` character and in case it does, use `strtok` to get the content on its left side.

PR #1308, opened 2024-07-11
fix: Allow inline comments in clamd.conf

userwiths:
This PR attempts to resolve the issue mentioned in #1175 

1. Regex has been changed in order to allow matching a given line even if it has a comment (separated with space or not) from the main expression.
2. When passing through an argument, check if it contains the comment character `#` and if it does, try to get the content on its left side.

Thoughts: 
1. I have some doubts `strtok` is the **best** choice.
2. Another approach, better imo, would be to get the first match from the regex expression. I failed to find an example in the project where `cli_regexec` returns the made matches (most likely in the 4th argument, but it seems its used mostly with `NULL` through out the project).

### DIFF
--- a/common/optparser.c
+++ b/common/optparser.c
@@ -61,6 +61,7 @@
 #endif
 
 #define MAXCMDOPTS 150
+#define MAX_OPTION_LINE_LENGTH 1024
 
 #define MATCH_NUMBER "^[0-9]+((( +)?#(.*))?)$"
 #define MATCH_SIZE   "^[0-9]+[KMG]?(( +)?#(.*))?$"
@@ -964,10 +965,11 @@ struct optstruct *optparse(const char *cfgfile, int argc, char **argv, int verbo
     FILE *fs = NULL;
     const struct clam_option *optentry;
     char *pt;
-    const char *name = NULL, *arg;
+    const char *name = NULL;
+    char *arg;
     int i, err = 0, lc = 0, sc = 0, opt_index, line = 0, ret;
     struct optstruct *opts = NULL, *opts_last = NULL, *opt;
-    char buffer[1024], *buff;
+    char buffer[MAX_OPTION_LINE_LENGTH], *buff;
     struct option longopts[MAXCMDOPTS];
     char shortopts[MAXCMDOPTS];
     regex_t regex;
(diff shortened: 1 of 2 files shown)


micahsnyder:
Thanks for the PR! I will try to review this after we have things lined up to publish 1.4.0, or after 1.4.0 ships.

micahsnyder (changes requested):
I rebased this PR to re-run the tests.

I found just minor clang-format issues.

In manual testing, I found that for string-based options (vs bool, size, or number args) -- spaces before a `#` end up getting included in the option value.

For example, I tried this for `freshclam.conf`:
```
DatabaseMirror http://localhost:8000 # database.clamav.net
```

The result was that it tried to download `daily.cvd` from:
`http://localhost:8000 /daily.cvd` (with a space in the middle).

I would have to remove the space between `8000` and `#` to make it work. Like this:
```
DatabaseMirror http://localhost:8000# database.clamav.net
```
But this doesn't look very natural and is bound to get complaints from users that try this feature.

I also had trouble with getting a bool to work. I tried this in `freshclam.conf`, but it didn't enable verbose mode:
```
LogVerbose yes #no
```

userwiths:
Yeah, the `clang-format` I probably forgot about, now it should be formated as its supposed to.
I expected the tests to fail with the issue about spaces after arg, but on second thought there were no inline comments in the tests, currently added trimming of trailing spaces

Tried to see how I could implement a test for this fix, but I'm not sure if its possible to printout the options even in debug mode, as its managed in `libclamav` and not here.

val-ms (approved):
Thanks for fixing the clang-format issue and also the trailing spaces issue. I tried a variety of variations of trailing spaces for options with strings, booleans, numbers. As far as I can tell it works nicely.

I am seeing a warning with this change when building:
```
/home/micah/dev/clamav/clamav-1/common/optparser.c: In function ‘optparse’:
/home/micah/dev/clamav/clamav-1/common/optparser.c:1195:35: warning: passing argument 1 of ‘strtok’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
 1195 |             arg          = strtok(arg, "#");
      |                                   ^~~
In file included from /home/micah/dev/clamav/clamav-1/common/optparser.c:36:
/usr/include/string.h:336:14: note: expected ‘char * __restrict__’ but argument is of type ‘const char *’
  336 | extern char *strtok (char *__restrict __s, const char *__restrict __delim)
      |              ^~~~~~
/home/micah/dev/clamav/clamav-1/common/optparser.c:1196:26: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
 1196 |             trim_comment = arg + strlen(arg) - 1;
      |                          ^
```

I took the liberty of fixing this and extending your change to also trim trailing tabs. I tidied the commits and rebased them with main so we've got all the latest commits in the branch. Pending a good run through GitHub Actions and Jenkins, I'm happy with this and will merge it.

val-ms:
Thanks again for your work on this!

Issue #1174, opened 2024-02-15
Clamd can't start if FailIfCvdOlderThan is defined and a white list is used


jmlrt:
Describe the bug
----------------

`clamd` fail to start when `FailIfCvdOlderThan` is defined in `clamd.conf` and a whitelist file is created in `/var/lib/clamav`.

How to reproduce the problem
----------------------------

```
$ echo "CVE_2012_0773-2" > /var/lib/clamav/my_whitelist.ign2
$ clamd
LibClamAV Error: cli_cvdverify: Can't read CVD header
LibClamAV Error: cl_cvdgetage: cvdgetfileage() failed for /var/lib/clamav/my_whitelist.ign2
ERROR: Broken or not a CVD file
$ clamconf -n
Checking configuration files in /etc/clamav

Config file: clamd.conf
-----------------------
AlertExceedsMax yes
FailIfCvdOlderThan 10
LocalSocket /home/clamav/clamd.sock
LogClean yes
LogFile  /home/clamav/clamd.log
LogFileMaxSize 0
MaxDirectoryRecursion 100
MaxFileSize 0
MaxFiles 0
MaxScanSize 0
MaxScanTime 0
MaxThreads 6
OnAccessExcludeUID 23
OnAccessIncludePath /clamav/files-to-scan/
OnAccessMaxFileSize 0
OnAccessMaxThreads 6
PidFile /home/clamav/clamd.pid
SelfCheck 72000
User clamav

Config file: freshclam.conf
---------------------------
PidFile = "/run/clamav/freshclam.pid"
DatabaseMirror = "database.clamav.net"

clamav-milter.conf not found

Software settings
-----------------
Version: 1.2.2
Optional features supported: MEMPOOL AUTOIT_EA06 BZIP2 LIBXML2 PCRE2 ICONV JSON

Database information
--------------------
Database directory: /var/lib/clamav
main.cvd: version 62, sigs: 6647427, built on Thu Sep 16 12:32:42 2021
bytecode.cvd: version 334, sigs: 91, built on Wed Feb 22 21:33:21 2023
daily.cvd: version 27181, sigs: 2052851, built on Sat Feb 10 09:34:05 2024
[3rd Party] my_whitelist.ign2: 1 sig
Total number of signatures: 8700370

Platform information
--------------------
uname: Linux 6.6.12-linuxkit #1 SMP Fri Jan 19 08:53:17 UTC 2024 x86_64
[...]

PR #1309, opened 2024-07-12
 fix: Allow `clamd` to start normally when a whitelist is present.

userwiths:
This PR is an attempt to fix issue #1174 

The given functionality was introduced in the [following commit](https://github.com/Cisco-Talos/clamav/commit/e4fe6654c1618bacf3bff9ed64c52615c8c53a97) and after taking a look at the [cli_cvdverify](https://github.com/Cisco-Talos/clamav/blob/e4fe6654c1618bacf3bff9ed64c52615c8c53a97/libclamav/cvd.c#L530) function I believe it is meant to verify only CVDs.

That is, it throws an error when encountering a clear text (whitelist, `.ign`,`.ign2`) file. I mimicked the way the allowed extensions in the Database folder were handled, and excluded the 2 clear text extensions mentioned above. This has locally allowed for the normal execution of `clamd`.

### DIFF
--- a/clamscan/manager.c
+++ b/clamscan/manager.c
@@ -1251,6 +1251,15 @@ int scanmanager(const struct optstruct *opts)
     if ((opt = optget(opts, "database"))->active) {
         while (opt) {
             if (optget(opts, "fail-if-cvd-older-than")->enabled) {
+                if (LSTAT(opt->strarg, &sb) == -1) {
+                    logg(LOGG_ERROR, "Can't access database directory/file: %s\n", opt->strarg);
+                    ret = 2;
+                    goto done;
+                }
+                if (!S_ISDIR(sb.st_mode) && !CLI_DBEXT_SIGNATURE(opt->strarg)) {
+                    opt = opt->nextarg;
+                    continue;
+                }
                 if (check_if_cvd_outdated(opt->strarg, optget(opts, "fail-if-cvd-older-than")->numarg) != CL_SUCCESS) {
                     ret = 2;
                     goto done;
--- a/libclamav/cvd.c
+++ b/libclamav/cvd.c
@@ -851,7 +851,7 @@ cl_error_t cl_cvdgetage(const char *path, time_t *age_seconds)
         if (!strcmp(dent->d_name, ".") || !strcmp(dent->d_name, ".."))
             continue;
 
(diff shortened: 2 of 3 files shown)


micahsnyder (changes requested):
Thank you for working on a solution to this bug.

The same issue occurs when adding individual signature files outside of a CVD or CLD archive. E.g. if you drop in "clamav.hdb" file containing a hash-based signature, then `clamd` loading will still fail with:
```

❯ ./install/sbin/clamd -F
LibClamAV Error: cli_cvdverify: Can't read CVD header
LibClamAV Error: cl_cvdgetage: cvdgetfileage() failed for /home/micah/workspace/clamav-micah-5/build/install/share/clamav/clamav.hdb
```

So rather than having `cl_cvdgetage` ignore specific extensions like "`.ign2`", it would be better to only verify for "`.cvd`" and "`.cld`".

userwiths:
Yeah, I'm not familiar with all of the extensions so the clarification is more than welcome.
As discussed now the `cl_cvdgetage` checks **only** `CVD` and `CLD` files.

micahsnyder:
Hi @userwiths  I thought of one other issue.  If anyone tries to run clamscan where they pass in specific databases rather than using a database directory, it will still fail.

E.g.:
```sh
❯ clamscan -d ~/clamav.hdb -d ~/.cvdupdate/database/bytecode.cvd --fail-if-cvd-older-than=5 /path/to/file
LibClamAV Error: cli_cvdverify: Can't read CVD header
ERROR: Broken or not a CVD file
```

I think to fix this, we'll need to check if the database path ends with ".cvd" or ".cld" in this location: https://github.com/userwiths/clamav/blob/8474e6a86d45ba2b084bd7d3719947db90aa5543/clamscan/manager.c#L1253-L1254

What do you think?

micahsnyder:
> I noticed that clamscan proceeds without a notice even if a file/directory passed with -d does not exist.

I wasn't able to reproduce this, or else I don't fully understand what you mean.  I believe it should fail if the `-d` file or directory does not exist.  So I'm good with your change either way. 

Your PR seems good to me.  I was doing some local testing after a rebase.  I'm going to squash it down to one commit as well and then force-push that.

micahsnyder (approved):

PR #1312, opened 2024-07-20
Clam 2277 extract images from ole2 (NOT READY FOR REVIEW)

ragusaa:


val-ms:
I have made a backup of this branch in the Cisco-Talos/clamav repo, in case the original goes away: https://github.com/val-ms/clamav/pull/new/CLAM-2277-ExtractImagesFromOle2

val-ms:
Closing stale PR. We have a pure rust ole2 parser prototype which does the same. Will either backport, or just opensource inkie and continue dev there.

PR #1318, opened 2024-07-26
INSTALL.md: change LLVM description

Artoria2e5:
Somehow the changes from #692 did not come with a doc update. The cap at version 13 is given from #574. I did not try to build it with LLVM 14, so I don't know whether it truly doesn't compile or simply fails due to the CMake thing.


micahsnyder (approved):
Thanks for identifying and submitting this fix. Looks good to me.

Artoria2e5:
Should #574 be closed then? At least the documentation is all good now.

micahsnyder:
> Should #574 be closed then? At least the documentation is all good now.

Good catch.  Thank you

Issue #1155, opened 2024-01-25
Clamav freshclam fails when /var/log/clamav/freshclam.log does not exist


Atrate:
Describe the bug
----------------

When `/var/log/clamav/freshclam.log` does not exist (such as when `/var/log` is mounted as `tmpfs` to extend SSD (or SD card on rpis) longevity), freshclam does not create the log file and simply crashes (the systemd freshclam service).

How to reproduce the problem
----------------------------

1. Mount `/var/log/` as `tmpfs`.
2. Reboot
3. Start `clamav-freshclam.service`

Clamconf:

```
Checking configuration files in /etc/clamav

Config file: clamd.conf
-----------------------
LogFile = "/var/log/clamav/clamd.log"
LogTime = "yes"
PidFile = "/run/clamav/clamd.pid"
TemporaryDirectory = "/tmp"
LocalSocket = "/run/clamav/clamd.ctl"
User = "clamav"

Config file: freshclam.conf
---------------------------
PidFile = "/run/clamav/freshclam.pid"
UpdateLogFile = "/var/log/clamav/freshclam.log"
DatabaseMirror = "database.clamav.net"

Config file: clamav-milter.conf
-------------------------------
LogFile = "/var/log/clamav/clamav-milter.log"
LogTime = "yes"
PidFile = "/run/clamav/clamav-milter.pid"
TemporaryDirectory = "/tmp"
User = "clamav"

Software settings
-----------------
Version: 1.2.1
Optional features supported: MEMPOOL AUTOIT_EA06 BZIP2 LIBXML2 PCRE2 ICONV JSON RAR

Database information
--------------------
Database directory: /var/lib/clamav
bytecode.cvd: version 334, sigs: 91, built on Wed Feb 22 22:33:21 2023
main.cvd: version 62, sigs: 6647427, built on Thu Sep 16 14:32:42 2021
daily.cvd: version 27164, sigs: 2051511, built on Wed Jan 24 10:45:32 2024
Total number of signatures: 8699029

Platform information
--------------------
uname: Linux 6.6.13-hardened1-1-hardened #1 SMP PREEMPT_DYNAMIC Sun, 21 Jan 2024 00:24:54 +0000 x86_64
OS: Linux, ARCH: x86_64, CPU: x86_64
Full OS version: "Arch Linux"
zlib version: 1.3 (1.3), compile flags: a9
platform id: 0x0a21bfbf08000000000d0201

Build information
-----------------
[...]

PR #1322, opened 2024-07-29
fix: Attempt to create log file if it does not exist.

userwiths:
Fixes: #1155 
It might seem a bit messy but the gist of the PR is that.

* We introduce `fc_upsert_logg_file` method, which attempts to create the directories & log file as required.
* We move the `drop_privileges` case/branch later during the execution. If we do this before we `fc_upsert_logg_file` our log file, then we can't use `root` privileges even if `freshclam` is started as `root`.
* In order to `chown` the directories (and file, overkill?) we introduce the field `dbOwner` in the `fc_config` structure.

Seems this has been handled in documentation as a [step of the setup](https://docs.clamav.net/manual/Usage/Configuration.html?highlight=log%20file#freshclamconf).

### DIFF
--- a/libfreshclam/libfreshclam.c
+++ b/libfreshclam/libfreshclam.c
@@ -123,6 +123,58 @@ const char *fc_strerror(fc_error_t fcerror)
     }
 }
 
+int fc_upsert_logg_file(fc_config *fcConfig)
+{
+    int ret = 0;
+    char* current_dir = "/";
+    char* file_path = strdup(fcConfig->logFile);
+    char* log_file = fcConfig->logFile;
+    char* token = strtok(file_path, "/");
+    FILE *logg_fp = NULL;
+    struct passwd *current_user = getpwuid(getuid());
+    struct passwd *db_owner = getpwnam(fcConfig->dbOwner);
+    current_dir = (char*)malloc(2);
+    strcpy(current_dir, "/");
+    STATBUF sb;
+
+    while (token != NULL) {
+        current_dir = (char*)realloc(current_dir, strlen(current_dir) + strlen(token) + 2);
+        strcat(current_dir, token);
+        token = strtok(NULL, "/");
+        if(token == NULL) {
+            break;
+        }
+        if(LSTAT(current_dir, &sb) == -1) {
+            if(mkdir(current_dir, 0755) == -1) {
+                printf("ERROR: Failed to create required directory %s. Will continue without writing in %s.\n", current_dir, log_file);
+                ret = -1;
+                goto cleanup;
+            }
+            if(chown(current_dir, db_owner->pw_uid, db_owner->pw_gid) == -1) {
+                printf("ERROR: Failed to change owner of %s to %s. Will continue without writing in %s.\n", current_dir, fcConfig->dbOwner, log_file);
+                ret = -1;
+                goto cleanup;
+            }
+        }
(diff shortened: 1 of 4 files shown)


userwiths:
@micahsnyder tagging you cause this might have been forgotten. Managed to get the CI all green and put it in review. Just a reminder.

libfreshclam/libfreshclam.c
```diff
@@ -123,6 +123,71 @@ const char *fc_strerror(fc_error_t fcerror)
     }
 }
 
+int fc_upsert_logg_file(fc_config *fcConfig)
+{
+    if (fcConfig->logFile == NULL) {
+        return 0;
+    }
+    int ret = 0, field_no          = 1;
+    char *current_path, *file_path = cli_safer_strdup(fcConfig->logFile), *token;
```
micahsnyder, line 132: These pointers should be initialized to NULL.  

The strdup call should also be on a separate line and should have error handling.

libfreshclam/libfreshclam.c
```diff
@@ -123,6 +123,71 @@ const char *fc_strerror(fc_error_t fcerror)
...
+    FILE *logg_fp = NULL;
+    token         = cli_strtok(file_path, field_no++, PATHSEP);
```
micahsnyder, line 134: I think it would be easier to follow if this line were moved down to just above the `while()` loop. 

On the topic of readability, Some blank lines between unrelated lines of code would be nice on the eyes.

libfreshclam/libfreshclam.c
```diff
@@ -123,6 +123,71 @@ const char *fc_strerror(fc_error_t fcerror)
...
+        strcat(current_path, PATHSEP);
+    }
+    if ((logg_fp = fopen(fcConfig->logFile, "at")) == NULL) {
+        printf("ERROR: Can't open %s in append mode (check permissions!).\n", fcConfig->logFile);
+        ret = -1;
+        goto cleanup;
+    }
+#ifndef _WIN32
+    lchown(fcConfig->logFile, db_owner->pw_uid, db_owner->pw_gid);
+#endif /* _WIN32 */
```
micahsnyder, line 176: Some inline comments on this process would be nice. 

My understanding is it basically tries to do a `mkdir -p` on the logfile directory path and change ownership to the database owner as needed.  Once all the directories are confirmed to exist and be owned by the right user, it will open a log file and change ownership as needed.

The logic seems good to me for linux/unix systems. I'll have to try it out.

I see an effort to compile on Windows but I don't think it'll work on Windows. Paths on windows start with a drive letter, unless it's a relative path.  Since this is called for all systems, I expect it to fail to run on Windows, or create some extra directories somewhere (maybe in the current directory?). 

This change is also going to require that the `freshclam.conf` `UpdateLogFile` option is always an absolute path.  I suspect it works okay as a relative path right now so this could possibly break someone's configuration.

micahsnyder (changes requested):
I have concerns that this will fail on Windows and break the ability to use relative paths for `UpdateLogFile` on all platforms.

opoplawski:
Any chance that this might get completed at some point?

val-ms:
If @userwiths is unable or uninterested in continuing this work because of my long lag times in responding, I don't blame them. Perhaps I can pick up where they left off when I have a little downtime from my primary tasks.

PR #1356, opened 2024-09-01
CMake: improve ncurses detection; separate TINFO_LIBRARY input

micahsnyder:
I found that on macOS, setting the libncurses.a and libtinfo.a into the same CURSES_LIBRARY variable does not find or link with libtinfo.a.

To fix this, this commit adds a separate TINFO_LIBRARY variable. In the end, CURSES_LIBRARIES and the Curses::curses CMake TARGET will still have both libraries set, if both are provided.

This fix is necessary if the ncurses was built with `--with-terminfo`. I think we got away without it on Linux because of pkg-config.

### DIFF
--- a/cmake/FindCURSES.cmake
+++ b/cmake/FindCURSES.cmake
@@ -13,7 +13,7 @@ Imported Targets
 This module provides the following imported targets, if found:
 
 ``Curses::curses``
-  The CURSES library
+  The CURSES library and possibly TINFO library
 
 Result Variables
 ^^^^^^^^^^^^^^^^
@@ -34,10 +34,14 @@ Cache Variables
 
 The following cache variables may also be set:
 
-``CURSES_INCLUDE_DIR``
-  The directory containing ``foo.h``.
+``NCURSES_INCLUDE_DIR``
+  The directory containing ``ncurses.h``.
+``PDCURSES_INCLUDE_DIR``
+  The directory containing ``curses.h``.
 ``CURSES_LIBRARY``
   The path to the CURSES library.
+``TINFO_LIBRARY``
+  The path to the TINFO library.
 
 #]=======================================================================]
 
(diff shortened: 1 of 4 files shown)


rsundriyal (dismissed):

cmake/FindCURSES.cmake
```diff
@@ -69,82 +73,98 @@ if(NCURSES_NOT_FOUND EQUAL -1)
...
-      VERSION_VAR CURSES_VERSION
+        FOUND_VAR CURSES_FOUND
+        REQUIRED_VARS
+            CURSES_LIBRARY
+            NCURSES_INCLUDE_DIR
+        VERSION_VAR CURSES_VERSION
+    )
+
+    include(FindPackageHandleStandardArgs)
+    find_package_handle_standard_args(TINFO
```
val-ms, line 84: I re-tested this after a rebase to double-check my work and encountered this warning:
```

-- Found CURSES: /home/micah/.mussels/install/host-static/lib/libncurses.a;/home/micah/.mussels/install/host-static/lib/libtinfo.a
CMake Warning (dev) at /home/micah/.local/lib/python3.8/site-packages/cmake/data/share/cmake-3.30/Modules/FindPackageHandleStandardArgs.cmake:441 (message):
  The package name passed to `find_package_handle_standard_args` (TINFO) does
  not match the name of the calling package (CURSES).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  cmake/FindCURSES.cmake:84 (find_package_handle_standard_args)
  CMakeLists.txt:551 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found TINFO: /home/micah/.mussels/install/host-static/lib/libtinfo.a
```

That's... not great. It implies the correct approach is to make a separate `FindTINFO.cmake` file. There is probably an alternative solution. I might investigate how CMake's own `FindOpenSSL.cmake` does it considering that it finds libssl and libcrypto, neither of which are named "OpenSSL".

val-ms:
I discovered that ncurses may create many separate static libs - not just libtinfo.a.  I'm not sure what to do about that and it made me question if we even need this change. That's why I haven't merged it yet. I want to investigate further to gain some confidence.

val-ms:
Rebased and will re-test

PR #1372, opened 2024-09-23
Resolve incompatible-pointer-types

Kangie:
Explicitly cast regex match to `const char **`; we don't modify the value once it's set like this; we should be safe.

### DIFF
--- a/libclamav/regex/engine.c
+++ b/libclamav/regex/engine.c
@@ -223,7 +223,7 @@ matcher(struct re_guts *g, const char *string, size_t nmatch,
 			dp = dissect(m, m->coldp, endp, gf, gl);
 		} else {
 			if (g->nplus > 0 && m->lastpos == NULL)
-				m->lastpos = (char **)cli_max_malloc((g->nplus+1) *
+				m->lastpos = (const char **)cli_max_malloc((g->nplus+1) *
 							sizeof(char *));
 			if (g->nplus > 0 && m->lastpos == NULL) {
 				free(m->pmatch);


libclamav/regex/engine.c
```diff
@@ -223,7 +223,7 @@ matcher(struct re_guts *g, const char *string, size_t nmatch,
 			dp = dissect(m, m->coldp, endp, gf, gl);
 		} else {
 			if (g->nplus > 0 && m->lastpos == NULL)
-				m->lastpos = (char **)cli_max_malloc((g->nplus+1) *
+				m->lastpos = (const char **)cli_max_malloc((g->nplus+1) *
```
micahsnyder, line 226: It feels like the correct fix is to remove `const` in the `lastpos` definition.  Check out the warnings in the "Files Changed" tab.

micahsnyder (changes requested):

Kangie:
Fixed, thanks.

Kangie:
test failure seems unrelated; it's in rust code.

libclamav/regex/engine.c
val-ms: Well now I'm confused. Now I see issues because `sp` is declared as `const char *sp;`
image

It looks like that was intended to be an improvement with an upgrade to the newer version of the regex module as well: https://github.com/Cisco-Talos/clamav/commit/87cdd70037048dead11f0d91e9d097d5e3917b10

Perhaps we should cast like:
```
m->lastpos[lev] = (const char*)sp;
```
(and for the second one)

What do you think @Kangie ?

PR #1415, opened 2024-12-03
Fix CVD dsig verification when hash starts with zeros

micahsnyder:
Occasionally the MD5 hash for RSA-based digital signature verification begins with zeros. A bug in how we convert the RSA decoded plain text from a big number back to a hex string causes it to write the number to the far left of the plain text buffer. If the number is smaller than a hash, then zero-padding ends up on the right when it should've been on the left.


tj-cisco (dismissed):

micahsnyder:
Per discussion with @TheRaynMan and @tj-cisco will add a retry in case the md5 for signing has the 00 prefix. That way new sigtool doesn't make CVD's that fail to verify on older clamav.

rsundriyal (approved):

Issue #1304, opened 2024-07-08
libclamav.map: Remove CLAMAV_1.0.0…CLAMAV_0.104.0


val-ms:
@sebastianas wishes to remove versioning from the libclamav.so libclamav.map.

As @opoplawski noted, the only value for what we have now appears to be in tracking when functions were added to the API. 
Since it is more work to maintain I suppose we may as well remove it next time we bump the libclamav SOVERSION.

Refs:
- https://github.com/Cisco-Talos/clamav/issues/1296
- https://github.com/Cisco-Talos/clamav/pull/804#issuecomment-1482360456

Kangie:
No objections here.

rma-x:
Is it intentional that the SOVERSION wasn't bumped to 13 although symbols (`CLAMAV_*`) were removed from the library?

This leads to the situation that packages that were linked to `libclamav.so.12` from ClamAV 1.4.x are unable to run with `libclamav.so.12` from ClamAV 1.5.x, because of these missing symbols. This includes our own `libfreshclam.so.3` which is still installed in the situation of upgrading from 1.4.3 to 1.5.0. It also causes issues with auto-generated RPM level package dependencies.

val-ms:
It was not intentional. 
When considering if we should bump [the CURRENT version for the SOVERSION in CMakeLists.txt for libclamav,](https://github.com/Cisco-Talos/clamav/blob/main/CMakeLists.txt#L39) I only considered if API's were removed or altered.  I had not considered the change to the libclamav.map regardign `CLAMAV_<VER>`. So that's my mistake. I'm sorry.

rma-x:
Thanks for the confirmation. Do you plan to fix it in 1.5.2 either by bumping the SOVERSION or by restoring the removed symbols and keeping them until the version needs to be bumped for other reasons? Of these two options I think restoring the symbols would be cleaner to do in a point release.

val-ms:
I think we'd get even more blowback if we changed the SOVERSION in a patch release. Even if it was to "correct" the SOVERSION. 

I don't know if restoring the symbols would be okay, either. I am not that familiar with symbol versioning.
It seems to me that MOVING [the affected symbols](https://github.com/Cisco-Talos/clamav/blob/rel/1.5/libclamav/libclamav.map#L62-L68) back under versioned sections would break things for anyone upgrading from 1.5.1 to 1.5.2. 
[...]

PR #1417, opened 2024-12-11
Clam 2677 2678 FIPS compliant CVD codesigning

val-ms:
Add FIPS-compliant CVD signing and verification.

This change bears similarities to https://github.com/Cisco-Talos/clamav/pull/1344, adding external code signing signature files (`.cvd.sign`) to accompany existing `.cvd` databases. 

The `.sign` take a form similar to other clamav signatures where comment lines start with a `#` and all other lines are signatures. Each signature line takes the form:

`min_flevel : max_flevel (optional) : sig_format : public_key_fingerprint : signature_text`

`min_flevel`:  the minimum flevel required to support the signature format.
`max_flevel`:  the maximum flevel supporting the signature format. This is optional, and may be empty. 
`signature_format`:  for the initial implementation, it must be: "pkcs7-pem".
`signature_text`:  must be printable ascii with no new lines. If "sig_format" is "pkcs7-pem" it is the base64 encoded part of the a PEM signature (aka header and footer removed) with all new-lines removed. if additional formats are added, this field would be used to store things like the hash of the file, the algorithm, plus any other required signing details. 

Adds X509 certificate chain based signing with PKCS7-PEM external signatures distributed alongside CVD's in a custom `.cvd.sign` format.

Adds a Rust implementation for parsing, verifying, and unpacking CVD files.

Now installs a 'certs' directory in the app config directory (e.g. <prefix>/etc/certs). The install location is configurable.
The CMake option to configure the CVD certs directory is:
  `-D CVD_CERTS_DIRECTORY=PATH`

New options to set an alternative CVD certs directory:
- Commandline for freshclam, clamd, clamscan, and sigtool is:
  `--cvdcertsdir PATH`
- Env variable for freshclam, clamd, clamscan, and sigtool is:
  `CVD_CERTS_DIR`
- Config option for freshclam and clamd is:
  `CVDCertsDirectory PATH`

Sigtool:
- Add sign/verify commands.
- Also verify CDIFF external digital signatures when applying CDIFFs.
- Place commonly used commands at the top of --help string.
- Fix up manpage.

Freshclam:
- Will try to download .sign files to verify CVDs and CDIFFs.
- Fix an issue where making a CLD would only include the CFG file for daily and not if patching any other database.

libclamav.so:
- Bump version to 13:0:1 (aka 12.1.0).
- Also remove libclamav.map versioning.
  Resolves: https://github.com/Cisco-Talos/clamav/issues/1304
- Add two new API's to the public clamav.h header:
  ```c
  extern cl_error_t cl_cvdverify_ex(const char *file,
                                    const char *certs_directory);

  extern cl_error_t cl_cvdunpack_ex(const char *file,
                                    const char *dir,
                                    bool dont_verify,
                                    const char *certs_directory);
  ```
  The original `cl_cvdverify` and `cl_cvdunpack` are deprecated.
- Add `cl_engine_field` enum option `CL_ENGINE_CVDCERTSDIR`.
  You may set this option with `cl_engine_set_str` and get it with `cl_engine_get_str`, to override the compiled in default CVD certs directory.

libfreshclam.so: Bump version to 4:0:0 (aka 4.0.0).

Adds sigtool sign/verify tests and test certs.

Make it so downloadFile doesn't throw a warning if the server doesn't have the .sign file.

Replace use of md5-based FP signatures in the unit tests with sha256-based FP signatures because the md5 implementation used by Python may be disabled in FIPS mode. 

Fixes: https://github.com/Cisco-Talos/clamav/issues/564
Fixes: https://github.com/Cisco-Talos/clamav/issues/1411

Replaces: https://github.com/Cisco-Talos/clamav/pull/1344


libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,552 @@
...
+                        store_builder.add_cert(self.root_ca.clone())?;
+                        let store = store_builder.build();
+
+                        // verify signature
+                        let certs = stack::Stack::new()?;
+                        let flags = Pkcs7Flags::DETACHED | Pkcs7Flags::BINARY;
+                        let mut output = Vec::new();
+
+                        let result =
+                            pkcs7.verify(&certs, &store, Some(data), Some(&mut output), flags);
```
micahsnyder, line 527: Need to make sure CRL / OSCP verification is not performed in the initial release, and to set up our actual signing kes/certs have that stuff set up for future use.

CMakeLists.txt
```diff
@@ -44,8 +44,8 @@ math(EXPR LIBCLAMAV_SOVERSION "${LIBCLAMAV_CURRENT} - ${LIBCLAMAV_AGE}")
 set(LIBCLAMAV_VERSION "${LIBCLAMAV_SOVERSION}.${LIBCLAMAV_AGE}.${LIBCLAMAV_REVISION}")
 HexVersion(LIBCLAMAV_VERSION_NUM ${LIBCLAMAV_CURRENT} ${LIBCLAMAV_REVISION} ${LIBCLAMAV_AGE})
 
-set(LIBFRESHCLAM_CURRENT  3)
-set(LIBFRESHCLAM_REVISION 2)
+set(LIBFRESHCLAM_CURRENT  4)
+set(LIBFRESHCLAM_REVISION 0)
 set(LIBFRESHCLAM_AGE      0)
```
micahsnyder, line 49: Must also bump libclamav version since adding 2 new API's

CMakeOptions.cmake
```diff
@@ -7,13 +7,19 @@ if(WIN32)
...
+        "ClamAV CA certificates directory.")
 else()
     set(APP_CONFIG_DIRECTORY
         "etc" CACHE STRING
         "App Config directory.")
     set(DATABASE_DIRECTORY
         "share/clamav" CACHE STRING
         "Database directory.")
+    set(CERTS_DIRECTORY
+        "etc/certs" CACHE STRING
```
micahsnyder, line 21:
```suggestion
        "${APP_CONFIG_DIRECTORY}/certs" CACHE STRING
```

libclamav/clamav.h
```diff
@@ -1139,6 +1140,15 @@ extern struct cl_cvd *cl_cvdparse(const char *head);
  */
 extern cl_error_t cl_cvdverify(const char *file);
```
micahsnyder, line 1141: mark this one as deprecated as well.

libclamav/cvd.c
```diff
@@ -519,197 +522,185 @@ void cl_cvdfree(struct cl_cvd *cvd)
...
     struct cl_engine *engine;
-    FILE *fs;
     cl_error_t ret;
-    int dbtype = 0;
-
-    if ((fs = fopen(file, "rb")) == NULL) {
-        cli_errmsg("cl_cvdverify: Can't open file %s\n", file);
-        return CL_EOPEN;
-    }
+    cvd_type dbtype = CVD_TYPE_CVD;
```
micahsnyder, line 534:
```suggestion
    cvd_type dbtype = CVD_TYPE_UNKNOWN;
```

libclamav/cvd.c
```diff
@@ -519,197 +522,185 @@ void cl_cvdfree(struct cl_cvd *cvd)
...
+            goto done;
+        }
+    }
 
-    ret = cli_cvdload(fs, engine, NULL, CL_DB_STDOPT | CL_DB_PUA, dbtype, file, 1);
+    if (!!cli_strbcasestr(file, ".cld")) {
+        dbtype = CVD_TYPE_CLD;
+    } else if (!!cli_strbcasestr(file, ".cud")) {
+        dbtype = CVD_TYPE_CUD;
+    }
```
micahsnyder, line 555:
```suggestion
    if (!!cli_strbcasestr(file, ".cvd")) {
        dbtype = CVD_TYPE_CVD;
    } else if (!!cli_strbcasestr(file, ".cld")) {
        dbtype = CVD_TYPE_CLD;
    } else if (!!cli_strbcasestr(file, ".cud")) {
        dbtype = CVD_TYPE_CUD;
    } else {
        cli_errmsg("cl_cvdverify: File is not a CVD, CLD, or CUD: %s\n", file);
        goto done;
    }
```

libclamav/cvd.c
```diff
@@ -519,197 +522,185 @@ void cl_cvdfree(struct cl_cvd *cvd)
...
-        return ret;
+    /* Open the cvd and read the header */
+    cvd = cvd_open(filename, &cvd_open_error);
+    if (!cvd) {
+        cli_errmsg("cli_cvdload: Can't open CVD file %s: %s\n", filename, ffierror_fmt(cvd_open_error));
+        goto done;
+    }
+
+    /* For actual .cvd files, verify the digital signature. */
+    if (dbtype == CVD_TYPE_CVD) {
```
micahsnyder, line 591: Do we want to also check this for CUD?  Probably. This would be a case where database is not signed with old MD5-RSA, and only signed with external PKCS7 sig.

libclamav/cvd.c
```diff
@@ -519,197 +522,185 @@ void cl_cvdfree(struct cl_cvd *cvd)
...
+        if (!cvd_verify(
+                cvd,
+                engine->certs_directory,
+                false,
```
micahsnyder, line 595: Consider if we want to make a config or commandline option to disable the older md5-based RSA checks, or simply let it fail if FIPS is enabled and the cert-based method doesn't work/isn't available.

I think it's okay either way.

libclamav_rust/src/cvd.rs
```diff
@@ -0,0 +1,800 @@
...
+            return ffi_error!(
+                err = err,
+                Error::CannotVerify(format!("Invalid destination path: {}", e))
+            );
+        }
+    };
+
+    match cvd.unpack_to(&destination_path) {
+        Ok(()) => {
+            println!("CVD unpacked successfully");
```
micahsnyder, line 586:
```suggestion
            debug!("CVD unpacked successfully");
```

libclamav/others.h
```diff
@@ -320,6 +321,7 @@ struct cl_engine {
     uint32_t ac_mindepth;
     uint32_t ac_maxdepth;
     char *tmpdir;
+    char* certs_directory;
```
micahsnyder, line 324: Check if I'm actually using this. I can't remember... i think it's just in the engine and not used in the ctx.

libclamav_rust/Cargo.toml
```diff
@@ -26,6 +26,13 @@ inflate = "0.4"
 bzip2-rs = "0.1"
 byteorder = "1.5"
 delharc = "0.6"
+clam-sigutil = { path = "../../clamav-signature-util" }
```
micahsnyder, line 29: This is temporarily pointing a copy of the rust clam-sigutil project in my workspace. Will change it to a github URL soon.

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,563 @@
...
+
+    let data = std::fs::read(target_file_path)?;
+    let pkcs7 = signer.sign(&data)?;
+
+    // Now convert the pkcs7 to a DigitalSig struct which may be converted to a .sign file signature line.
+    let signature = DigitalSig::Pkcs7(pkcs7);
+    let mut sig_bytes: SigBytes = SigBytes::new();
+    signature.append_sigbytes(&mut sig_bytes)?;
+
+    // If the signature file already exists, open it and append the signature to it.
```
micahsnyder, line 193: Consider adding a parameter to replace the .sign file with a new, or a append it (up to the user).

CMakeLists.txt
micahsnyder: Done

micahsnyder:
`sigtool --sign`: After debug-logging the certs for the cert stack, should info-print:
- cert Serial
- CommonName 
- the issuer commonName (if an issuer exists)

micahsnyder:
For verifying,

1. Load all ca certs from the ca certs directory into the x509 store and put the x509 store into the Verifier.

2. The Verifier should be stored in the cl_engine.

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,588 @@
...
+            if append {
+                options.append(true);
+            } else {
+                options.truncate(true);
+            }
+        }
+
+        let mut writer = std::io::BufWriter::new(options.open(signature_file_path)?);
+
+        if !signature_file_path.exists() || !append {
```
micahsnyder, line 230:
```suggestion
        if !signature_file_path.exists() {
```

Should still create the header in append mode.

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,588 @@
...
+    // Now convert the pkcs7 to a DigitalSig struct which may be converted to a .sign file signature line.
+    let signature = DigitalSig::Pkcs7(pkcs7);
+    let mut sig_bytes: SigBytes = SigBytes::new();
+    signature.append_sigbytes(&mut sig_bytes)?;
+
+    let mut writer = {
+        let mut options = std::fs::OpenOptions::new();
+        options.write(true).create(true);
+
+        if signature_file_path.exists() {
```
micahsnyder, line 220: record if it exists. Open() will make it exist.

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,588 @@
...
+                options.truncate(true);
+            }
+        }
+
+        let mut writer = std::io::BufWriter::new(options.open(signature_file_path)?);
+
+        if !signature_file_path.exists() || !append {
+            writer.write_all(b"#clamsign-1.0\n")?;
+        } else if append {
+            writer.seek(std::io::SeekFrom::End(0))?;
```
micahsnyder, line 233:
```suggestion
```

append mode should drop us at the end of the file. This is probably not needed.

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,588 @@
...
+            } else {
+                options.truncate(true);
+            }
+        }
+
+        let mut writer = std::io::BufWriter::new(options.open(signature_file_path)?);
+
+        if !signature_file_path.exists() || !append {
+            writer.write_all(b"#clamsign-1.0\n")?;
+        } else if append {
```
micahsnyder, line 232:
```suggestion
        } else {
```

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,588 @@
...
+        if signature_file_path.exists() {
+            if append {
+                options.append(true);
+            } else {
+                options.truncate(true);
+            }
+        }
+
+        let mut writer = std::io::BufWriter::new(options.open(signature_file_path)?);
+
```
micahsnyder, line 229:
```suggestion

        // TODO: When introducing next clamsign-version, will need to re-do the header if adding a signature to an existing file where the clamsign-version is older. Or potentially fail out because appending to the older clamsign-version is incompatible. 

```

libclamav_rust/src/codesign.rs
```diff
@@ -0,0 +1,588 @@
...
+}
+
+pub struct Verifier {
+    root_ca: X509,
+}
+
+impl Verifier {
+    pub fn new(root_ca_path: &Path) -> Result<Self, Error> {
+        let root_ca_bytes = std::fs::read(root_ca_path)?;
+        let root_ca: X509 = X509::from_pem(&root_ca_bytes)?;
```
micahsnyder, line 511: Should try to read from a pem stack not just a single one, in case it's a bundle. And we should be doing this for the certs_directory for each file in the certs directory. That way the verifier can verify all CVD's.

rsundriyal (approved):

libclamav/libclamav.map
TheRaynMan, line 1: Consolidating out-of-date version markers seems clear, but shouldn't CLAMAV_PUBLIC remain? Maybe moved to new line 71?

val-ms: It does remain (it's line 1).  You're looking at the equivalent of html closing tag (a la `</CLAMAV_PUBLIC>`).

certs/clamav-beta.crt
TheRaynMan, line 1: Ok for beta, will be root CA only for release.

libclamav/libclamav.map
TheRaynMan: Ah, just a bookend to line 1. Nevermind.

TheRaynMan (approved):

PR #1446, opened 2025-01-30
Fix unit tests srcdir var - env vars are case sensitive

kolbma:
In unit tests has been used `getenv("srcdir")` but the set variable is called `SRCDIR`.  
So on platforms where env vars are case sensitive, the unit tests failed 1.4.2 and master at least.

### DIFF
--- a/unit_tests/check_bytecode.c
+++ b/unit_tests/check_bytecode.c
@@ -501,7 +501,7 @@ END_TEST
 
 static void runload(const char *dbname, struct cl_engine *engine, unsigned signoexp)
 {
-    const char *srcdir = getenv("srcdir");
+    const char *srcdir = getenv("SRCDIR");
     char *str;
     unsigned signo = 0;
     int rc;
--- a/unit_tests/check_clamav.c
+++ b/unit_tests/check_clamav.c
@@ -1899,7 +1899,7 @@ void errmsg_expected(void)
 int open_testfile(const char *name, int flags)
 {
     int fd;
-    const char *srcdir = getenv("srcdir");
+    const char *srcdir = getenv("SRCDIR");
     char *str;
 
     if (!srcdir) {


val-ms (commented):
While it looks this way -- SRCDIR as defined by CMake is not an environment variable but is actually a preprocessor defined here: https://github.com/Cisco-Talos/clamav/blob/9cb28e51e62746a4b1d64a368cd9236e93ff40bf/unit_tests/CMakeLists.txt#L138

As noted in the comment [here](https://github.com/Cisco-Talos/clamav/blob/902623972d981399027dae3a4734b0e0141966c4/unit_tests/check_bytecode.c#L510) and [here](https://github.com/Cisco-Talos/clamav/blob/45c6938be878ca877589d6412367bc0c43bc32bf/unit_tests/check_clamav.c#L1907), `srcdir` as an environment variable had been defined by Automake.  Support for reading `srcdir` from an environment variable should probably be removed, unless there is a legitimate case for running the unit tests where `srcdir` is different than the location when it had been originally compiled. 

Did you find that this change fixed something in your build/test environment? What are your steps for building/testing clamav?  Are you moving the files after build but before running the tests?

kolbma:
It is built in Alpine Linux package build system, maybe this is the cause, why the package maintainer doesn't update to 1.4.2.

cmake -B build -G Ninja
cmake --build build  
ctest -C build  

Complete here...  
https://gitlab.alpinelinux.org/alpine/aports/-/blob/3.21-stable/community/clamav/APKBUILD#L138. 
https://gitlab.alpinelinux.org/alpine/aports/-/blob/3.21-stable/community/clamav/APKBUILD#L159. 

This worked until 1.4.1.  
With 1.4.2 there is missing a path part in the middle in front of unit_tests/input/... and with the patch the correct path is used.  
All other unit tests work.

val-ms:
The changes between 1.4.1 and 1.4.2 are few: https://github.com/Cisco-Talos/clamav/compare/clamav-1.4.1...clamav-1.4.2

I don't see anything related to `srcdir` / `SRCDIR`. I suspect that something else was wrong with the build that failed tests. 

My preferred change would be to drop the `getenv`'s instead, like this:
```patch
diff --git a/unit_tests/check_bytecode.c b/unit_tests/check_bytecode.c
index a07e3b0a5..5bc3429f2 100644
--- a/unit_tests/check_bytecode.c
+++ b/unit_tests/check_bytecode.c
@@ -501,17 +501,13 @@ END_TEST

 static void runload(const char *dbname, struct cl_engine *engine, unsigned signoexp)
 {
-    const char *srcdir = getenv("srcdir");
     char *str;
     unsigned signo = 0;
     int rc;
-    if (!srcdir) {
-        /* when run from automake srcdir is set, but if run manually then not */
-        srcdir = SRCDIR;
-    }
-    str = malloc(strlen(srcdir) + 1 + strlen(dbname) + 1);
+
+    str = malloc(strlen(SRCDIR) + 1 + strlen(dbname) + 1);
     ck_assert_msg(!!str, "malloc");
-    sprintf(str, "%s" PATHSEP "%s", srcdir, dbname);
+    sprintf(str, "%s" PATHSEP "%s", SRCDIR, dbname);

     rc = cl_load(str, engine, &signo, CL_DB_STDOPT);
     ck_assert_msg(rc == CL_SUCCESS, "failed to load %s: %s\n",
diff --git a/unit_tests/check_clamav.c b/unit_tests/check_clamav.c
index 1d55f1763..ce2f331ba 100644
--- a/unit_tests/check_clamav.c
+++ b/unit_tests/check_clamav.c
@@ -1899,17 +1899,11 @@ void errmsg_expected(void)
 int open_testfile(const char *name, int flags)
 {
     int fd;
-    const char *srcdir = getenv("srcdir");
     char *str;

-    if (!srcdir) {
-        /* when run from automake srcdir is set, but if run manually then not */
-        srcdir = SRCDIR;
-    }
-
-    str = malloc(strlen(name) + strlen(srcdir) + 2);
+    str = malloc(strlen(name) + strlen(SRCDIR) + 2);
     ck_assert_msg(!!str, "malloc");
-    sprintf(str, "%s" PATHSEP "%s", srcdir, name);
+    sprintf(str, "%s" PATHSEP "%s", SRCDIR, name);

     fd = open(str, flags);
     ck_assert_msg(fd >= 0, "open() failed: %s", str);
```

It works okay for me, but can you test it?

val-ms:
I've added my suggestion to https://github.com/Cisco-Talos/clamav/pull/1450, replacing this PR

Issue #1014, opened 2023-08-29
Forged ZIP archive manages to hide the Trojan from clamav


devgs:
Describe the bug
-----------------------

It seems that bad actors have managed to find a 'hole' in a ZIP format that allows them to inject a hidden content, in such a way, that it isn't visible to **clamav**.

In our case, it's a virus, so I'm not even sure whether it's wise to attach it here. That's why I'm going to add `.gif` to the file name as a weak safety measure:

[forged.zip.gif](https://github.com/Cisco-Talos/clamav/assets/1633661/f589c7ed-63bb-4f05-a349-18eed5e38036)

If the archive is extracted, using **unzip**, on Linux, then the resulting file (another ZIP archive) is now detected as having the virus.

Here's how [the other parties see the original file](https://www.virustotal.com/gui/file/3676ee70492d30e156eac1d1b8e7a99e18deb7beabd64dc89bbf18c6a9326579).

How to reproduce the problem
-----------------------

**1.** Have the **forged.zip.gif** as and example.

**2.** Rename it to **forged.zip**

**3.** Run **clamscan**:

```
clamscan --verbose --archive-verbose -d /var/db/clamav/ /home/foo/2/forged.zip
Loading:    28s, ETA:   0s [========================>]   12.88M/12.88M sigs
Compiling:   4s, ETA:   0s [========================>]       42/42 tasks

Scanning /home/foo/2/forged.zip
/home/foo/2/forged.zip: OK

----------- SCAN SUMMARY -----------
Known viruses: 12878904
Engine version: 1.0.1
Scanned directories: 0
Scanned files: 1
Infected files: 0
Data scanned: 1.03 MB
Data read: 0.96 MB (ratio 1.06:1)
Time: 41.003 sec (0 m 41 s)
Start Date: 2023:08:29 14:40:21
End Date:   2023:08:29 14:41:02
```

**4.** Use shell (note the **351470 extra bytes at beginning or within zipfile**):

```
$ ll
total 1008
drwxrwxr-x 2 foo foo    4096 Aug 29 14:22 ./
drwxr-xr-x 8 foo foo   12288 Aug 29 14:22 ../
-rw-rw-r-- 1 foo foo 1013510 Aug 29 14:22 forged.zip
$ file forged.zip
forged.zip: Zip archive data, at least v2.0 to extract
$ unzip forged.zip
Archive:  forged.zip
warning [forged.zip]:  351470 extra bytes at beginning or within zipfile
[...]

PR #1460, opened 2025-02-27
Extend zip parser to be less reliant on the 

frsell1:
Zip file concatenation is an evasion method used by threat actors in the wild: https://perception-point.io/blog/evasive-concatenated-zip-trojan-targets-windows-users/. ClamAV currently relies on the last central directory it can find and will therefore miss the first zip in the concatenation.

The proposed addition to the zip parser searches for local file headers between and around the files already found from the central directory header. It then adds these found files to the list of zip records found by the central directory.

The idea is that scanning for a local file header is time intensive, so we should minimize the number of bytes we need to scan. At the same time, we need to scan for local file headers since we cannot trust the central directory header as a source of truth.

There was already functionality that would look for local file headers after the found end of central directory header (at the end of the `cli_unzip` function). This required the appended local file headers to occur consecutively and it did not search for local file headers at the top of the zip file.

### DIFF
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -1256,7 +1256,7 @@ cl_error_t index_local_file_headers_within_bounds(
                                             coff,
                                             fsize - coff,
                                             NULL,
-                                            total_file_count,
+                                            total_file_count + 1,
                                             NULL,
                                             &ret,
                                             ctx,
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -1069,7 +1069,7 @@ cl_error_t index_the_central_directory(
 
             num_record_blocks++;
             /* zero out the memory for the new records */
-            memset(&(zip_catalogue[records_count]), 0, sizeof(struct zip_record) * (ZIP_RECORDS_CHECK_BLOCKSIZE * num_record_blocks - records_count));
+            memset(&(zip_catalogue[records_count]), 0, sizeof(struct zip_record) * (ZIP_RECORDS_CHECK_BLOCKSIZE * (num_record_blocks - records_count)));
         }
     } while (1);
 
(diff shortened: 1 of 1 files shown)


val-ms:
Fixes https://github.com/Cisco-Talos/clamav/issues/1014

val-ms (changes requested):
I created some feature/regression tests to verify that this fixes a few scenarios and ran into 3 bugs.  I've created fixes for those, and then added some opinionated code cleanup on top of that.  

It's too much to add in code review comments here, so I'm linking to my branch and if you're good with those changes, you or I can pull them onto your branch.

https://github.com/Cisco-Talos/clamav/compare/main...val-ms:clamav:CLAM-2722-update_zip_parser-tests

val-ms:
I should add that I LOVE the direction you went for solving this problem, and greatly appreciate your effort to make the code match our existing style.

frsell1:
Thanks for the review and fixes. Just merged your changes to this branch

frsell1:
Fixed a bug in `index_local_file_headers` where the end_offset would be set to the end of the next file in the zip. This caused `index_local_file_headers` to extract duplicate files that were already extracted in `index_the_central_directory`

val-ms:
This PR is currently causing two of our internal test suite tests to fail:
image

I'll investigate

DailyInvestors:
Good Morning, seeing how we haven't had the pleasure to introduce ourselves. I go by the nickname Daily Investors, in resetting a system now and running a root check. But got hit hard today after I made our post. But I have been studying Cyber Security for four years now, am fluent in Python and Java, and being in the stock market, I saw all this coming years ago. I will be helping with your program eventually. But at least for me, I'm getting hit hard in America frfr, FBI came out today, and admitted the truth
https://soloist.ai/dailyinvestors

https://snsim.moomoo.com/share/server/soakc?lang=en-us

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

val-ms:
Squashed the last two commits together to tidy up, and so the commits will be signed.

val-ms (approved):

Issue #1440, opened 2025-01-25
Windows: Log outputs OEM format, shows strange character encodings for unsupported utf8 characters in file names


ember91:
Describe the bug
----------------

Scanning a file with unicode characters in its name outputs the name in a strange encoding that neither is UTF-8 nor the system encoding.

How to reproduce the problem
----------------------------

This was reproduced on a fresh Windows installation of `clamav-1.4.2.win.x64.msi` with system encoding CP1252.

I created an empty file called `file_öταБЬℓσ.txt`.

I ran the following Python script in the same directory as the test file I created. Ensure your editor is set to UTF-8. Replace the path to `clamscan.exe`:

```python
import locale
import subprocess as sp

print('Preferred encoding:', locale.getpreferredencoding())
result = sp.run([r'C:\Program Files\ClamAV\clamscan.exe', 'file_öταБЬℓσ.txt'], stdout=sp.PIPE, stderr=sp.PIPE)
print(result.stdout)
```

Which outputs (I cut some of the output short with "..."):

```text
Preferred encoding: cp1252
b'C:\\Users\\User\\clamav_unicode_test\\file_\x94\xe7\xe0??l\xe5.txt: ...'
```

The output text is encoded as CP437, with some characters escaped with "?". Try yourself in Python with `b'\x94\xe7\xe0??l\xe5.txt'.decode('cp437')`. I have no idea why. I expected it to output it either in UTF-8, which would be for the best, or CP1252 which is the system encoding of my Windows installation.

In the [documentation](https://docs.clamav.net/manual/Usage/Scanning.html) it says that "As a side note, console output (stdin and stderr) will always be OEM encoded, even when redirected to a file.". 

Output of PowerShell `[System.Text.Encoding]::Default`:

```text
IsSingleByte      : True
BodyName          : iso-8859-1
EncodingName      : Western European (Windows)
HeaderName        : Windows-1252
WebName           : Windows-1252
WindowsCodePage   : 1252
IsBrowserDisplay  : True
IsBrowserSave     : True
IsMailNewsDisplay : True
IsMailNewsSave    : True
EncoderFallback   : System.Text.InternalEncoderBestFitFallback
DecoderFallback   : System.Text.InternalDecoderBestFitFallback
IsReadOnly        : True
CodePage          : 1252
```

Output from `clamconf.exe -n`:

```text
Checking configuration files in C:\Program Files\ClamAV

Config file: clamd.conf
-----------------------
[...]

PR #1461, opened 2025-02-27
On Windows, use CP_UTF8 over CP_OEMCP for output

ember91:
CP_ACP makes more sense than CP_OEMCP. It is more common that applications use ACP over OEM.

Tested by inputting a file with unicode characters in its name. Now it is presented in the ACP. Also ran `ctest -C RelWithDebInfo` but I'm not sure if it tests those kinds of things.

An alternative would be to encode it as UTF-8. Then all unicode characters could be represented.

### DIFF
--- a/clambc/bcrun.c
+++ b/clambc/bcrun.c
@@ -262,6 +262,10 @@ int main(int argc, char *argv[])
     int fd = -1;
     unsigned tracelevel;
 
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     if (check_flevel())
         exit(1);
 
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -177,7 +177,9 @@ int main(int argc, char **argv)
     if (check_flevel())
         exit(1);
 
-#ifndef _WIN32
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#else /* !_WIN32 */
     memset(&sa, 0, sizeof(sa));
     sa.sa_handler = SIG_IGN;
     sigaction(SIGHUP, &sa, NULL);
--- a/clamdscan/clamdscan.c
+++ b/clamdscan/clamdscan.c
@@ -68,7 +68,9 @@ int main(int argc, char **argv)
     struct optstruct *opts;
     const struct optstruct *opt;
     char buffer[26];
-#ifndef _WIN32
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#else /* !_WIN32 */
     struct sigaction sigact;
 #endif
 
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -1577,6 +1577,10 @@ int main(int argc, char *argv[])
     struct timeval tv_last, tv;
     unsigned i;
 
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     atexit(cleanup);
     setup_connections(argc, argv);
     init_ncurses(global.num_clamd, default_colors);
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -68,7 +68,9 @@ int main(int argc, char **argv)
     time_t date_start, date_end;
 
     char buffer[26];
-#ifndef _WIN32
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#else /* !_WIN32 */
     sigset_t sigset;
 #endif
     struct optstruct *opts;
--- a/clamsubmit/clamsubmit.c
+++ b/clamsubmit/clamsubmit.c
@@ -204,6 +204,10 @@ int main(int argc, char *argv[])
     char *url_for_auth_token;
     char *url_for_presigned_cookie;
 
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     curl_global_init(CURL_GLOBAL_ALL);
 
     clam_curl = curl_easy_init();
--- a/common/output.c
+++ b/common/output.c
@@ -505,35 +505,6 @@ void mprintf(loglevel_t loglevel, const char *str, ...)
     va_end(args);
     buff[len - 1] = 0;
 
-#ifdef _WIN32
-    do {
-        int tmplen    = len + 1;
-        wchar_t *tmpw = malloc(tmplen * sizeof(wchar_t));
-        char *nubuff;
-        if (!tmpw)
-            break;
-        if (!MultiByteToWideChar(CP_UTF8, MB_ERR_INVALID_CHARS, buff, -1, tmpw, tmplen)) {
-            free(tmpw);
-            break;
-        }
-        /* FIXME CHECK IT'S REALLY UTF8 */
-        nubuff = (char *)malloc(tmplen);
-        if (!nubuff) {
-            free(tmpw);
-            break;
-        }
-        if (!WideCharToMultiByte(CP_OEMCP, 0, tmpw, -1, nubuff, tmplen, NULL, NULL)) {
-            free(nubuff);
-            free(tmpw);
-            break;
-        }
-        free(tmpw);
-        if (len > sizeof(buffer))
-            free(abuffer);
-        abuffer = buff = nubuff;
-        len            = sizeof(buffer) + 1;
-    } while (0);
-#endif
     if (loglevel == LOGG_ERROR) {
         if (!mprintf_stdout)
             fd = stderr;
--- a/sigtool/sigtool.c
+++ b/sigtool/sigtool.c
@@ -4132,6 +4132,10 @@ int main(int argc, char **argv)
     const char *cvdcertsdir = NULL;
     STATBUF statbuf;
 
+#ifdef _WIN32
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     if (check_flevel())
         exit(1);
 


val-ms:
Sorry I haven't looked at this yet. I will review it soon as the 1.5 beta is out and pressure is off. If all goes well I'd love to slip this in between the beta and stable release. Thank you for putting in the PR.

val-ms:
Augh I forgot to comment on this after doing some testing. I compiled it and compared this PR versus the latest release using the script suggested in https://github.com/Cisco-Talos/clamav/issues/1440:

Here's running it with 1.4.2:
```
>>> result = sp.run([r'C:\\Users\\micasnyd\\clams\\clamav-1.4.2.win.x64\\clamscan.exe', '-r', '-d', r'C:\Users\micasnyd\clamav.hdb', r'.\f
>>> print(result.stdout)
b'C:\\Users\\micasnyd\\dev\\clamav\\clamav-3\\build\\file_\x94\xe7\xe0??l\xe5.txt: OK\r\n\r\n----------- SCAN SUMMARY -----------\r\nKnown viruses: 1\r\nEngine version: 1.4.2\r\nScanned directories: 0\r\nScanned files: 1\r\nInfected files: 0\r\nData scanned: 0.00 MB\r\nData read: 0.00 MB (ratio 0.00:1)\r\nTime: 0.050 sec (0 m 0 s)\r\nStart Date: 2025:05:20 17:10:34\r\nEnd Date:   2025:05:20 17:10:34\r\n'
```
I can decode using 'cp437' to get something slightly better:
```
>>> result.stdout.decode('cp437')
'C:\\Users\\micasnyd\\dev\\clamav\\clamav-3\\build\\file_öτα??lσ.txt: OK\r\n\r\n----------- SCAN SUMMARY -----------\r\nKnown viruses: 1\r\nEngine version: 1.4.2\r\nScanned directories: 0\r\nScanned files: 1\r\nInfected files: 0\r\nData scanned: 0.00 MB\r\nData read: 0.00 MB (ratio 0.00:1)\r\nTime: 0.050 sec (0 m 0 s)\r\nStart Date: 2025:05:20 17:10:34\r\nEnd Date:   2025:05:20 17:10:34\r\n'            
```

Here's running it with this PR:
```
>>> result = sp.run([r'.\install\clamscan.exe', '-r', '-d', r'C:\Users\micasnyd\clamav.hdb', r'.\file_öταБЬℓσ.txt'], stdout=sp.PIPE, stderr=sp.PIPE)
>>> print(result.stdout)
b'C:\\Users\\micasnyd\\dev\\clamav\\clamav-3\\build\\file_\xf6ta??ls.txt: OK\r\n\r\n----------- SCAN SUMMARY -----------\r\nKnown viruses: 1\r\nEngine version: 1.5.0-beta\r\nScanned directories: 0\r\nScanned files: 1\r\nInfected files: 0\r\nData scanned: 0.00 MB\r\nData read: 0.00 MB (ratio 0.00:1)\r\nTime: 0.078 sec (0 m 0 s)\r\nStart Date: 2025:05:20 17:12:00\r\nEnd Date:   2025:05:20 17:12:00\r\n'
```
I can decode using 'cp1252' to get something slightly better:
```
>>> result.stdout.decode('cp1252')
'C:\\Users\\micasnyd\\dev\\clamav\\clamav-3\\build\\file_öta??ls.txt: OK\r\n\r\n----------- SCAN SUMMARY -----------\r\nKnown viruses: 1\r\nEngine version: 1.5.0-beta\r\nScanned directories: 0\r\nScanned files: 1\r\nInfected files: 0\r\nData scanned: 0.00 MB\r\nData read: 0.00 MB (ratio 0.00:1)\r\nTime: 0.078 sec (0 m 0 s)\r\nStart Date: 2025:05:20 17:12:00\r\nEnd Date:   2025:05:20 17:12:00\r\n'
```

I just don't see this as being much better.  Am I missing something or misunderstanding something?

Something worth noting, ClamAV uses Rust under the hood in latest versions for log messages and they do a much better job with unicode. If I run this with `--debug` to see those logs:
```
❯ ~\clams\clamav-1.3.1.win.x64\clamscan.exe -d C:\Users\micasnyd\clamav.hdb .\file_öταБЬℓσ.txt --debug
```
I find this:
```
LibClamAV debug: Checking realpath of .\file_öταБЬℓσ.txt
LibClamAV debug: cli_get_filepath_from_handle: File path for handle 000000000000071C is: C:\Users\micasnyd\dev\clamav\clamav-3\build\file_öταБЬℓσ.txt
LibClamAV debug: Recognized ASCII text
LibClamAV debug: clean_cache_check: 1d074f253ef1505995e3318300958125 is negative
LibClamAV debug: Descriptor[3]: Continuing after file scan resulted with: No viruses detected
LibClamAV debug: hashtab: Freeing hashset, elements: 0, capacity: 0
LibClamAV debug: Descriptor[3]: Continuing after file scan resulted with: No viruses detected
LibClamAV debug: in cli_scanscript()
LibClamAV debug: hashtab: Freeing hashset, elements: 0, capacity: 0
LibClamAV debug: Descriptor[3]: Continuing after file scan resulted with: No viruses detected
LibClamAV debug: cli_magic_scan: returning 0  at line 5037
LibClamAV debug: clean_cache_add: 1d074f253ef1505995e3318300958125 (level 0)
LibClamAV debug: Descriptor[3]: Continuing after file scan resulted with: No viruses detected
C:\Users\micasnyd\dev\clamav\clamav-3\build\file_öτα??lσ.txt: OK
```
As you can see, whenever the filename is printed in the debug logs, it looks great. 

For logs, we're printing to stderr through Rust with this function: https://github.com/Cisco-Talos/clamav/blob/main/libclamav_rust/src/logging.rs#L88-L103

Where this is called:
- https://github.com/Cisco-Talos/clamav/blob/8ee78f8e79e1afe68d7dc90b1bf7eee4f861e8ff/libclamav/others_common.c#L126-L189

Maybe we should use Rust to print the application logs as well. It would at least look nice. I don't know if I'm otherwise missing the point with the codepage concern.

ember91:
Yes I don't claim that the result will be much better because it supports more characters; both will with your ACP and OEM encodings support the same number of characters (256). I claim that using ACP over OEM is better since ACP is more common than the OEM encoding and that the user more often expects ACP over OEM.

If we want to be able to represent all characters we can display them as UTF-8 instead. I understand that we can just show the rust logs and that has some merit, but in this case we can just change the constant I changed to `CP_UTF8` over `CP_ACP` or `CP_OEM` as I discusses before. That comes with some risks but not that many to be fair. 

A minor thing is that we can also see in this case that it does some approximations of characters. For example it seems to use 't' over 'τ' (greek small letter tau). That can be turned off in the translation function.

val-ms:
I'm sorry for the lag time responding to you. Personally I feel like being able to render all characters is ideal. So I lean towards UTF8. If we need to document somewhere that the next clamav version outputs UTF8 so that someone parsing the output can correctly decode it, then I think we should do that. 

However, I just tested changing the PR code to be like this:
```
diff --git a/common/output.c b/common/output.c
index 27157cd4f..c0112f945 100644
--- a/common/output.c
+++ b/common/output.c
@@ -522,7 +522,7 @@ void mprintf(loglevel_t loglevel, const char *str, ...)
             free(tmpw);
             break;
         }
-        if (!WideCharToMultiByte(CP_ACP, 0, tmpw, -1, nubuff, tmplen, NULL, NULL)) {
+        if (!WideCharToMultiByte(CP_UTF8, 0, tmpw, -1, nubuff, tmplen, NULL, NULL)) {
             free(nubuff);
             free(tmpw);
             break;
```

And the output is like this:
```
❯ .\install\clamscan.exe .\file_öταБЬℓσ.txt
Compiling:   0s, ETA:   0s [========================>]       10/10 tasks

C:\Users\micasnyd\dev\clamav\clamav-2\build\file_├╢╧ä╬▒╨æ╨¼Γäô╧â.txt: OK
```

I'm not really sure how to do it correctly in Win32 C.

ember91:
Try adding 'SetConsoleOutputCP(CP_UTF8);' at the initialization of the application. Although note that it's Windows-only, so hide it behind a define.

val-ms:
@ember91 Wow that seems to have done it. Thanks!

I applied this change on top of your current PR:
```patch
diff --git a/clamd/clamd.c b/clamd/clamd.c
index 4feeee82b..f70b3fb51 100644
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -177,7 +177,9 @@ int main(int argc, char **argv)
     if (check_flevel())
         exit(1);

-#ifndef _WIN32
+#if defined(_WIN32)
+    SetConsoleOutputCP(CP_UTF8);
+#else /* !_WIN32 */
     memset(&sa, 0, sizeof(sa));
     sa.sa_handler = SIG_IGN;
     sigaction(SIGHUP, &sa, NULL);
diff --git a/clamdscan/clamdscan.c b/clamdscan/clamdscan.c
index 761b43c09..2859afbaa 100644
--- a/clamdscan/clamdscan.c
+++ b/clamdscan/clamdscan.c
@@ -72,6 +72,10 @@ int main(int argc, char **argv)
     struct sigaction sigact;
 #endif

+#if defined(_WIN32)
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     if ((opts = optparse(NULL, argc, argv, 1, OPT_CLAMDSCAN, OPT_CLAMSCAN, NULL)) == NULL) {
         mprintf(LOGG_ERROR, "Can't parse command line options\n");
         exit(2);
diff --git a/clamdtop/clamdtop.c b/clamdtop/clamdtop.c
index 9721ed649..08d067a41 100644
--- a/clamdtop/clamdtop.c
+++ b/clamdtop/clamdtop.c
@@ -1549,6 +1549,10 @@ int main(int argc, char *argv[])
     struct timeval tv_last, tv;
     unsigned i;

+#if defined(_WIN32)
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     atexit(cleanup);
     setup_connections(argc, argv);
     init_ncurses(global.num_clamd, default_colors);
diff --git a/clamscan/clamscan.c b/clamscan/clamscan.c
index 00402c751..417742661 100644
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -77,7 +77,9 @@ int main(int argc, char **argv)
     if (check_flevel())
         exit(2);

-#if !defined(_WIN32)
+#if defined(_WIN32)
+    SetConsoleOutputCP(CP_UTF8);
+#else /* !_WIN32 */
     if (!setlocale(LC_CTYPE, "")) {
         mprintf(LOGG_WARNING, "Failed to set locale\n");
     }
diff --git a/common/output.c b/common/output.c
index 27157cd4f..c0112f945 100644
--- a/common/output.c
+++ b/common/output.c
@@ -522,7 +522,7 @@ void mprintf(loglevel_t loglevel, const char *str, ...)
             free(tmpw);
             break;
         }
-        if (!WideCharToMultiByte(CP_ACP, 0, tmpw, -1, nubuff, tmplen, NULL, NULL)) {
+        if (!WideCharToMultiByte(CP_UTF8, 0, tmpw, -1, nubuff, tmplen, NULL, NULL)) {
             free(nubuff);
             free(tmpw);
             break;
diff --git a/sigtool/sigtool.c b/sigtool/sigtool.c
index 7b24300b9..be7a60da4 100644
--- a/sigtool/sigtool.c
+++ b/sigtool/sigtool.c
@@ -4122,6 +4122,10 @@ int main(int argc, char **argv)
     struct optstruct *opts;
     STATBUF sb;

+#if defined(_WIN32)
+    SetConsoleOutputCP(CP_UTF8);
+#endif
+
     if (check_flevel())
         exit(1);
```

A scan of the file now looks like this:
```
❯ .\install\clamscan.exe -r -d C:\Users\micasnyd\clamav.hdb .\file_öταБЬℓσ.txt
Compiling:   0s, ETA:   0s [========================>]       10/10 tasks

C:\Users\micasnyd\dev\clamav\clamav-2\build\file_öταБЬℓσ.txt: OK

----------- SCAN SUMMARY -----------
Known viruses: 1
Engine version: 1.5.0-beta
Scanned directories: 0
Scanned files: 1
Infected files: 0
Data scanned: 0.00 MB
Data read: 0.00 MB (ratio 0.00:1)
Time: 0.022 sec (0 m 0 s)
Start Date: 2025:05:30 14:21:39
End Date:   2025:05:30 14:21:39
```

I confirmed the same from both clamdscan.exe and clamd.exe.

clamd.exe:
```
❯ .\install\clamd.exe
Limits: Global time limit set to 120000 milliseconds.
Limits: Global size limit set to 419430400 bytes.
Limits: File size limit set to 104857600 bytes.
Limits: Recursion level limit set to 17.
Limits: Files limit set to 10000.
Limits: MaxEmbeddedPE limit set to 41943040 bytes.
Limits: MaxHTMLNormalize limit set to 41943040 bytes.
Limits: MaxHTMLNoTags limit set to 8388608 bytes.
Limits: MaxScriptNormalize limit set to 20971520 bytes.
Limits: MaxZipTypeRcg limit set to 1048576 bytes.
Limits: MaxPartitions limit set to 50.
Limits: MaxIconsPE limit set to 100.
Limits: MaxRecHWP3 limit set to 16.
Limits: PCREMatchLimit limit set to 100000.
Limits: PCRERecMatchLimit limit set to 2000.
Limits: PCREMaxFileSize limit set to 104857600.
Archive support enabled.
Image (graphics) scanning support enabled.
Detection using image fuzzy hash enabled.
AlertExceedsMax heuristic detection disabled.
Heuristic alerts enabled.
Portable Executable support enabled.
ELF support enabled.
Mail files support enabled.
OLE2 support enabled.
PDF support enabled.
SWF support enabled.
HTML support enabled.
XMLDOCS support enabled.
HWP3 support enabled.
OneNote support enabled.
Self checking every 600 seconds.
Listening daemon: PID: 49948
MaxQueue set to: 100
C:\Users\micasnyd\dev\clamav\clamav-2\build\file_öταБЬℓσ.txt: OK
```

clamdscan.exe:
```
❯ .\install\clamdscan.exe .\file_öταБЬℓσ.txt
C:\Users\micasnyd\dev\clamav\clamav-2\build\file_öταБЬℓσ.txt: OK

----------- SCAN SUMMARY -----------
Infected files: 0
Time: 0.025 sec (0 m 0 s)
Start Date: 2025:05:30 14:25:47
End Date:   2025:05:30 14:25:47
```

Unfortunately, it didn't work for `clamdtop.exe`. I suspect that's because our pdcurses library dependency doesn't support utf8. 
image
We're presently using https://github.com/wmcbrine/PDCurses
So I guess the change to `clamdtop.c` is not required. 

What do you think of this change otherwise?

ember91:
If you have interactive terminal input you need `SetConsoleCP` as well. But I'm guessing you don't.

To me the changes look fine.

With just a quick read it seems like pdcurses can be built to support UTF-8 but no idea if you're building it yourselves.

ember91:
I can make the changes.

Three things:
* I can't find `clambc.c` you mentioned but I think I have added it in the files that matter that have a `main()`
* I haven't found where to put `UTF8=Y`. Is it `cmake/FindCURSES.cmake`?
* In `common/output.c` lines 508-536 it now looks at a glance like we're converting from UTF-8 to UTF-16 and then back to UTF-8, so I guess the whole `#ifdef _WIN32` block could be removed?

val-ms:
> * I can't find `clambc.c` you mentioned but I think I have added it in the files that matter that have a `main()`

Re clambc - that's my mistake. The [main function is in "bcrun.c"](https://github.com/Cisco-Talos/clamav/blob/main/clambc/bcrun.c#L267). 

> * I haven't found where to put `UTF8=Y`. Is it `cmake/FindCURSES.cmake`?

We build pdcurses and other third party C/C++ library dependencies separately using Mussels. I have already added a commit with the UTF8 change to my preexisting PR updating some of those to newer versions: https://github.com/Cisco-Talos/clamav-mussels-cookbook/pull/66/commits

> * In `common/output.c` lines 508-536 it now looks at a glance like we're converting from UTF-8 to UTF-16 and then back to UTF-8, so I guess the whole `#ifdef _WIN32` block could be removed?

Hah you're right. I don't see any value in converting from utf8 to utf16 and back either. Deleting that entire `#ifdef _WIN32` block seems like the right move.

ember91:
It's been updated

Fi418:
For Windows 10 1093 (and above) we can set the process code page using a manifest as part of the build process.

https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
  <assemblyIdentity type="win32" name="..." version="6.0.0.0"/>
  <application>
    <windowsSettings>
      <activeCodePage xmlns="http://schemas.microsoft.com/SMI/2019/WindowsSettings">UTF-8</activeCodePage>
    </windowsSettings>
  </application>
</assembly>
```

I used the 1.5.0 prebuilt binaries and applied the manifest per the documentation: `mt.exe -manifest manifest.xml -outputresource:clamscan.exe;#1`.

```
b'D:\\source\\clamav-1.5.0-beta.win.x64\\file_\x94\xe7\xe0??l\xe5.txt: Empty file
```

```
b'D:\\source\\clamav-1.5.0-beta.win.x64\\file_\xc3\xb6\xcf\x84\xce\xb1\xd0\x91\xd0\xac\xe2\x84\x93\xcf\x83.txt: Empty file
```

This could be a simpler solution, but it's not fully tested.

Sophie0x2E (approved):

val-ms:
Just manually rebased to fix a merge conflict in sigtool and then force-pushed.

PR #1462, opened 2025-02-27
Clam 2740 devcontainer template

jhumlick:
POC showing two devcontainers - one for Debian and another for AlmaLinux.

The AlmaLinux container fails one unit test, which is documented in the Dockerfile for it. This will need to be investigated later.


val-ms (changes requested):
Tried it again and had good luck using this in vscode from my WSL2 environment. However, I needed to make these changes to get debugging working.

jhumlick:
I have pushed all of the suggested changes.

val-ms (dismissed):

rsundriyal (dismissed):
Looks good. I have tested both on the MacOS machine and it works fine.

.devcontainer/clamav-almalinux/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+COPY . /src/
+
+ENV CARGO_HOME=/src/build
+
+ENV HOME /home/${REMOTE_USER}
+
+# Install and enable EPEL and CRB repositories
+RUN dnf -y install epel-release && dnf config-manager --set-enabled crb
+
```
val-ms, line 19:
```suggestion

# Declare container args
ARG REMOTE_USER
ARG REMOTE_UID

# Set working directory
WORKDIR /src

# Copy current directory contents into the container
COPY . /src/

# Set environment variables
ENV CARGO_HOME=/src/build
ENV HOME=/home/${REMOTE_USER}

# Install and enable EPEL and CRB repositories
RUN dnf -y install epel-release && dnf config-manager --set-enabled crb

# Install necessary packages
```
Changes so the structure and comments are as identical as possible to the debian one.

.devcontainer/clamav-debian/Dockerfile
```diff
@@ -0,0 +1,105 @@
+
+# Use a slim Debian image
+FROM index.docker.io/library/debian:12-slim AS clamav-debian-devcontainer-macos
+
+# Declare args again after FROM to make them available in this build stage
```
val-ms, line 5:
```suggestion
# Declare container args
```

Looks to me like this comment was copied from a multi-stage dockerfile.

.devcontainer/clamav-debian/devcontainer.json
val-ms, line 1: Some of the comments in this file do not mirror those int he clamav-almalinux equivalent. As far as I can tell, they could be tidied up to be nearly identical.

val-ms (changes requested):

.devcontainer/clamav-almalinux/devcontainer.json
```diff
@@ -0,0 +1,48 @@
...
+    ],
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "rust-lang.rust-analyzer",
+                "ms-azuretools.vscode-docker",
+                "eamodio.gitlens",
+                "vadimcn.vscode-lldb",
+                "webfreak.debug",
+                "ms-vscode-remote.remote-containers"
```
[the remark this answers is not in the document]
val-ms: The last commit removed these (by mistake?)

val-ms (changes requested):

.devcontainer/clamav-almalinux/Dockerfile
```diff
@@ -0,0 +1,116 @@
...
+    -e "s|^\#\(ScriptedUpdates\).*|\1 yes|" \
+    /clamav/etc/clamav/freshclam.conf.sample > /clamav/etc/clamav/freshclam.conf \
+    && sed -e "s|^\(Example\)|# \1|" \
+    -e "s|.*\(MilterSocket\) .*|\1 inet:7357|" \
+    -e "s|.*\(User\) .*|\1 clamav|" \
+    -e "s|^\#\(LogFile\) .*|\1 /var/log/clamav/milter.log|" \
+    -e "s|^\#\(LogTime\).*|\1 yes|" \
+    -e "s|.*\(ClamdSocket\) .*|\1 unix:/tmp/clamd.sock|" \
+    /clamav/etc/clamav/clamav-milter.conf.sample > /clamav/etc/clamav/clamav-milter.conf
+# && ctest -V --timeout 3000
```
val-ms, line 110: For me, ctest also fails on Debian, though only the `clamd` ones and with this error:
```
4: ERROR: Wed Aug  6 18:14:15 2025 -> TCP: Cannot bind to [localhost]:3319: Address already in use
4: ERROR: Wed Aug  6 18:14:15 2025 -> Cannot set socket permission for clamd-test.socket to 666
```

Unsure what that's all about. Weirdly, AlmaLinux devcontainer ones fail all over the place. Especially the valgrind ones.  Very strange. 

I'm not really sure what I'm saying here.  I'm just as happy not having it build clamav at all as I am having it run the tests.  So I guess I'm happy with this even though it is different from the Debian one.

val-ms (changes requested):

.devcontainer/clamav-almalinux/Dockerfile
val-ms: To be honest, I don't think we should be building clamav in the dockerfile for the dev container. I think the the point of a dev container is to provide a development environment so you can do the build as desired. I'm inclined to remove the clamav build steps entirely from both Dockerfiles.

val-ms (approved):
Thanks for your work and for letting me be fussy with this review. I'm good with it now.

PR #1463, opened 2025-03-02
 updated unzip.c

Arian104:
Added the index_local_file_headers_within_bounds function to index local file headers between two file offsets.
Added the index_local_file_headers function to add files not present in the central directory to the catalogue.
Modified the cli_unzip function to handle local file headers not referenced by the central directory after calling index_central_dir.

### DIFF
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -1167,11 +1167,411 @@ done:
     return status;
 }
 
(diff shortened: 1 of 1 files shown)


val-ms:
This looks like you are attempting to claim credit for @frsell1's prior pull request https://github.com/Cisco-Talos/clamav/pull/1460.  I plan to test and review that PR later this week.

Issue #706, opened 2022-09-23
Broken N/A `MEMSTATS` metrics when using `STATS` command


quantranhong1999:
Describe the bug
----------------

Can not get proper `MEMSTATS` metrics when using `clamd` STATS command.

How to reproduce the problem
----------------------------

On upper clamav version 0.104.x, send `STATS` command to clamav official container (port 3310) we will get a response with broken N/A `MEMSTATS`:
```
POOLS: 1

STATE: VALID PRIMARY
THREADS: live 1  idle 0 max 10 idle-timeout 30
QUEUE: 0 items
	STATS 0.000076 
Error processing command. ERROR 
MEMSTATS: heap N/A mmap N/A used N/A free N/A releasable N/A pools 1 pools_used 1227.795M pools_total 1227.847M
END 
``` 

The previous 0.102.4 version response was OK:
```
POOLS: 1

STATE: VALID PRIMARY
THREADS: live 1  idle 0 max 12 idle-timeout 30
QUEUE: 0 items
	STATS 0.000069 

MEMSTATS: heap 10.145M mmap 0.129M used 3.232M free 6.913M releasable 0.126M pools 1 pools_used 1051.970M pools_total 1052.007M
END 
```

Attachments
-----------

shaileshlaghane:
Facing the same issue with v1.3.0, any workaround ?

jullianow:
Facing the same problem.
Any solution?

mailsonpeixe:
Facing the same issue on version 1.4.1.

moreirajo:
Same issue on 1.4.2, deployed in kubernetes

jullianow:
@val-ms Can we have your attention on this?

We are facing issues with ClamAV monitoring that make it difficult to analyze how to improve the service's memory consumption, as the STATS command is failing.

Thank you.

jullianow:
@val-ms thanks so much.

jullianow:
@val-ms Any plans to release a version of the docker image with this fix?

val-ms:
@jullianow not until 1.5.0 is published. Unless you want to play with an unstable image.

jullianow:
@val-ms Any forecast of when version 1.5.0 will be released? Is there any roadmap where can I follow the progress of this version?
Thank you for your attention

val-ms:
@jullianow Sorry we do not have a public roadmap. 1.5.0 is blocked by two high priority issues:
[...]

PR #1465, opened 2025-03-08
clamd: Fix bug reporting memory stats, used by clamdtop

jhumlick:
ClamD's STATS API reports process memory stats on systems that
provide the `mallinfo()` system call.
This feature is used by ClamDTOP to show process memory usage.
When we switched to the CMake build system, we neglected to add the
check for the `mallinfo()` system call and so broke ClamD memory
usage reporting.

This commit adds the CMake check for `mallinfo()` and sets
HAVE_MALLINFO, if found.

Fixes: https://github.com/Cisco-Talos/clamav/issues/706

Jira: CLAM-2742

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -778,6 +778,8 @@ include(CheckFTS)
 include(CheckUnamePosix)
 # Check support for file descriptor passing
 include(CheckFDPassing)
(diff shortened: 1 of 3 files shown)


jhumlick:
I have updated the commit message.

val-ms (approved):

PR #1470, opened 2025-03-19
Windows: code quality improvement for --move and --remove options

val-ms:
When the --move or --remove options are used, ClamAV carefully traverses the file path one layer at a time so as to avoid following a directory that is a symlink or reparse point.
We do this for directories, but could also do it for files. Only an admin should be able to create a reparse point for a file, but it is better to be consistent.

Thank you to Maxim Suhanov for reporting this issue.


TheRaynMan (approved):
Known option change, looks good to me.

val-ms:
Waiting on internal test pipelines. Seems like a firewall issue or something to our win 11 arm64 test node. The move/remove/copy action tests worked fine though for win 10 x64, win 10 x86, and win 8 x64.

val-ms:
Test pipelines looking okay now.

PR #1486, opened 2025-04-14
Fix stack-buffer-overflow in parse_regex due to missing bounds checks

Shivam7-1:
issue link: https://issues.oss-fuzz.com/issues/388922799 
PR description
parse_regex iterates over the input pattern using the *last index but fails to enforce bounds checking before accessing p[*last]. When passed malformed or fuzzed input, this can lead to out-of-bounds reads, causing a crash or undefined behavior.

Fix:
Added strict (*last < pSize) bounds checking to:
-- The main loop condition in parse_regex
-- All cases where p[*last] is accessed (including switch statements and conditions)
Ensured safe handling of escape sequences, character classes, and alternates.
Added early exit if *last exceeds pSize to prevent further processing of invalid input.

### DIFF
--- a/libclamav/regex_suffix.c
+++ b/libclamav/regex_suffix.c
@@ -275,7 +275,6 @@ static struct node *parse_regex(const uint8_t *p, const size_t pSize, size_t *la
     struct node *tmp;
 
     while (*last < pSize && p[*last] != '$' && p[*last] != '\0') {
-        if (*last >= pSize) break;  // Safe guard
         switch (p[*last]) {
             case '|':
                 ++*last;
--- a/libclamav/regex_suffix.c
+++ b/libclamav/regex_suffix.c
@@ -274,7 +274,8 @@ static struct node *parse_regex(const uint8_t *p, const size_t pSize, size_t *la
     struct node *right;
     struct node *tmp;
 
-    while (p[*last] != '$' && p[*last] != '\0') {
+    while (*last < pSize && p[*last] != '$' && p[*last] != '\0') {
+        if (*last >= pSize) break;  // Safe guard
         switch (p[*last]) {
             case '|':
                 ++*last;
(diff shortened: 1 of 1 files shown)


libclamav/regex_suffix.c
```diff
@@ -274,7 +274,8 @@ static struct node *parse_regex(const uint8_t *p, const size_t pSize, size_t *la
     struct node *right;
     struct node *tmp;
 
-    while (p[*last] != '$' && p[*last] != '\0') {
+    while (*last < pSize && p[*last] != '$' && p[*last] != '\0') {
+        if (*last >= pSize) break;  // Safe guard
```
val-ms, line 278: I don't think this safe guard is needed. The check in the `while()`-condition does the same thing at a better place.
```suggestion
```

val-ms (changes requested):
There's an improvement to be made just from reading the code. I still need to think about it harder and play with it.

libclamav/regex_suffix.c
Shivam7-1: Hii @val-ms Thanks for reviewing i had commited above suggested changes
Thanks

val-ms (approved):
I have been unable to reproduce the [oss-fuzz issue](https://issues.oss-fuzz.com/issues/388922799) using oss-fuzz.

E.g. I run `python3 infra/helper.py reproduce clamav clamav_dbload_PDB_fuzzer ~/Downloads/clusterfuzz-testcase-minimized-clamav_dbload_PDB_fuzzer-5501437374693376` and it doesn't trigger.

I did run into an unrelated memory leak introduced in 1.5 development that is triggering with all oss-fuzz runs and for which I need to fix to get oss-fuzz testing back up and running. So I made a fix for that here ( https://github.com/Cisco-Talos/clamav/pull/1489 ). But even on that branch, I couldn't reproduce this issue.

That said, I agree with your change.

PR #1489, opened 2025-04-21
Fix minor memory leak

val-ms:
The memory allocated for certs_directory in the ClamAV engine is not free'd when the engine is free'd. This isn't readily apparent when using a mempool because the mempool itself is free'd and the issue is masked.

Also: I found that the location in `cl_engine_free()` where we estimate the number of tasks incorrectly placed the fuzzy_hashmap free task in the block where we free the `test_root`, rather than up in the `engine->root` for-loop. Fixed.

Edit: I have also added a fix for a leak when multiple idb signatures are loaded. I discovered this while trying to come up with simple repro steps for anyone wishing to manually test this fix.

Repro steps on Linux, not Mac (requires `valgrind`):

1. Build clamav with the cmake option `-D DISABLE_MPOOL=ON`

2. Run `ctest -V`


libclamav/others.h
```diff
@@ -285,9 +285,9 @@ struct icomtr {
 
 struct icon_matcher {
     char **group_names[2];
-    unsigned int group_counts[2];
+    uint32_t group_counts[2];
```
jhumlick, line 288: I'm just curious as to why the change from unsigned int, which is almost always 32-bits, to uint32_t was made. Should I make similar changes when I edit older code?

libclamav/readdb.c
```diff
@@ -1524,30 +1541,56 @@ static cl_error_t cli_loadidb(FILE *fs, struct cl_engine *engine, unsigned int *
...
+        if (matcher->group_names[0]) {
+            for (i = 0; i < matcher->group_counts[0]; i++)
+                MPOOL_FREE(engine->mempool, matcher->group_names[0][i]);
+            MPOOL_FREE(engine->mempool, matcher->group_names[0]);
+        }
+        if (matcher->group_names[1]) {
+            for (i = 0; i < matcher->group_counts[1]; i++)
+                MPOOL_FREE(engine->mempool, matcher->group_names[1][i]);
+            MPOOL_FREE(engine->mempool, matcher->group_names[1]);
+        }
```
jhumlick, line 1588: If you need to make a minor change here, you're going to have to do it twice.

I'd write this as:

for (i = 0; i < 2; i++) {
    if (matcher->group_names[i]) {
        for (j = 0; j < matcher->group_counts[i]; j++)
            MPOOL_FREE(engine->mempool, matcher->group_names[i][j]);
        MPOOL_FREE(engine->mempool, matcher->group_names[i]);
    }
}

jhumlick (approved):

libclamav/others.h
val-ms: The lengths of older C types, such as `char`, `short`, `int`, `long`, vary by OS and arch. `int` is more reliably 32-bits than `long`, but both *may* vary by OS and architecture: https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models

I habitually switch `unsigned`, `int`, and `long` types to `stdint.h` types like `uint32_t` and `uint64_t`. For data, `uint8_t` is better than `unsigned char`. For strings, `char` is fine. `int8_t` might technically be better to represent data underlying utf8. utf8 is hard in C, and ClamAV code generally does a bad job with it. 

Weirdly enough, `off_t` is also poorly defined. It's still everywhere and ought to be replaced with `ssize_t` if negative values *need* to be supported, or `size_t` otherwise. Most clamav offsets are relative to the start of the file so negative is almost never required. Disclaimer: I've introduced bugs by converting types like this from signed to unsigned, so you have to be really careful when doing that. 

On that topic, sentinel values for indicating an error are a code smell. Ideally we'd never use them and would always return an error code (e.g. using `cl_error_t`) and then pass the back values through function parameters. It's just considerable effort to rewrite to do it that way with existing code, and there is risk we'd introduce a bug by switching.

jhumlick (dismissed):

jhumlick (approved):

jhumlick (approved):

PR #1491, opened 2025-04-30
CMake/Rust openssl detection fix

val-ms:
We observed build failures on Ubuntu 20.04 ARM64 because the Rust code saw extra OpenSSL dependencies in the OPENSSL_LIBS environment variable and was confused.

This change switches from using OPENSSL_LIBRARIES, which may have extra dependencies for libcrypto/libssl, to only use OPENSSL_CRYPTO_LIBRARY and OPENSSL_SSL_LIBRARY.


rsundriyal (approved):
Looks good. Tests worked fine.

PR #1495, opened 2025-05-07
libclamav: Disable MyDoom heuristic by default.

jhumlick:
The MyDoom heuristic has a track record of false positives and detection should be handled via signatures instead.

CLAM-623


val-ms (approved):
Looks good!

PR #1513, opened 2025-05-30
Wait for the mockup server process to exit before moving on

### LINKED ISSUE #1512: Fix a race condition in freshclam_test.py
Describe the bug
----------------

Some tests in `freshclam_test.py` sporadically fail, because there is a race condition between killing the mockup server of a test and starting the next one.

How to reproduce the problem
----------------------------

Build ClamAV and run the test suite. Every once in a while one of the freshclam tests fails to start its mockup server with an "Address already in use" error, followed by a "Connection refused" from the respective client, because the old server has meanwhile terminated and the new one wasn't started. We (SUSE) saw it happen more often on s390x and ppc64le than on other architectures, and my guess is that it is more likely to happen under high load.

Solution
-----------

The testcase object should not only send an asynchronous `SIGTERM` to the server process, but actually wait for it to terminate before moving on, so that the next test can again listen on port 8001:

```
--- unit_tests/freshclam_test.py.orig
+++ unit_tests/freshclam_test.py
@@ -50,6 +50,7 @@ class TC(testcase.TestCase):
     def tearDown(self):
         if TC.mock_mirror != None:
             TC.mock_mirror.terminate()
+            TC.mock_mirror.join()
             TC.mock_mirror = None
 
         # Clear the database directory
```

Alternatively the test cases could use individual port numbers (ideally randomized), so that the server processes don't interfere with each other.

rma-x:
This fixes #1512, a race condition that triggers when the next server starts quicker than the previous one exits after signalling it asynchronously.

### DIFF
--- a/unit_tests/freshclam_test.py
+++ b/unit_tests/freshclam_test.py
@@ -50,6 +50,7 @@ class TC(testcase.TestCase):
     def tearDown(self):
         if TC.mock_mirror != None:
             TC.mock_mirror.terminate()
+            TC.mock_mirror.join()
             TC.mock_mirror = None
 
         # Clear the database directory


val-ms (approved):
Looks good in the linux/mac github workflows and on windows/linux/mac builds in jenkins. 
I didn't attempt to reproduce the race condition, but the change makes sense and didn't cause any unexpected problems. 

Thanks for the help @rma-x !

PR #1514, opened 2025-06-04
clamd: Add missing scan option checks for PDF and HTML URIs

val-ms:


clamd/server-th.c
```diff
@@ -1384,9 +1384,17 @@ int recvloop(int *socketds, unsigned nsockets, struct cl_engine *engine, unsigne
         options.dev |= CL_SCAN_DEV_COLLECT_SHA;
 #endif
 
-    /* JSON check to prevent engine loading if specified without libjson-c */
-    if (optget(opts, "GenerateMetadataJson")->enabled)
+    if (optget(opts, "GenerateMetadataJson")->enabled) {
         options.general |= CL_SCAN_GENERAL_COLLECT_METADATA;
+    }
+
+    if (optget(opts, "JsonStoreHTMLURIs")->enabled) {
```
jhumlick, line 1391: I almost think we should either give an error or enable CL_SCAN_GENERAL_COLLECT_METADATA here if that option is not set. I can imaging getting a bug report if the option to store HTML or PDF URI entries is enabled but nothing happens.

jhumlick (approved):

jhumlick (approved):

jhumlick (dismissed):

jhumlick (approved):

PR #1519, opened 2025-06-17
Fix double-extraction of OOXML-based office documents

val-ms:
File-type scanning for OOXML-based office documents results in double-extraction, because we also extract OOXML document components in the appopriate parser modules.

CLAM-1412


val-ms:
Never mind will put this change as a separate commit in https://github.com/val-ms/clamav/commits/CLAM-1859-sha256-cache/ to avoid merge conflicts.

PR #1520, opened 2025-06-17
Add win32_compat dependency to libraries using pthread.h

zaddach:
A couple of intermediate libraries (such as regex, bytecode_runtime, ...) use the `pthread.h` header. On Windows, it is made available through the `PThreadW32` library, and those libraries need to "link" against the `win32_compat` library to be able to see the header.


val-ms:
Just rebased and force-pushed to see this run through CI/CD.

val-ms (approved):
CI/CD (github and jenkins) look good.

PR #1522, opened 2025-06-18
Fix out of bounds read in UDF parser

val-ms:
A pointer representing file identifiers and file entries may be added to a list for later processing before validating the length of the data is within the given volume descriptor size.

The fix moves the size check to occur before adding it to the list.

Issue reported by volticks, @movx64 on Twitter working with Trend Micro Zero Day Initiative.

Note: This was also fixed in ClamAV 1.4.3.

### DIFF
--- a/libclamav/udf.c
+++ b/libclamav/udf.c
@@ -684,17 +684,21 @@ static cl_error_t findFileIdentifiers(const uint8_t *const input, PointerList *p
     size_t fidDescSize;
 
     while (FILE_IDENTIFIER_DESCRIPTOR == tagId) {
+        /* This is how far into the Volume we already are. */
+        bufUsed     = buffer - input;
+        fidDescSize = getFileIdentifierDescriptorSize((FileIdentifierDescriptor *)buffer);
 
-        /*Check that it's safe to read the header. */
+        /* Check that it's safe to save the file identifier pointer for later use */
+        if (VOLUME_DESCRIPTOR_SIZE < (fidDescSize + bufUsed)) {
+            break;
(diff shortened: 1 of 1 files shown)


bentasker:
@val-ms will this ultimately get fixed in 1.3 too?

val-ms:
@bentasker 1.3 is a non-LTS release and is past end of life. You'll have to upgrade to 1.4+ for the fix, or roll back to 1.0 LTS.

tj-cisco (approved):

bentasker:
@val-ms Oh!

The [support matrix](https://docs.clamav.net/faq/faq-eol.html#version-support-matrix) suggested it was still in support (expected EOL 1.5 + 4 months)

But, it [looks like](https://docs.clamav.net/faq/faq-eol.html#regular-non-lts-feature-releases) it's because it's more than 4 months since 1.4 released? 

(Sorry, not trying to hijack the thread - just intending to go and update [endoflife.date](https://endoflife.date/clamav) and want to make sure I've understood)

Thanks

val-ms:
> because it's more than 4 months since 1.4 released

@bentasker exactly. 

> Non-LTS feature releases will be supported with critical patch versions for at least four (4) months from the initial publication date of the next feature release or until the next-next feature release is published.

It has been more than 4 months since 1.4.0 was published, so 1.3 is past end of life. The policy is written this way to let us to rapidly publish non-LTS feature versions, if needed, without having to backport security patches for many versions.

bentasker:
Thank you very much!

Issue #1534, opened 2025-07-05
Malformed zip files cause infinity loop in unzip module (1.5.0 beta)


dmknght:
Describe the bug
----------------

There's an issue calculating value of variable `coff` (index of for loop) which causes infinity loop and high amount of RAM in unzip module. This bug is affecting `.zip` files and any file format that uses `zip` like document files (`.odp` for example).
Using `--debug` flag, an infinity loop can be seen

<img width="717" height="356" alt="Image" src="https://github.com/user-attachments/assets/9c2cf307-e796-4723-bbdd-20cdb65c11cb" />

Bug happened in the for loop of function `index_local_file_headers_within_bounds`

<img width="823" height="465" alt="Image" src="https://github.com/user-attachments/assets/27a34d3c-1631-4b8e-adcc-0c085f45ac9c" />

- for loop increases `coff`'s value by 1
- value of `coff` is added by return value of `parse_local_file_header` --> This function has a logical bug of calculating value that I'll explain bellow
- value of `coff` decreases by 1.

=> Anytime `parse_local_file_header` returns value **0**, value of variable `coff` is gonna be kept at `0`, causing infinity loop

<img width="672" height="1032" alt="Image" src="https://github.com/user-attachments/assets/598d4a5d-3e13-484a-8f4c-8eeef3d10b97" />

Inside function `parse_local_file_header`, value of `size_of_fileheader_and_data` which is used as **return value** is 0 by default.

When function detects error (branch `fname out of file` in this particular test), it goes to `done` label. The return value hasn't changed (it's still 0).

<img width="422" height="175" alt="Image" src="https://github.com/user-attachments/assets/4a2557f3-1262-4580-b0b4-c53ee71d46e0" />

=> Any error happens while parsing header is gonna to make infinity loop because label `done` doesn't change value.

Attachment file is a malformed `zip` file. Similar issue can be re-created using `radamsa` tool to generate malformed zip, document files.

[out-3.zip](https://github.com/user-attachments/files/21078912/out-3.zip)

Fi418:
I can reproduce the issue with other malformed zip files too. @val-ms you can assign this to me.

val-ms:
@Sophie0x2E thank you for volunteering to work on this! I was going to work on it today but I got distracted. 
If something comes up and you don't have time for it, please let me know and I'll pick it up.

PR #1539, opened 2025-07-08
Improve handling of malformed ZIP files

Sophie0x2E:
This addresses https://github.com/Cisco-Talos/clamav/issues/1534 and makes some minor improvements for maintainability when handling NULL terminated strings.

Solves loops when unzipping ZIP64 as well but only due to an overlapping cause.

### DIFF
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -660,8 +660,10 @@ static unsigned int parse_local_file_header(
     zip += LOCAL_HEADER_flen;
     zsize -= LOCAL_HEADER_flen;
 
-        cli_dbgmsg("cli_unzip: local header - ZMDNAME:%d:%.255s:%u:%u:%x:%u:%u:%u\n",
-               ((LOCAL_HEADER_flags & F_ENCR) != 0), name, LOCAL_HEADER_usize, LOCAL_HEADER_csize, LOCAL_HEADER_crc32, LOCAL_HEADER_method, file_count, ctx->recursion_level);
+    cli_dbgmsg("cli_unzip: local header - ZMDNAME:%d:%.255s:%u:%u:" /*%x:%u:*/ "%u:%u\n",
+               ((LOCAL_HEADER_flags & F_ENCR) != 0), name, usize, csize, file_count, ctx->recursion_level);
+    /* ZMDfmt virname:encrypted(0-1):filename(exact|*):usize(exact|*):csize(exact|*):crc32(exact|*):method(exact|*):fileno(exact|*):maxdepth(exact|*) */
+
     /* Scan file header metadata. */
     if (cli_matchmeta(ctx, name, LOCAL_HEADER_csize, LOCAL_HEADER_usize, (LOCAL_HEADER_flags & F_ENCR) != 0, file_count, LOCAL_HEADER_crc32) == CL_VIRUS) {
         *ret = CL_VIRUS;
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -776,6 +776,11 @@ done:
         free(original_filename);
     }
 
+    if (size_of_fileheader_and_data == 0) {
+        size_of_fileheader_and_data = zip - local_header;
+        *ret = CL_EPARSE;
+    }
+
     return size_of_fileheader_and_data;
 }
 
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -625,6 +625,8 @@ static unsigned int parse_local_file_header(
     uint32_t nsize  = 0;
     const char *src = NULL;
 
+    zip = local_header + SIZEOF_LOCAL_HEADER;
+
     if (!(local_header = fmap_need_off(map, loff, SIZEOF_LOCAL_HEADER))) {
         cli_dbgmsg("cli_unzip: local header - out of file\n");
         goto done;
@@ -638,7 +640,6 @@ static unsigned int parse_local_file_header(
         goto done;
     }
 
-    zip = local_header + SIZEOF_LOCAL_HEADER;
     zsize -= SIZEOF_LOCAL_HEADER;
 
     if (zsize <= LOCAL_HEADER_flen) {
@@ -775,6 +776,7 @@ done:
     if (NULL != original_filename) {
         free(original_filename);
     }
+
 
     if (size_of_fileheader_and_data == 0) {
         size_of_fileheader_and_data = zip - local_header;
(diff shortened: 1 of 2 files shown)


libclamav/unzip.c
```diff
@@ -641,30 +641,26 @@ static unsigned int parse_local_file_header(
...
     src   = fmap_need_ptr_once(map, zip, nsize);
     if (nsize && (NULL != src)) {
         memcpy(name, zip, nsize);
-        name[nsize] = '\0';
         if (CL_SUCCESS != cli_basename(name, nsize, &original_filename)) {
             original_filename = NULL;
         }
-    } else {
-        name[0] = '\0';
     }
```
val-ms, line 659: I really like these suggested changes. Thank you.

libclamav/unzip.c
```diff
@@ -780,6 +776,11 @@ static unsigned int parse_local_file_header(
         free(original_filename);
     }
 
+    if (size_of_fileheader_and_data == 0) {
+        size_of_fileheader_and_data = zip - local_header;
```
val-ms, line 780: Unfortunately we can't use `zip` after `done:` because it isn't set until after the first `goto done;`.

But I think it's fine if it returns `0` if we set `*ret = CL_EPARSE` as you suggested 👍. We just need to check for either `0` or, more importantly, `CL_EPARSE` in the calling functions.

I think this may be a better solution.  What do you think?

```diff
diff --git a/libclamav/unzip.c b/libclamav/unzip.c
index b7b0ba6a1..35a88979a 100644
--- a/libclamav/unzip.c
+++ b/libclamav/unzip.c
@@ -776,6 +776,10 @@ done:
         free(original_filename);
     }

+    if (size_of_fileheader_and_data == 0) {
+        *ret = CL_EPARSE;
+    }
+
     return size_of_fileheader_and_data;
 }

@@ -1245,8 +1249,7 @@ cl_error_t index_local_file_headers_within_bounds(
         if (!(ptr = fmap_need_off_once(map, coff, 4)))
             continue;
         if (cli_readint32(ptr) == ZIP_MAGIC_LOCAL_FILE_HEADER) {
-            // increment coff by the size of the found local file header + file data
-            coff += parse_local_file_header(
+            size_t file_record_size = parse_local_file_header(
                 map,
                 coff,
                 fsize - coff,
@@ -1259,13 +1262,16 @@ cl_error_t index_local_file_headers_within_bounds(
                 1,
                 NULL,
                 &(zip_catalogue[index]));
-            // decrement coff by 1 to account for the increment at the end of the loop
-            coff -= 1;

-            if (CL_EPARSE != ret) {
+            if (file_record_size != 0 && CL_EPARSE != ret) {
                 // Found a record.
+                cli_dbgmsg("cli_unzip: Found a record\n");
                 index++;
                 total_file_count++;
+
+                // increment coff by the size of the found local file header + file data
+                // but decrement by 1 to account for the increment at the end of the loop
+                coff += file_record_size - 1;
             }

             if (ret == CL_VIRUS) {
```

Sophie0x2E: The zip declaration can be moved, but your suggestion is cleaner. It does needs an else clause for CL_EPARSE which I added.

Latest commit works against the same test files.

libclamav/unzip.c
```diff
@@ -780,6 +780,16 @@ static unsigned int parse_local_file_header(
...
 
+    if (*ret == CL_EPARSE) {
+        cli_dbgmsg("cli_unzip: Skipping local header for CL_EPARSE\n");
+        if (zip && local_header) {
+            size_of_fileheader_and_data = zip - local_header;
+        } else {
+            // Shouldn't happen, but best to log if it does due to changes
+            cli_dbgmsg("cli_unzip: zip %p local_header %p\n", zip, local_header);
+        }
+    }
```
val-ms, line 791: I don't think this is necessary at all. The calling function should not use the return value if `*ret` is not `CL_SUCCESS`. So there is no obligation to try to make it be correct.

At least I think that's a decent plan.  Can you think of a reason why it should return `size_of_fileheader_and_data` if there was a parse error?

libclamav/unzip.c
```diff
@@ -641,31 +641,27 @@ static unsigned int parse_local_file_header(
     zip = local_header + SIZEOF_LOCAL_HEADER;
     zsize -= SIZEOF_LOCAL_HEADER;
 
-    memset(name, '\0', 256);
-
     if (zsize <= LOCAL_HEADER_flen) {
         cli_dbgmsg("cli_unzip: local header - fname out of file\n");
         fmap_unneed_off(map, loff, SIZEOF_LOCAL_HEADER);
         goto done;
```
val-ms, line 647: This probably needs `*ret = CL_VIRUS;` as well

val-ms (changes requested):
Oh Good call setting `*ret` in other error cases. It seems like we should set it to `CL_SUCCESS`  immediately before the `done:` label as well. 

I haven't tried it on my system. I'll give it a go on my system.

libclamav/unzip.c
Sophie0x2E: It's not necessary. I only added it in case a mistake is done on new additions to the function since not all features of ZIP archives are present.

It can be removed.

libclamav/unzip.c
```diff
@@ -757,13 +755,15 @@ static unsigned int parse_local_file_header(
     if (LOCAL_HEADER_flags & F_USEDD) {
         if (zsize < 12) {
             cli_dbgmsg("cli_unzip: local header - data desc out of file\n");
+            *ret = CL_EPARSE;
             goto done;
         }
         zsize -= 12;
         if (fmap_need_ptr_once(map, zip, 4)) {
-            if (cli_readint32(zip) == ZIP_MAGIC_FILE_BEGIN_SPLIT_OR_SPANNED) {
+            if (zip && cli_readint32(zip) == ZIP_MAGIC_FILE_BEGIN_SPLIT_OR_SPANNED) {
```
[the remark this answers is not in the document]
Sophie0x2E: I only did it to silence a warning, it can be removed.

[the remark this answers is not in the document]
val-ms: You could probably do like this to silence the warning as well :
```c
zip = fmap_need_ptr_once(map, zip, 4);
if (NULL != zip) {
    ...
```
The `fmap_need_ptr_once()` function returns `zip` after verifying that the data pointed to by zip of length 4 is within the file and is paged into memory (for a file backed memory map).

val-ms:
@Sophie0x2E The more I looked at the ZIP module the more I wanted to refactor. It didn't make sense to ask you to make so many changes, so I started a new branch and went to town on it: https://github.com/Cisco-Talos/clamav/pull/1548

Thank you so much for finding the initial issue and for your other proposed improvements in this PR. I included them in my new PR and tried to give you credit in the commit message.

PR #1542, opened 2025-07-16
Heap corruption on Windows unit tests

Fi418:
This should solve the Debug build issue mentioned on https://docs.clamav.net/manual/Development/development-builds.html#vcpkg. I only ran into it with address sanitizer enabled, so I'm not sure its the same issue or a separate one. The error referenced is from the debug heap.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/crt-debug-heap-details?view=msvc-170

I noticed and fixed a double free that could also cause it, but didn't run into it on my machine.

### DIFF
--- a/libclamav/readdb.c
+++ b/libclamav/readdb.c
@@ -5655,8 +5655,10 @@ cl_error_t cl_engine_free(struct cl_engine *engine)
     pthread_mutex_unlock(&cli_ref_mutex);
 #endif
 
-    if (engine->stats_data)
+    if (engine->stats_data) {
         free(engine->stats_data);
+        engine->stats_data = NULL;
+    }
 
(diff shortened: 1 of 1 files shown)


val-ms:
Thank you @Sophie0x2E  both of these fixes make sense!  I appreciate the help.

The github actions workflow always fail on windows at this time due to:
1. using an old windows runner which I was trying to fix with this work in progress PR https://github.com/Cisco-Talos/clamav/pull/1524
2. some linking issue with the vcpkg openssl package (no idea how to solve that). 

But in addition to that, I'll also run it through our internal jenkins pipelines and that will test builds on windows x64, x86, and arm64 with some extra feature tests.

PR #1546, opened 2025-07-31
libclamav: Add PDF stats for metadata collection.

jhumlick:
More PDF statistics were requested for feature parity. If metadata collection is enabled, the following additional PDF statistics will be collected:
  - Number of Automatic Actions
  - Number of Streams
  - Number of Objects
  - Number of Object Streams
  - Number of Trailers
  - Number of URIs
  - Number of Xrefs

Additionally, some of the parsing logic was fixed during testing of these features.

CLAM-2820


val-ms:
I was able to reproduce the failed windows test on my local windows machine. It is concerning, since there is no OS-specific logic in the pdf parser that I'm aware of which would cause it not to run that `pdf.stats.nxref++;` line 3913. Perhaps there is an issue with `if (memcmp(q, "%%EOF", 5) == 0)` on Windows?

I'm going to have to diff the debug logs on Windows vs another platform, or else debug-step through with breakpoints to find out what's different.

PR #1550, opened 2025-08-13
Jenkins: Remove GitGuardian stage

val-ms:
We have GitHub Advanced Security (GHAS) enabled.
GitGuardian is no longer used.

CLAM-2836


rsundriyal (approved):
Looks good.

PR #1552, opened 2025-08-13
ZIP: Fix NULL-dereference for OOXML scans

val-ms:
I accidentally introduced a NULL-dereference bug when scanning any OOXML file in https://github.com/Cisco-Talos/clamav/pull/1548

I overlooked the test failure out of haste. 😔

The NULL-dereference happens because the `unzip_search()` feature allowed searching some other file than the one that is currently being scanned, which you would do by setting `ctx` to NULL and setting an `fmap` parameter instead.
In practice, the current layer's `fmap` from the `ctx` was always passed in.

This fix makes it so the `unzip_search()` and related functions only take the `ctx` parameter and do not have and `fmap` or `fsize` field (Note: the `fsize` was never needed, because `fmap->len` take care of that).

CLAM-2837

### DIFF
--- a/libclamav/filetypes.c
+++ b/libclamav/filetypes.c
@@ -278,7 +278,7 @@ const struct ooxml_ftcodes {
         }                                                                       \
     } while (0)
 
-cli_file_t cli_determine_fmap_type(fmap_t *map, const struct cl_engine *engine, cli_file_t basetype)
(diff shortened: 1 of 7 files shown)


TheRaynMan (approved):
Thanks for the unzip.c edits.

jhumlick (approved):

PR #1554, opened 2025-08-18
Tests: fix valgrind complaints regarding open file descriptors

val-ms:


clamdscan/client.c
```diff
@@ -175,7 +175,7 @@ int16_t ping_clamd(const struct optstruct *opts)
     char *errchk                = NULL;
     uint64_t i                  = 0;
     const struct optstruct *opt = NULL;
-    int64_t sockd;
+    int64_t sockd               = 1;
```
jhumlick, line 178: Could this cause an issue with line 266?

jhumlick (approved):

clamdscan/client.c
val-ms: omg was supposed to be -1.

val-ms, line 178:
```suggestion
    int64_t sockd               = -1;
```

jhumlick (dismissed):

PR #1555, opened 2025-08-18
Some fixes for GNU/Hurd

pinotree:
This PR provides some fixes needed to help making clamav build again on GNU/Hurd, like it used to in the past:
- improve the build system for detecting GNU/Hurd, and tweak build flags
- run `lsb_release` as diagnostic
- use POSIX rlimits

There are a couple more fixes needed; this is a good part of the changes needed.


val-ms:
Thanks for your PR. Will review soon. It won't make it into the 1.5.0 release candidate though, which we're preparing to publish tomorrow.

val-ms (approved):
I'm okay with these changes. I can't test them on GNU/Hurd but I did verify they don't break anything in our test pipelines. Thanks for the contribution!

PR #1558, opened 2025-08-22
When loading a CVD, skip .info file DSIG check if used .sign file

val-ms:
If a CVD was verified using a `.sign` file, there's no need to verify the legacy SHA2-256-based RSA DSIG when loading the `.info` file.

In this commit, I've reused the old `CL_DB_ENHANCED` flag to indicate if a CVD was verified using the `.sign` file. This differentiates from using the `CL_DB_SIGNED` flag, which is used for the weaker MD5-based RSA DSIG that would be found within a CVD header.

This gets ClamAV away from requiring the legacy CVD DSIG capability so that in the future we might publish CVD's which *only* have the more modern external `.sign` digital signatures.

CLAM-2851


val-ms:
Augh realized a mistake. Will close and revisit.

PR #1563, opened 2025-08-29
Issue 1506 Solaris requires option -z gnu-version-script-compat

awatkins1966:
The Solaris 11.4 linker supports a subset of GNU ld version scripts,
 but requires a "-z gnu-version-script-compat" to enable it.


awatkins1966:
Hello,

Just added this PR for ClamAV 1.4 branch.

Thanks

val-ms:
Hi @awatkins1966 

We don't merge PR's directly to release branches. If we merge this fix, it goes to main first, then we may backport it when working on the next 1.4 patch version with a PR that targets a patch version development branch, like `dev/1.4.4`

More details on or git workflow, here https://docs.clamav.net/manual/Development/clamav-git-work-flow.html

val-ms:
I don't see any issues with this change in a visual review of the code (haven't tested it yet). We can keep this PR around in draft form until we're ready to work on 1.4.4. 
But please create a pull request from a branch based on `main`.

Issue #1506, opened 2025-05-22
ClamAV 1.4.2 on solaris gives ld: fatal: option --version-script requires option -z gnu-version-script-compat to be specified


awatkins1966:
Describe the bug
----------------

I think you no longer support Solaris, but I still wish to report this.
When you build yoiu get the followinf error:

ld: fatal: option --version-script requires option -z gnu-version-script-compat to be specified
collect2: error: ld returned 1 exit status
gmake[2]: *** [libclamunrar/CMakeFiles/clamunrar.dir/build.make:865: libclamunrar/libclamunrar.so.12.0.3] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:1356: libclamunrar/CMakeFiles/clamunrar.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2

This is a know Solaris feature (https://blogs.oracle.com/solaris/post/regular-expression-and-glob-matching-for-mapfiles), but there is a fix

How to reproduce the problem
----------------------------

./clamconf/clamconf -n
Checking configuration files in /usr/local/clamav/etc

clamd.conf not found

freshclam.conf not found

clamav-milter.conf not found

Software settings
-----------------
Version: 1.4.2
Optional features supported: MEMPOOL AUTOIT_EA06 ICONV

Database information
--------------------
Database directory: /usr/local/clamav/share/clamav
WARNING: freshclam.conf and clamd.conf point to different database directories
print_dbs: Can't open directory /usr/local/clamav/share/clamav

Platform information
--------------------
uname: SunOS
OS: SunOS, ARCH: i386, CPU: i386
Full OS version:                              Oracle Solaris 11.4 X86
zlib version: 1.3.1 (1.3.1), compile flags: a9
platform id: 0x1111d4d408000000000e0200

Build information
-----------------
GNU C: 14.2.0 (14.2.0)
sizeof(void*) = 8
Engine flevel: 212, dconf: 212

Attachments
-----------
 Adding the lines in a few CMakeLists.txt files would solve it:
clamav-1.4.2/libclamunrar_iface/CMakeLists.txt, clamav-1.4.2/libclamav/CMakeLists.txt, clamav-1.4.2/libclamunrar/CMakeLists.txt, clamav-1.4.2/libfreshclam/CMakeLists.txt 
Fir example:

[patchfile.patch](https://github.com/user-attachments/files/20389900/patchfile.patch)

[...]

PR #1569, opened 2025-09-03
Solaris requires -z gnu-version-script-compat to be specified

awatkins1966:
The Solaris 11.4 linker supports a subset of GNU ld version scripts, but requires a "-z gnu-version-script-compat" to enable it.


awatkins1966:
I removed the old PR which was connect to 1.4 release,
Hopefully this is to the main branch and fixes [ClamAV 1.4.2 on solaris gives ld: fatal: option --version-script requires option -z gnu-version-script-compat to be specified #1506](https://github.com/Cisco-Talos/clamav/issues/1506)
Thanks

val-ms (approved):

val-ms:
Thanks @awatkins1966 looks good to me!

PR #1570, opened 2025-09-08
clamd: Do not close FDs in server-th that were opened in clamd.

jhumlick:
ClamD opens at least one socket that is then passed to server-th as newly allocated memory. server-th then appends to this structure with additional FDs as it handles connections. While cleaning up during server shutdown, server-th loops through all FDs and closes them, followed by clamd closing the FDs it opened, which have now been previously closed by server-th. This fix skips closing the FDs in server-th that were opened in clamd.

CLAM-2850


val-ms (approved):
I'm good with this. It makes sense and appears to fix that clamd valgrind issue.

PR #1571, opened 2025-09-15
Reduce unnecessary scanning of embedded file FPs

val-ms:
When embedded file type recognition finds a possible embedded file, it is being scanned as a new embedded file even if it turns out it was a false positive and parsing fails. My solution is to pre-parse the file headers as little possible to determine if it is valid. If possible, also determine the file size based on the headers. That will make it so we don't have to scan additional data when the embedded file is not at the very end.

This commit adds header checks prior to embedded ZIP, ARJ, and CAB scanning. For these types I was also able to use the header checks to determine the object size so as to prevent excessive pattern matching.

TODO: Add the same for RAR, EGG, 7Z, NULSFT, AUTOIT, IShield, and PDF.

This commit also removes duplicate matching for embedded MSEXE. The embedded MSEXE detection and scanning logic was accidentally creating an extra duplicate layer in between scanning and detection because of the logic within the `cli_scanembpe()` function. That function was effectively doing the header check which this commit adds for ZIP, ARJ, and CAB but minus the size check. Note: It is unfortunately not possible to get an accurage size from PE file headers.
The `cli_scanembpe()` function also used to dump to a temp file for no reason since FMAPs were extended to support windows into other FMAPs. So this commit removes the intermediate layer as well as dropping a temp file for each embedded PE file.

Further, this commit adds configuration and DCONF safeguards around all embedded file type scanning.

And this commit moves some of the exclusions for embedded file type recognition relating to ZIP-based formats over to only apply to handling of embedded ZIPs. This is okay now that scan metadata JSON is only recorded for valid ZIPs and will not record many bogus entries for a series of simple PK magic bytes.
The intention is to identify EXEs that may be appended or otherwise somehow hidden within OOXML document in a way that ClamAV doesn't find through ZIP-based extraction.

Finally, this commit adds a set of tests to validate proper extraction of embedded ZIP, ARJ, CAB, and MSEXE files.

CLAM-2862


libclamav/unarj.c
```diff
@@ -56,9 +56,9 @@
 #define CHAR_BIT (8)
 #endif
 #define MAXMATCH 256
-#ifndef FALSE
-#define FALSE (0)
-#define TRUE (1)
+#ifndef false
+#define false (0)
+#define true (1)
```
TheRaynMan, line 61:
```suggestion
```

PR #1584, opened 2025-10-02
Fix potential memory leak

JiangJias:
Add free and replace with textDestroy to avoid potential memory leak.

### DIFF
--- a/libclamav/stats.c
+++ b/libclamav/stats.c
@@ -208,6 +208,8 @@ void clamav_stats_add_sample(const char *virname, const unsigned char *md5, size
                 ;
             p = realloc(sample->virus_name, sizeof(char **) * (i + 1));
             if (!(p)) {
+                for (i = 0; sample->virus_name[i] != NULL; i++)
+                    free(sample->virus_name[i]);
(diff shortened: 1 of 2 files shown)


val-ms:
Thanks for another PR! I'm working on prep to publish ClamAV 1.5.0 early next week and won't be able to merge this until after.

val-ms:
Interestingly, there was overlap with this change and a larger PR i'm working on for the email module: https://github.com/Cisco-Talos/clamav/pull/1720/changes#diff-44edf90419994260f33966e8a9d984e79383352ab996de74121c2f227b85eae5

My PR didn't entirely fix the issue you found in text.c. I'm going to update it to do that rather than take this PR because it will have a merge conflcit otherwise.

The changes to the stats module are likely good other than the issue Copilot just found. But the capability to send those stats was yanked a long time ago (removed in version 0.100.0). So the stats code is dead code. Instead of working on code quality tweaks to dead code, we should just delete the stats code.

Because of this, I'm going to close this PR rather than merge it.

val-ms:
I added a commit to that other PR and gave you credit in the commit message.

PR #1586, opened 2025-10-03
Windows: fix number of arguments in function call

val-ms:
Fix the number of NULL arguments in `scanmem.c` call to `cl_scandesc_ex()`.


csbg-draynor:
Let's add the same comments on the last six arguments to the invocation of cl_scandesc_ex in clamscan/manager.c , then it should be good.

csbg-draynor (approved):
90% comment changes, but worth doing that as we did this.

TheRaynMan (approved):

TheRaynMan (approved):

PR #1607, opened 2025-10-17
libclamav: Add support for Y-type signatures in wdb files.

jhumlick:
There are some cases where it is desired to allow a single domain to have any displayed address and not count that as phishing. An example of this would be the domain for outlook URL checker, or Google safe browsing. If a wdb file contains a Y type entry only the real domain will be matched, not the real and displayed domain.

CLAM-2426

### DIFF
--- a/libclamav/phish_allow_list.c
+++ b/libclamav/phish_allow_list.c
@@ -46,10 +46,6 @@ cl_error_t phish_allow_list_match(const struct cl_engine* engine, char* real_url
 {
     const char* info; /*unused*/
     cli_dbgmsg("Phishing: looking up in allow list: %s:%s; host-only:%d\n", real_url, display_url, hostOnly);
-    // if (hostOnly == 2) {
-    //     // Real URL only lookup for Y-type WDB signatures
-    //     return engine->phish_allow_list_matcher ? regex_list_match(engine->phish_allow_list_matcher, real_url, "", NULL, 1, &info, 2) : 0;
-    // }
     return engine->phish_allow_list_matcher ? regex_list_match(engine->phish_allow_list_matcher, real_url, display_url, NULL, hostOnly, &info, is_allow_list_lookup) : 0;
 }
 
--- a/unit_tests/check_regex.c
+++ b/unit_tests/check_regex.c
@@ -85,7 +85,7 @@ START_TEST(empty)
     ck_assert_msg(!!preg, "malloc");
     rc = cli_regex2suffix(pattern, preg, cb_fail, NULL);
     free(preg);
-    ck_assert_msg(rc == REG_EMPTY, "empty pattern");
+    ck_assert_msg(rc == CL_ERROR, "empty pattern");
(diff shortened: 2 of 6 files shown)


libclamav/phish_allow_list.c
```diff
@@ -42,39 +42,43 @@
...
-cl_error_t allow_list_match(const struct cl_engine* engine, char* real_url, const char* display_url, int hostOnly)
+cl_error_t phish_allow_list_match(const struct cl_engine* engine, char* real_url, const char* display_url, int hostOnly, int is_allow_list_lookup)
 {
     const char* info; /*unused*/
     cli_dbgmsg("Phishing: looking up in allow list: %s:%s; host-only:%d\n", real_url, display_url, hostOnly);
-    return engine->allow_list_matcher ? regex_list_match(engine->allow_list_matcher, real_url, display_url, NULL, hostOnly, &info, 1) : 0;
+    // if (hostOnly == 2) {
+    //     // Real URL only lookup for Y-type WDB signatures
+    //     return engine->phish_allow_list_matcher ? regex_list_match(engine->phish_allow_list_matcher, real_url, "", NULL, 1, &info, 2) : 0;
+    // }
```
csbg-draynor:
```suggestion
```
Obsolete comment

unit_tests/check_regex.c
```diff
@@ -294,7 +294,7 @@ START_TEST(regex_list_match_test)
 
     realurl = cli_safer_strdup(rtest->realurl);
     rc      = regex_list_match(&matcher, realurl, rtest->displayurl, NULL, 1, &info, 1);
-    ck_assert_msg(rc == rtest->result, "regex_list_match");
+    ck_assert_msg(rc == (cl_error_t)rtest->result, "regex_list_match");
```
jhumlick: Change to two different asserts

jhumlick: That requires splitting cli_regex2suffix return code from result

unit_tests/check_regex.c
```diff
@@ -333,33 +333,33 @@ static void psetup_impl(int load2)
...
         signo = 0;
-        rc    = load_regex_matcher(engine, engine->domain_list_matcher, f, &signo, 0, 0, NULL, 1);
+        rc    = load_regex_matcher(engine, engine->phish_protected_domain_matcher, f, &signo, 0, 0, NULL, 1);
         ck_assert_msg(rc == CL_SUCCESS, "load_regex_matcher");
         fclose(f);
 
         ck_assert_msg(signo == 4, "Incorrect number of signatures: %u, expected %u", signo, 4);
     }
-    loaded_2 = load2;
+    loaded_2 = load2; // Were the allow list tests supposed to be wrapped in if (loaded_2) {} ?
```
jhumlick: global static, used elsewhere

unit_tests/check_regex.c
jhumlick: Addressed in CLAM-2925

PR #1612, opened 2025-10-21
Windows: Fix tests when building with examples enabled

val-ms:


jhumlick (changes requested):
Example tests can fail with ENABLE_EXAMPLES=ON and ENABLE_APP=OFF on non-Windows. On the non-Windows path, EX_SCAN_CALLBACKS / EX_CL_CVDUNPACK are set inside if(ENABLE_APP) at [unit_tests/CMakeLists.txt#L266](https://github.com/Cisco-Talos/clamav/blob/f06e72ed028221da2d3b34821f7897afe240332a/unit_tests/CMakeLists.txt#L266), but the examples test is added whenever ENABLE_EXAMPLES is on at [#L416](https://github.com/Cisco-Talos/clamav/blob/f06e72ed028221da2d3b34821f7897afe240332a/unit_tests/CMakeLists.txt#L416). The Python tests then unconditionally do Path(os.getenv(...)) at [ex_scan_callbacks_test.py#L62](https://github.com/Cisco-Talos/clamav/blob/f06e72ed028221da2d3b34821f7897afe240332a/unit_tests/examples/ex_scan_callbacks_test.py#L62) and [ex_cl_cvdunpack_test.py#L29](https://github.com/Cisco-Talos/clamav/blob/f06e72ed028221da2d3b34821f7897afe240332a/unit_tests/examples/ex_cl_cvdunpack_test.py#L29). That env-var assignment should move outside if(ENABLE_APP) for non-Windows too. This same issue is present on current main

PR #1618, opened 2025-10-28
clamd: fix MaxQueue off-by-one in acceptloop

MegaManSec:
Block when recv_fds->nfds >= MaxQueue and only signal when queue < limit, preventing a +1 over-accept.

This bug was found with ZeroPath.

### DIFF
--- a/clamd/server-th.c
+++ b/clamd/server-th.c
@@ -571,7 +571,7 @@ static void *acceptloop_th(void *arg)
             /* don't accept unlimited number of connections, or
              * we'll run out of file descriptors */
             pthread_mutex_lock(recv_fds->buf_mutex);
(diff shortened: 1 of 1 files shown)


val-ms:
Thank you for the PR. 

I reviewed this early this week and must have not submitted my comment. I believe your change is technically correct but does not fix any behavioral issue or technical problem.  As far as I could tell from reading the code, the MaxQueue limit is not used for any allocation and so the only consequence of going higher than the queue limit is more thing in the queue. 

The default queue limit is 200. If my understanding is correct, the existing version actually allows for 201 items in he queue and your fix makes it actually limit to 200. 

Is that right?

MegaManSec:
Yes, it's just a tiny correctional but, nothing serious at all.

MegaManSec:
Can we merge?

PR #1635, opened 2025-11-21
libclamav: Malformed regexes result in misleading memory error messages.

jhumlick:
The regex parsing code converted the return code of anything that did not return CL_SUCCESS to CL_EMEM in several places, resulting in a return code indicative of a malformed database returning a memory error instead. This change passes the real return code up the chain and also prevents trying to realloc 0 bytes when we don't have anything to realloc due to a malformed database.

CLAM-2191

### DIFF
--- a/libclamav/regex_list.c
+++ b/libclamav/regex_list.c
@@ -816,6 +816,10 @@ static cl_error_t add_pattern_suffix(void *cbdata, const char *suffix, size_t su
 
         if (CL_SUCCESS != ret) {
             cli_hashtab_delete(&matcher->suffix_hash, suffix, suffix_len);
+            /* Check if there is anything to shrink back to */
+            if (n == 0) {
+                goto done;
+            }
             /*shrink the size back to what it was.*/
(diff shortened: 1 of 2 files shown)


val-ms (approved):
The code change looks good to me. 
I manually tested with a signature that used to cause the error and I see the output changed from:
```
❯ ~/clams/1.5.1/bin/clamscan -d ./test.wdb blah
LibClamAV Warning: cli_max_realloc(): File or section is too large to scan (0 bytes). For your safety, ClamAV limits how much memory an operation can allocate to 1073741824 bytes
LibClamAV Error: Can't load ./test.wdb: Can't allocate memory
ERROR: Can't allocate memory
```
to 
```
❯ ./install/bin/clamscan -d ./test.wdb blah
LibClamAV Error: Can't load ./test.wdb: Malformed database
ERROR: Malformed database
```
which resolves the biggest concern. If you want to go above and beyond, we could also report the line number in the file which has the invalid signature. But this is much better than before so I'm happy.

We should run it through the jenkins pipelines before merge but it's looking good to me.

val-ms:
Rebased to get the fixes for github windows workflow, and to trigger the test pipelines.

PR #1642, opened 2025-12-05
Use non deprecated OpenSSL accessors for ASN1_STRING values.

bob-beck:
This changes your time conversion routine to use
ASN1_STRING_get0_data() and ASN1_STRING_length() rather than directly using the values from the object.

OpenSSL plans to make ASN1_STRING opaque soon:
https://github.com/openssl/openssl/issues/29117

The accessor in question is in OpenSSL 1.1 as well as BoringSSL and LibreSSL so should be widely available.

As the returned value is const, and you are changing it to pass to strptime, I've made it make a copy to work with.

Since I added another allocation I converted it to single return to ensure everything gets freed appropriately.

### DIFF
--- a/libclamav/crypto.c
+++ b/libclamav/crypto.c
@@ -1760,17 +1760,18 @@ struct tm *cl_ASN1_GetTimeT(ASN1_TIME *timeobj)
 #endif
 
     if (timeobj == NULL || (data = ASN1_STRING_get0_data(timeobj)) == NULL)
-        goto err;
+        goto done;
 
-    if ((str = strndup(data, ASN1_STRING_length(timeobj))) == NULL)
-        goto err;
+    str = CLI_STRNDUP(data, ASN1_STRING_length(timeobj));
+    if (NULL == str)
+        goto done;
 
     if (strlen(str) < 12)
-        goto err;
+        goto done;
 
     t = (struct tm *)calloc(1, sizeof(struct tm));
     if (!(t))
-        goto err;
+        goto done;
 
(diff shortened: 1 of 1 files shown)


libclamav/crypto.c
```diff
@@ -1758,16 +1759,19 @@ struct tm *cl_ASN1_GetTimeT(ASN1_TIME *timeobj)
...
     if (strlen(str) < 12)
-        return NULL;
+        goto err;
 
     t = (struct tm *)calloc(1, sizeof(struct tm));
     if (!(t))
-        return NULL;
+        goto err;
 
     if (timeobj->type == V_ASN1_UTCTIME) {
```
botovq:
```suggestion
    if (ASN1_STRING_type(timeobj) == V_ASN1_UTCTIME) {
```

Similarly in the else if for `V_ASN1_GENERALIZEDTIME` a few lines down, which github doesn't let me add.

val-ms (changes requested):

bob-beck:
> Reading over it, the code looks really good. Thank you for identifying the deprecated API issue and taking initiative to solve it! I also appreciate adding the variable initialization and `goto`-error handling.
> 
> The style is close to what I prefer and you've volunteered your time to help the project, so I'm not going to fuss too much. One small ask - if you could use "done" for the label instead of "err", that would align with the label we use elsewhere and makes it compatible with some convenience macros we define in `others.h`: https://github.com/Cisco-Talos/clamav/blob/main/libclamav/others.h#L1346-L1541
> 
> I'm running it through CI (GitHub Actions then later Jenkins) and will check it later to see how that goes.

Changed the label

PR #1655, opened 2025-12-22
Fixes clamdscan "Unable to determine the filepath given the file descriptor"

### LINKED ISSUE #1653: clamdscan "Unable to determine the filepath given the file descriptor"
The error appears when using the following command on openbsd.

/usr/local/bin/clamdscan -c /etc/clamav/clamd.conf -m --fdpass -i -l /var/log/clamav/scan.log /archive

Previous issues raised the case for other OSs.

finchy:
Fixes #1653

Updated the CheckFDPassing.cmake file to ensure proper support for file descriptor passing on OpenBSD by conditionally defining _XOPEN_SOURCE. The logic now attempts to compile without this definition first, and if that fails, retries with it. Additionally, included fdpassing.h at the beginning of clamdcom.c to ensure correct macro definitions.

### DIFF
--- a/cmake/CheckFDPassing.cmake
+++ b/cmake/CheckFDPassing.cmake
@@ -55,30 +55,63 @@ if(HAVE_CONTROL_IN_MSGHDR)
         set(EXTRA_COMPILE_DEFINITIONS "${EXTRA_COMPILE_DEFINITIONS} -DHAVE_SYS_UIO_H=1")
     endif()
 
-    # Try without _XOPEN_SOURCE first
-    try_run(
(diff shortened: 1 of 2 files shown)


Mohataseem89:
This is a solid portability fix. i especially like that the CMake logic first attempts compilation without `_XOPEN_SOURCE` before redefining it conditionally that avoids unnecessarily altering feature test macros on platforms where it is not required

one question: did you verify whether redefining `_XOPEN_SOURCE` could implicitly affect other headers (e.g., socket or sys/stat behavior) on OpenBSD builds of clamdscan? Sometimes feature-test macros subtly change struct definitions

PR #1659, opened 2026-01-05
Fix unit test issue on newer versions of macOS

val-ms:
Using DYLD_LIBRARY_PATH to modify the dynamic library search path is blocked on some versions of macOS.

CLAM-2878


val-ms:
I no longer believe this is the correct solution. I will have to play with it more when I have time.

PR #1678, opened 2026-02-24
clamav: Add support for PDFium to clamav cmake

jhumlick:
In order to support utilizing PDFium to render pdf files to image and then hash the image, we need to be able to link against PDFium. This change adds support to link at build time.

CLAM-2817


jhumlick:
Mistakenly created this PR with the wrong branch/name

PR #1679, opened 2026-02-24
clamav: Add support for PDFium to clamav cmake

jhumlick:
In order to support utilizing PDFium to render pdf files to image and then hash the image, we need to be able to link against PDFium. This change adds support to link at build time.

CLAM-2817

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,7 +349,7 @@ if (WIN32)
     set(CMAKE_INSTALL_SYSTEM_RUNTIME_DESTINATION .)
     if(ENABLE_WINDOWS_INSTALL_THIRDPARTY_DEPENDENCIES)
         include(InstallRequiredSystemLibraries)
-    endif()	
+    endif()
 
     set(CPACK_GENERATOR                  "ZIP;WIX")
     set(CPACK_PACKAGE_VENDOR             "Cisco Systems, Inc.")
@@ -535,13 +535,7 @@ if(NOT WIN32)
 endif()
 
 if(ENABLE_PDFIUM)
-    find_package(PDFIUM QUIET)
-    if(PDFIUM_FOUND)
-        set(HAVE_PDFIUM 1)
-    else()
-        set(ENABLE_PDFIUM OFF)
-        message(STATUS "PDFium not found; PDFium support disabled")
-    endif()
+    find_package(PDFIUM REQUIRED MODULE)
 endif()
 
 if(ENABLE_JSON_SHARED)
@@ -1270,6 +1264,14 @@ ${b}        JSON support:       ${e}
 ${_}            json-c          ${e}${JSONC_INCLUDE_DIRS}
 ${_}                            ${e}${JSONC_LIBRARIES}
 ${b}        Threading support:  ${e}")
+if(ENABLE_PDFIUM)
+message("\
+${_}            pdfium          ${e}${PDFIUM_INCLUDE_DIRS}
+${_}                            ${e}${PDFIUM_LIBRARIES}")
+else()
+message("\
+${o}            no              ${e}")
+endif()
 if(WIN32)
 message("\
 ${_}            pthread-win32   ${e}${PThreadW32_INCLUDE_DIRS}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ endif()
 set(ENABLE_DOXYGEN_DEFAULT          OFF)
 set(ENABLE_UNRAR_DEFAULT            ON)
 set(ENABLE_SYSTEMD_DEFAULT          ON)
+set(ENABLE_PDFIUM_DEFAULT           ON)
 
 # See CMakeOptions.cmake for additional options.
 include(CMakeOptions.cmake)
@@ -531,6 +532,16 @@ if(NOT WIN32)
     find_package(Iconv REQUIRED)
     # Set variable required by libclamav to use iconv
     set(HAVE_ICONV 1)
+endif()
+
+if(ENABLE_PDFIUM)
+    find_package(PDFIUM QUIET)
+    if(PDFIUM_FOUND)
+        set(HAVE_PDFIUM 1)
+    else()
+        set(ENABLE_PDFIUM OFF)
+        message(STATUS "PDFium not found; PDFium support disabled")
+    endif()
 endif()
 
 if(ENABLE_JSON_SHARED)
--- a/libclamav/CMakeLists.txt
+++ b/libclamav/CMakeLists.txt
@@ -450,6 +450,11 @@ if(ENABLE_SHARED_LIB)
             LibXml2::LibXml2
             JSONC::jsonc )
 
+    if(ENABLE_PDFIUM AND PDFIUM_FOUND)
+        target_link_libraries( clamav PUBLIC ${PDFIUM_LIBRARY} )
+        target_compile_definitions( clamav PRIVATE HAVE_PDFIUM=1 )
+    endif()
+
     if(WIN32)
         target_link_libraries( clamav
             PUBLIC
@@ -573,6 +578,10 @@ if(ENABLE_STATIC_LIB)
             PCRE2::pcre2
             LibXml2::LibXml2
             JSONC::jsonc )
+        if(ENABLE_PDFIUM AND PDFIUM_FOUND)
+            target_link_libraries( clamav_static PUBLIC ${PDFIUM_LIBRARY} )
+            target_compile_definitions( clamav_static PRIVATE HAVE_PDFIUM=1 )
+        endif()
         if (ENABLE_UNRAR)
             target_link_libraries( clamav_static PUBLIC ClamAV::libunrar_iface_static ClamAV::libunrar_iface_iface)
         endif()
--- a/libclamav/CMakeLists.txt
+++ b/libclamav/CMakeLists.txt
@@ -450,9 +450,8 @@ if(ENABLE_SHARED_LIB)
             LibXml2::LibXml2
             JSONC::jsonc )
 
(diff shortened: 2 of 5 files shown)


libclamav/CMakeLists.txt
```diff
@@ -450,6 +450,11 @@ if(ENABLE_SHARED_LIB)
             LibXml2::LibXml2
             JSONC::jsonc )
 
+    if(ENABLE_PDFIUM AND PDFIUM_FOUND)
+        target_link_libraries( clamav PUBLIC ${PDFIUM_LIBRARY} )
+        target_compile_definitions( clamav PRIVATE HAVE_PDFIUM=1 )
```
val-ms:
```suggestion
```

Instead of this, I would like to add the HAVE_PDFIUM definition using a `#cmakedefine` macro in `clamav-config.h.cmake.in`. E.g.: https://github.com/Cisco-Talos/clamav/blob/CLAM-2502-pdf-first-page-rendering/clamav-config.h.cmake.in#L304-L305

You would just add this, and it will automatically add `#define HAVE_PDFIUM 1` to the generated `clamav-config.h` file if `HAVE_PDFIUM` is set within CMake:
```h
/* pdfium is supported */
#cmakedefine HAVE_PDFIUM 1
```

libclamav/CMakeLists.txt
```diff
@@ -573,6 +578,10 @@ if(ENABLE_STATIC_LIB)
             PCRE2::pcre2
             LibXml2::LibXml2
             JSONC::jsonc )
+        if(ENABLE_PDFIUM AND PDFIUM_FOUND)
+            target_link_libraries( clamav_static PUBLIC ${PDFIUM_LIBRARY} )
+            target_compile_definitions( clamav_static PRIVATE HAVE_PDFIUM=1 )
```
val-ms:
```suggestion
```

Same comment as above.

CMakeLists.txt
```diff
@@ -533,6 +534,16 @@ if(NOT WIN32)
...
 endif()
 
+if(ENABLE_PDFIUM)
+    find_package(PDFIUM QUIET)
+    if(PDFIUM_FOUND)
+        set(HAVE_PDFIUM 1)
+    else()
+        set(ENABLE_PDFIUM OFF)
+        message(STATUS "PDFium not found; PDFium support disabled")
+    endif()
```
val-ms:
```suggestion
    find_package(PDFIUM REQUIRED)
    set(HAVE_PDFIUM 1)
```

If `ENABLE_PDFIUM` is set, then I think it should fail if it can't find PDFIUM.  I think that means remove "QUIET"  (?) and add "REQUIRED". 

```suggestion
    find_package(PDFIUM REQUIRED)
```

That makes it simpler logic and less likely to have unexpected behavior if someone asks for PDFIUM but then doesn't get it. 

This will also make it so you only have to check if `ENABLE_PDFIUM` *or* `HAVE_PDFIUM` later to link it and don't need to check both.

libclamav/CMakeLists.txt
```diff
@@ -573,6 +578,10 @@ if(ENABLE_STATIC_LIB)
             PCRE2::pcre2
             LibXml2::LibXml2
             JSONC::jsonc )
+        if(ENABLE_PDFIUM AND PDFIUM_FOUND)
```
val-ms:
```suggestion
        if(ENABLE_PDFIUM)
```

libclamav/CMakeLists.txt
```diff
@@ -450,6 +450,11 @@ if(ENABLE_SHARED_LIB)
             LibXml2::LibXml2
             JSONC::jsonc )
 
+    if(ENABLE_PDFIUM AND PDFIUM_FOUND)
```
val-ms:
```suggestion
    if(ENABLE_PDFIUM)
```

CMakeLists.txt
```diff
@@ -110,6 +110,7 @@ endif()
 set(ENABLE_DOXYGEN_DEFAULT          OFF)
 set(ENABLE_UNRAR_DEFAULT            ON)
 set(ENABLE_SYSTEMD_DEFAULT          ON)
+set(ENABLE_PDFIUM_DEFAULT           ON)
```
val-ms, line 113: I am unsure if this should be on or off by default. We will be shipping it on, and we'd like others to use it... so... ON is probably the right call.

libclamav/CMakeLists.txt
```diff
@@ -573,6 +578,10 @@ if(ENABLE_STATIC_LIB)
             PCRE2::pcre2
             LibXml2::LibXml2
             JSONC::jsonc )
+        if(ENABLE_PDFIUM AND PDFIUM_FOUND)
+            target_link_libraries( clamav_static PUBLIC ${PDFIUM_LIBRARY} )
```
val-ms:
```suggestion
            target_link_libraries( clamav_static PUBLIC PDFIUM::PDFIUM )
```

Ideally this would be a native CMake "target" and not a variable. CMake that will make it so CMake provides not only the linking but also adds the headers to the include path when compiling `clamav_static`.

libclamav/CMakeLists.txt
```diff
@@ -450,6 +450,11 @@ if(ENABLE_SHARED_LIB)
             LibXml2::LibXml2
             JSONC::jsonc )
 
+    if(ENABLE_PDFIUM AND PDFIUM_FOUND)
+        target_link_libraries( clamav PUBLIC ${PDFIUM_LIBRARY} )
```
val-ms:
```suggestion
        target_link_libraries( clamav PUBLIC PDFIUM::PDFIUM )
```

Ideally this would be a native CMake "target" and not a variable. CMake that will make it so CMake provides not only the linking but also adds the headers to the include path when compiling `clamav`.

val-ms (changes requested):
Looks like this is missing a `FindPDFIUM.cmake` file. Did you forget to `git add` it, or is it finding PDFIUM some other way?

CMakeLists.txt
Sanesecurity: Adding pdfium is a nice move.

random early thoughts...

a) is there a big hit on scan speed for large pdfs?
b) will it support passworded pdfs or skip them?
c) what sig type to match extracted image hdb, ldb etc.

val-ms: The pdfium integration is intended to improve detection by enabling us to create image fuzzy hash signatures for PDF first page renders. It isn't expected to replace ClamAV's existing PDF parsing capability at this time.

> a) is there a big hit on scan speed for large pdfs?

Because it is only rendering the first page, I do not believe there will be a significant scan speed hit. @jhumlick did a thorough analysis comparing pdfium with another library for document rendering, ImageMagic, using a large dataset of assorted PDFs. The results of that analysis and concerns about the open source license for GhostScript, an ImageMagic dependency, lead us to move forward with pdfium instead.

I once we have pdfium integrated into ClamAV, I will also be interested to see a scan performance comparison of the ClamAV main branch vs the pdfium PR using those same files. We also have concerns about performance, and the availability of the pdfium library. So the pdfium dependency will be optional. Even when provided, the pdf rendering capability will also be optional. 

> b) will it support passworded pdfs or skip them?

I'm not sure what will happen if we try to render passworded pdf's. I imagine it will fail and have to skip them.

> c) what sig type to match extracted image hdb, ldb etc.

I believe we'll be treating the rendered image as a contained layer that is a normalized version of the PDF. The expectation is to match using ldb signatures with the image fuzzy hash subsignature feature, though I suppose rendered PDFs would be also be matched against all signatures that can work with JPG images (i.e. target type 0 and target type 5).

CMakeLists.txt
```diff
@@ -533,6 +534,10 @@ if(NOT WIN32)
     set(HAVE_ICONV 1)
 endif()
 
+if(ENABLE_PDFIUM)
+    find_package(PDFIUM REQUIRED MODULE)
```
val-ms, line 538: The "MODULE" keyword is new to me. I like it. I think it may be useful for some of the other ones where I've seen them ship broken CMake Config files. Though we rely heavily on libcurl's CMake config file, which in 8.18.0 IS broken. Fixing it is my current time sink.

CMakeLists.txt
```diff
@@ -1252,6 +1257,9 @@ ${_}                            ${e}${LIBXML2_LIBRARIES}
 ${b}        RegEx support:      ${e}
 ${_}            libpcre2        ${e}${PCRE2_INCLUDE_DIRS}
 ${_}                            ${e}${PCRE2_LIBRARIES}
+${b}        PDF Rendering:      ${e}
+${_}            pdfium          ${e}${PDFIUM_INCLUDE_DIRS}
+${_}                            ${e}${PDFIUM_LIBRARIES}
```
val-ms: I had been thinking we could do an if/else around this like we do with Thread support, to say "no" if PDF Rendering" is not enabled. Like:
```cmake
if(ENABLE_PDFIUM)
message("\
${_}            pdfium          ${e}${PDFIUM_INCLUDE_DIRS}
${_}                            ${e}${PDFIUM_LIBRARIES}
else()
message("\
${o}            no              ${e}")
endif()
```

it is a little more complicated because you have to end the message function and restarted it. But would look prettier than heaving empty values for pdfium when not enabled.

val-ms (changes requested):

val-ms (approved):

Sanesecurity:
Just in case this is useful for testing... A pdf ocr code phishing attempt I received.

Assuming the image extracts ok from the pdf, the image will contain a qrcode... a qrcode matcher or hasher would be interesting.

If the image does extract, it won't be a good hash for detection as it has the non-unique sanesecurity in the text of the image.

[sanesecurity_Year-End_Bonus_Form.pdf](https://github.com/user-attachments/files/25616555/sanesecurity_Year-End_Bonus_Form.pdf)

PR #1694, opened 2026-03-24
Clam 2502 pdf first page rendering

jhumlick:
Add runtime PDF image fuzzy hash rendering feature using PDFium

Add clamscan/clamd options to enable or disable PDF render fuzzy hashing,
support DPI or WIDTHxHEIGHT canvas render settings, preserve rendered output
with leave-temps, and add a clamscan unit test for canvas parsing.
    
CLAM-2817


clamd/server-th.c
```diff
@@ -58,6 +59,51 @@
...
+    }
+
+    if ((0 == strcmp(value, "jpeg")) || (0 == strcmp(value, "jpg"))) {
+        *format = 2;
+        return 1;
+    }
+
+    return 0;
+}
+
```
val-ms: I'm not sure if it's a style thing or of there's a functional reason. I'd strongly prefer to move static function definitions below the `#include` section, and even below the global variable declarations. Somewhere around like 147. 

That said, I noticed it's also duplicated in clamscan.c. I'd prefer not to have duplicate code, though this is pretty tiny and maybe outside of `common/misc.c` I don't see a good place to put it.

libclamav/others.h
val-ms, line 1: There are convenience macros for referencing the config options here https://github.com/Cisco-Talos/clamav/blob/82dbc5a28c759eaff2303200133377b9e4c511e1/libclamav/others.h#L565

I recommend adding `SCAN_PARSE_PDF_IMAGE_FUZZY_HASH` for the `CL_SCAN_PARSE_PDF_IMAGE_FUZZY_HASH`

libclamav/scanners.c
```diff
@@ -4349,13 +4349,28 @@ static cl_error_t dispatch_prescan_callback(clcb_pre_scan cb, cli_ctx *ctx, cons
...
+
+    if ((ctx->options->parse & CL_SCAN_PARSE_PDF_IMAGE_FUZZY_HASH) &&
+        (ctx->recursion_level > 0) &&
+        (ctx->recursion_stack[ctx->recursion_level].attributes & LAYER_ATTRIBUTES_NORMALIZED) &&
+        (ctx->recursion_stack[ctx->recursion_level - 1].type == CL_TYPE_PDF)) {
+        return true;
+    }
+
+    return false;
+}
```
val-ms, line 4378: I see one issue here. I think `OTHER_CONF_IMAGE_FUZZY_HASH` should gate all image fuzzy hashing, and not just non-PDF image fuzzy hashing.  

I don't remember if we've talked about it before, [DCONF is a signature-based config system whereby we can enable or disable engine features via signature update](https://docs.clamav.net/manual/Signatures/DynamicConfig.html).

Another option is to introduce an extra DCONF setting which disables PDF rendering. That would be more precise, in caase there is a serious issue with pdf rendering and we want to globally disable it. 

Also, Codex suggested making a separate function to make it easier to unerstand the complex recursion stack attribute/file check. I agree it does help with readability.

So either this:
```c
static bool is_pdf_rendered_image_layer(cli_ctx *ctx)
{
    return (ctx->recursion_level > 0) &&
           (ctx->recursion_stack[ctx->recursion_level].attributes & LAYER_ATTRIBUTES_NORMALIZED) &&
           (ctx->recursion_stack[ctx->recursion_level - 1].type == CL_TYPE_PDF);
}

static bool should_calculate_image_fuzzy_hash(cli_ctx *ctx)
{
    if (!(DCONF_OTHER & OTHER_CONF_IMAGE_FUZZY_HASH)) {
        return false;
    }

    if (SCAN_PARSE_IMAGE_FUZZY_HASH) {
        return true;
    }

    if (SCAN_PARSE_IMAGE_FUZZY_HASH && is_pdf_rendered_image_layer(ctx)) {
        return true;
    }

    return false;
}
```

Or this:
```c
static bool is_pdf_rendered_image_layer(cli_ctx *ctx)
{
    return (ctx->recursion_level > 0) &&
           (ctx->recursion_stack[ctx->recursion_level].attributes & LAYER_ATTRIBUTES_NORMALIZED) &&
           (ctx->recursion_stack[ctx->recursion_level - 1].type == CL_TYPE_PDF);
}

static bool should_calculate_image_fuzzy_hash(cli_ctx *ctx)
{
    if ((DCONF_OTHER & OTHER_CONF_IMAGE_FUZZY_HASH) &&
        (SCAN_PARSE_IMAGE_FUZZY_HASH)) {
        return true;
    }

    if ((DCONF_OTHER & OTHER_CONF_PDF_IMAGE_FUZZY_HASH) &&
        (SCAN_PARSE_IMAGE_FUZZY_HASH && is_pdf_rendered_image_layer(ctx)) {
        return true;
    }

    return false;
}
```

I kind of lean towards the second option.

libclamav/scanners.c
```diff
@@ -4400,6 +4415,104 @@ static cl_error_t calculate_fuzzy_image_hash(cli_ctx *ctx, cli_file_t type)
...
+        }
+
+        cli_dbgmsg("Rendered PDF image written to %s\n", rendered_image_path);
+
+        status = cli_magic_scan_desc_type(rendered_image_fd, rendered_image_path, ctx, rendered_image.image_type,
+                                          rendered_image_name, LAYER_ATTRIBUTES_NORMALIZED);
+    } else {
+        status = cli_magic_scan_buff(rendered_image.image_data, rendered_image.image_len, ctx,
+                                     rendered_image.image_type, rendered_image_name, LAYER_ATTRIBUTES_NORMALIZED);
+    }
```
val-ms:
```suggestion
    status = cli_magic_scan_buff(rendered_image.image_data, rendered_image.image_len, ctx,
                                     rendered_image.image_type, rendered_image_name, LAYER_ATTRIBUTES_NORMALIZED);

```

Consider simplifying this to always use cli_magic_scan_buff(). If users need the rendered image left on disk, they can combine `--leave-temps` with `--force-to-disk`, which already causes nested memory scans to be materialized and retained. That would remove this custom temp-file write/scan path and rely on the existing nested-buffer scan behavior.

This would change leave-temps-only behavior: rendered PDF images would no longer be preserved unless force-to-disk is also enabled.

What do you think? The upside I think is it's simpler code. There is another downside in that you don't get to choose the temp filename. The metadata JSON will indicate it's a normalized layer and an image, but that will be your only hint.

libclamav_rust/src/fuzzy_hash.rs
```diff
@@ -422,6 +422,10 @@ pub fn fuzzy_hash_calculate_image(buffer: &[u8]) -> Result<Vec<u8>, Error> {
         Err(_) => return Err(Error::ImageLoadPanic()),
     };
 
+    fuzzy_hash_calculate_dynamic_image(og_image)
+}
+
+pub fn fuzzy_hash_calculate_dynamic_image(og_image: DynamicImage) -> Result<Vec<u8>, Error> {
```
val-ms, line 428: No objection to this. I don't see the `fuzzy_hash_calculate_dynamic_image` being used anywhere else, but I can imagine it might be nice to have in the future.

libclamav_rust/src/pdf_render.rs
```diff
@@ -0,0 +1,189 @@
+/*
+ *  Render the first page of a PDF document as jpeg in a [u8].
+ *
+ *  Copyright (C) 2023 Cisco Systems, Inc. and/or its affiliates. All rights reserved.
+ *
+ *  Authors: Micah Snyder
```
val-ms: Feel free to change this to your name and Solal Jacob (if any of the code from the og prototype is in here ( https://github.com/Cisco-Talos/clamav/compare/main...vertrex:clamav:pdf-rendering )

val-ms (changes requested):
Overall this looks like very good work.

Codex found a couple things, and I found one or two less useful more fussy things.

- P1: `ENABLE_PDFIUM=OFF` builds are broken. [Cargo.toml](/Users/val/dev/clamav/clamav-1/libclamav_rust/Cargo.toml:8) unconditionally pulls in `pdfium-render` with static bindings, and [lib.rs](/Users/val/dev/clamav/clamav-1/libclamav_rust/src/lib.rs:38) unconditionally compiles `pdf_render`, which calls static PDFium bindings at [pdf_render.rs](/Users/val/dev/clamav/clamav-1/libclamav_rust/src/pdf_render.rs:80). But [libclamav/CMakeLists.txt](/Users/val/dev/clamav/clamav-1/libclamav/CMakeLists.txt:453) only links `PDFIUM::pdfium` when `ENABLE_PDFIUM` is ON. Repro: `cmake -S . -B build -D ENABLE_PDFIUM=OFF` then `make -j12` fails linking `libclamav.dylib` with many undefined `_FPDF_*` symbols. The Rust dependency/module needs to be Cargo-feature gated from CMake, or PDFium becomes a required link dependency even when disabled.

- P3: The new clamd config options are missing from the installed man page. [common/optparser.c](/Users/val/dev/clamav/clamav-1/common/optparser.c:485) adds `ScanPDFImageFuzzyHash`, `PDFRenderDPI`, `PDFRenderCanvas`, and `PDFRenderFormat`, and the sample configs were updated, but [docs/man/clamd.conf.5.in](/Users/val/dev/clamav/clamav-1/docs/man/clamd.conf.5.in:574) still jumps from `ScanImageFuzzyHash` to `AlertBrokenExecutables`. Packaged users relying on `man clamd.conf` will not see the new settings.

I haven't tried a PDFium-enabled build yet.

PR #1697, opened 2026-03-30
Fix short writes in scanner temp output paths

val-ms:
Fix an EGG comment temp-file bug where the write length used the number of comments instead of the length of the current comment.

Also switch nearby raw write paths in scanners.c to cli_writen() so short writes are handled correctly for RAR comments, extracted EGG temp files, normalized script output, and UTF-16 HTML temp output.

CLAM-2946

### DIFF
--- a/libclamav/scanners.c
+++ b/libclamav/scanners.c
@@ -281,7 +281,7 @@ static cl_error_t cli_scanrar_file(const char *filepath, int desc, cli_ctx *ctx)
                 cli_dbgmsg("RAR: ERROR: Failed to open output file\n");
             } else {
                 cli_dbgmsg("RAR: Writing the archive comment to temp file: %s\n", comment_fullpath);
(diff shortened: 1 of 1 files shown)


jhumlick:
`libclamav/scanners.c`, hunk `@@ -690,8 +690,10 @@ static cl_error_t cli_scanegg(...)`: this fixes the EGG comment length bug, but it still leaves the same short-write class in place for that temp-file path because it continues to call raw `write()` instead of `cli_writen()`. A positive short write will still truncate the kept temp file silently after only logging an error. Since the PR’s goal is to harden temp output writes, I’d recommend changing this line to `cli_writen(comment_fd, comments[i], comment_len) != comment_len` as well.

val-ms:
I think this is fixed and ready for re-review

jhumlick (approved):

PR #1702, opened 2026-04-03
Fix incorrect file name reference in error message within sigtool

ipnerds:
The error message printed when sigtool --testsigs cannot open the file provides the incorrect file reference creating a confusing situation for the end user.  This PR fixes that reference to match the file that was attempted.


val-ms (approved):
Thanks for the fix!

PR #1703, opened 2026-04-03
Clamonacc: Fix multithreaded scanning issue; Remove the scan mutex safely

### LINKED ISSUE #1698: Clamonacc Feature Request: Parallelize clamd->clamonacc scan results
Currently, even though clamonacc uses a thread pool, those aforementioned threads are primarily bottlenecked on these lines of code:
https://github.com/Cisco-Talos/clamav/blob/dad11f9b977faad11dff9ab6a26207294a1db5fb/clamonacc/scan/thread.c#L128-L134

Wherein each thread in the thread pool compete for the same `onas_scan_lock` mutex.
And once one thread acquires the `onas_scan_lock` mutex, 
all other threads have to wait for the running thread to *both* send the scan request *to* clamd, 
*and then* receive the scan result *from* clamd.

onas_client_scan snippet:
https://github.com/Cisco-Talos/clamav/blob/dad11f9b977faad11dff9ab6a26207294a1db5fb/clamonacc/client/client.c#L553-L564
https://github.com/Cisco-Talos/clamav/blob/dad11f9b977faad11dff9ab6a26207294a1db5fb/clamonacc/client/client.c#L564-L567

Here's a screenshot of what this looks like from an intel VTune analysis (`/opt/intel/oneapi/vtune/2025.0/bin64/vtune -collect threading -knob stack-size=0 -data-limit=0 -finalization-mode=full --target-process clamonacc`):

Image

Image

Image

Image

---

#### Wherein the code is refactored so that the `onas_scan_lock` mutex is only held while the scan *request* is *sent from* clamonacc to clamd (clamonacc->clamd), but **not** held while a clamonacc thread is waiting to *receive* the scan *result* (clamd->clamonacc)

Here's my thoughts on how this can be implemented (note: code snippets below lack error handling): 
- Create a new scantype: `ASYNC_FILDES`: basically a version of `FILDES` for clamonacc
- Have each clamonacc thread in the threadpool create & initialize a dedicated RX unix socket before the while loop in [`void* thread_do(struct thread* thread_p)`](https://github.com/Cisco-Talos/clamav/blob/dad11f9b977faad11dff9ab6a26207294a1db5fb/clamonacc/c-thread-pool/thpool.c#L319)
The dedicated RX unix sockets' pathname should be a combination of a configurable prefix (`RxSocketPrefix`) defaulting to `/var/run/clamav/clamonaccRX` + the thread number (so for 5 clamonacc threads, /var/run/clamav/clamonaccRX0-/var/run/clamav/clamonaccRX4)
```c
int rxsockfd = socket(AF_UNIX, SOCK_STREAM, 0);
struct sockaddr_un rxserver;
memset((char *)&rxserver, 0, sizeof(rxserver));
rxserver.sun_family = AF_UNIX;

//thread_p->id:int --> INT_MIN=-2147483648->11 chars, ".sock"->5 chars, '\0'->1 char
//max suffix size: 17 chars
char suffix[17];
int suffix_len = sprintf(suffix, "%i.sock", thread_p->id);
strncpy(rxserver.sun_path, optget(opts, "RxSocketPrefix")->strarg, sizeof(rxserver.sun_path) - suffix_len);
rxserver.sun_path[sizeof(rxserver.sun_path) - suffix_len - 1] = '\0';
strcat(rxserver.sun_path, suffix);
bind(rxsockfd, (struct sockaddr *)&rxserver, sizeof(struct sockaddr_un));

//change socket ownership to clamav
//(note: add include <pwd.h>)
struct passwd* pwd = getpwnam(optget(opts, "User")->strarg);
fchown(rxsockfd, pwd->pw_uid, pwd->pw_gid);
fchmod(rxsockfd, 600); //don't let non-clamav users hijack rxsockfd

backlog = optget(opts, "MaxConnectionQueueLength")->numarg;
logg(LOGG_INFO_NF, "thread_do(): Setting Rx queue length to %d\n", backlog);

listen(rxsockfd, backlog);
```

-  modify [`int onas_send_fdpass(int sockd, int fd)`](https://github.com/Cisco-Talos/clamav/blob/dad11f9b977faad11dff9ab6a26207294a1db5fb/clamonacc/client/protocol.c#L155) so that it passes both the fd of the file to be scanned, and the fd of the clamonacc rx socket:
```c
static int onas_send_fdpass(int sockd, int rxsockd, int fd)
{
...
 int passfds[] = {fd, rxsockfd};
 unsigned char fdbuf[CMSG_SPACE(sizeof(passfds))];
 const char zFILDES[] = "zASYNC_FILDES";
...
 msg.msg_controllen      = CMSG_LEN(sizeof(passfds));
 cmsg                    = CMSG_FIRSTHDR(&msg);
 cmsg->cmsg_len          = CMSG_LEN(sizeof(passfds));
 cmsg->cmsg_level        = SOL_SOCKET;
 cmsg->cmsg_type         = SCM_RIGHTS;
 memcpy(CMSG_DATA(cmsg), passfds, sizeof(passfds));
...
```

After `onas_send_fdpass` is run, the thread should release the `onas_scan_lock` mutex
Then do `accept()` on `rxsockfd`, and select & recv on the `rxsockfd` for the scan results

- edit clamd code so that for `ASYNC_FILDES`, it takes the second passed fd -- `rxsockfd` -- and instead of sending the scan results via the main socket, it does `connect()` + `send()` via `rxsockfd`. Ensure the scan result is sent directly by the clamd worker thread that handled the scan, so that the main clamd thread can simultaneously process new requests from the main socket.
After `send()` returns, it should then do `close()` on sockfd, so that future `connect()`s on sockfd don't fail 
also in clamd, optionally, as a safety measure, also reject the scan request if the main socket connection isn't a local unix socket or if `fstat()` on `rxsockfd` indicates the passed `rxsockfd` isn't owned by (username) clamav

EDIT: now that I think about it, the clamonacc threads would need to create rxsockfd as non-blocking, and then before each call that changes from blocking io->non-blocking io, there would need to be a ‘poll()’ call watching both rxsockfd and sockfd, but w/ all events masked out for sockfd (which makes it just watch for `POLLHUP’ & other exception-events). Otherwise I think there’d be a slight chance of clamonacc hanging after clamd closes.

val-ms:
A scan mutex is preventing concurrent scans from clamonacc's thread pool.

Removing the on-access scan mutex exposed thread-unsafe state in clamonacc and common helper code, and it left us without test coverage for concurrent client traffic. Under load this could miscount ALLMATCH detections, race connection-status logging, and hide regressions in fd-passing behavior.

Fix the shared state so concurrent scans can proceed safely by making the disconnect log state mutex-protected, tracking ALLMATCH filenames per scan request, and updating the action failure counters atomically.

Add a Linux-only clamonacc client unit test that exercises concurrent STREAM, ALLMATCH, and FDPASS scans against fake daemons.

Resolves: https://github.com/Cisco-Talos/clamav/issues/1698

CLAM-2251

### DIFF
--- a/clamonacc/client/client.c
+++ b/clamonacc/client/client.c
@@ -47,6 +47,7 @@
 #include <errno.h>
 #include <dirent.h>
 #include <fcntl.h>
+#include <pthread.h>
 
 #ifdef HAVE_SYS_UIO_H
 #include <sys/uio.h>
@@ -70,6 +71,33 @@
 #include "socket.h"
 
 #include "../clamonacc.h"
+
+static pthread_mutex_t onas_disconnect_lock = PTHREAD_MUTEX_INITIALIZER;
+static bool onas_client_disconnected        = false;
+
+static void onas_client_note_connection_failure(CURLcode curlcode)
+{
+    pthread_mutex_lock(&onas_disconnect_lock);
+
+    if (!onas_client_disconnected) {
+        logg(LOGG_ERROR, "ClamClient: Connection to clamd failed, %s.\n", curl_easy_strerror(curlcode));
+        onas_client_disconnected = true;
+    }
+
+    pthread_mutex_unlock(&onas_disconnect_lock);
+}
+
+static void onas_client_note_connection_success(void)
+{
+    pthread_mutex_lock(&onas_disconnect_lock);
+
+    if (onas_client_disconnected) {
+        logg(LOGG_INFO, "ClamClient: Connection to clamd re-established.\n");
+        onas_client_disconnected = false;
+    }
+
+    pthread_mutex_unlock(&onas_disconnect_lock);
+}
 
 void onas_print_server_version(struct onas_context **ctx)
 {
@@ -535,7 +563,6 @@ int onas_client_scan(const char *tcpaddr, int64_t portnum, int32_t scantype, uin
     CURLcode curlcode = CURLE_OK;
     int errors        = 0;
     int ret;
-    static bool disconnected = false;
 
     *infected = 0;
 
@@ -552,17 +579,11 @@ int onas_client_scan(const char *tcpaddr, int64_t portnum, int32_t scantype, uin
 
     curlcode = curl_easy_perform(curl);
     if (CURLE_OK != curlcode) {
-        if (!disconnected) {
-            logg(LOGG_ERROR, "ClamClient: Connection to clamd failed, %s.\n", curl_easy_strerror(curlcode));
-            disconnected = true;
-        }
+        onas_client_note_connection_failure(curlcode);
         curl_easy_cleanup(curl);
         return CL_ECREAT;
     }
-    if (disconnected) {
-        logg(LOGG_INFO, "ClamClient: Connection to clamd re-established.\n");
-        disconnected = false;
-    }
+    onas_client_note_connection_success();
 
     if ((ret = onas_dsresult(curl, scantype, maxstream, fname, fd, timeout, &ret, err, ret_code)) >= 0) {
         *infected = ret;
--- a/clamonacc/client/protocol.c
+++ b/clamonacc/client/protocol.c
@@ -70,6 +70,27 @@
 
 static const char *scancmd[] = {"CONTSCAN", "MULTISCAN", "INSTREAM", "FILDES", "ALLMATCHSCAN"};
 
+struct onas_reply_state {
+    char last_filename[PATH_MAX + 1];
+};
+
+static int onas_count_virus_hit(struct onas_reply_state *reply_state, int scantype, const char *filename)
+{
+    if (scantype != ALLMATCH) {
+        return 1;
+    }
+
+    if ((NULL == filename) || (0 != strcmp(filename, reply_state->last_filename))) {
+        if (NULL != filename) {
+            strncpy(reply_state->last_filename, filename, PATH_MAX);
+            reply_state->last_filename[PATH_MAX] = '\0';
+        }
+        return 1;
+    }
+
+    return 0;
+}
+
 /* Issues an INSTREAM command to clamd and streams the given file
  * Returns >0 on success, 0 soft fail, -1 hard fail */
 static int onas_send_stream(CURL *curl, const char *filename, int fd, int64_t timeout, uint64_t maxstream)
@@ -237,9 +258,12 @@ int onas_dsresult(CURL *curl, int scantype, uint64_t maxstream, const char *file
     int infected = 0, len = 0, beenthere = 0;
     char *bol, *eol;
     struct onas_rcvln rcv;
+    struct onas_reply_state reply_state;
     STATBUF sb;
     int sockd                                                        = -1;
     int (*recv_func)(struct onas_rcvln *, char **, char **, int64_t) = NULL;
+
+    memset(&reply_state, 0, sizeof(reply_state));
 
     sockd = onas_get_sockd();
 
@@ -351,19 +375,10 @@ int onas_dsresult(CURL *curl, int scantype, uint64_t maxstream, const char *file
                 goto done;
 
             } else if (!memcmp(eol - 7, " FOUND", 6)) {
-                static char last_filename[PATH_MAX + 1] = {'\0'};
-                *(eol - 7)                              = 0;
-                *printok                                = 0;
+                *(eol - 7) = 0;
+                *printok   = 0;
 
-                if (scantype != ALLMATCH) {
-                    infected++;
-                } else {
-                    if (filename != NULL && strcmp(filename, last_filename)) {
-                        infected++;
-                        strncpy(last_filename, filename, PATH_MAX);
-                        last_filename[PATH_MAX] = '\0';
-                    }
-                }
+                infected += onas_count_virus_hit(&reply_state, scantype, filename);
 
                 if (filename) {
                     if (scantype >= STREAM) {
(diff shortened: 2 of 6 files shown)


clamonacc/client/protocol.c
```diff
@@ -237,10 +258,13 @@ int onas_dsresult(CURL *curl, int scantype, uint64_t maxstream, const char *file
     int infected = 0, len = 0, beenthere = 0;
     char *bol, *eol;
     struct onas_rcvln rcv;
+    struct onas_reply_state reply_state;
     STATBUF sb;
     int sockd                                                        = -1;
     int (*recv_func)(struct onas_rcvln *, char **, char **, int64_t) = NULL;
 
+    memset(&reply_state, 0, sizeof(reply_state));
+
```
sharkautarch, line 267: ~~seems like you went from `static char last_filename[PATH_MAX + 1]` to a local struct of array of char variable~~
~~I'm assuming the reason why this is a struct of char array instead of a plain char array is to prevent array to pointer implicit conversion when passing as a parameter to a function~~
~~Just add the `static` to the beginning of the variable decl, and remove `memset(&reply_state,...)`~~
~~static initialization will make it be initialized to zero:~~
~~https://cppreference.net/c/language/initialization.html~~
> objects with static and thread-local storage duration are empty-initialized 
> ...
>  ~~In some cases, an object is empty-initialized if it is not initialized explicitly, that is:~~
> ...
>  ~~all elements of arrays, all members of structs, and the first members of unions are empty-initialized, recursively, plus all padding bits are initialized to zero~~ 

~~Edit: this should probably be a static *thread_local*, otherwise weirdness may occur~~


**EDIT**: I just now realized that the `static char last_filename[PATH_MAX + 1]` thing originally being a static variable was dubious to begin with: `last_filename` is only really relevant when receiving scan results of multiple files in the same transaction
So `reply_state’ being a normal local variable makes sense

jhumlick (approved):

common/actions.c
```diff
@@ -58,6 +58,12 @@
 void (*action)(const char *) = NULL;
 unsigned int notmoved = 0, notremoved = 0;
 
+#ifdef _WIN32
+#define ACTION_COUNTER_INC(counter) InterlockedIncrement((volatile LONG *)&(counter))
+#else
+#define ACTION_COUNTER_INC(counter) __sync_add_and_fetch(&(counter), 1)
```
sharkautarch, line 64: Minor comment:
Note that according to this gcc documentation page: https://gcc.gnu.org/onlinedocs/gcc-15.2.0/gcc/_005f_005fatomic-Builtins.html
the `__sync_*` stuff are actually legacy builtins, and that it's advised to instead use the `__atomic_*` builtins when writing new code
AFAIK, the main difference between `__atomic_*` builtins and `__sync_*` builtins is that most `__sync_*` builtins enforce full memory barriers, whereas `__atomic_*` builtins have an extra parameter to specify memory ordering constraints. And also, the docs say that the `__sync_*` builtins only work on 1, 2, 4 and 8 byte memory locations, whereas `__atomic_*` builtins can additionally work on 16 byte memory locations if the target architecture supports 16 byte atomics. 

```suggestion
#define ACTION_COUNTER_INC(counter) __atomic_add_fetch(&(counter), 1, __ATOMIC_SEQ_CST)
```

clamonacc/client/client.c
```diff
@@ -71,6 +72,33 @@
...
+    pthread_mutex_lock(&onas_disconnect_lock);
+
+    if (onas_client_disconnected) {
+        logg(LOGG_INFO, "ClamClient: Connection to clamd re-established.\n");
+        onas_client_disconnected = false;
+    }
+
+    pthread_mutex_unlock(&onas_disconnect_lock);
+}
+
```
sharkautarch, line 101: Instead of doing pthread mutex lock & unlock, you could do atomic exchange on `onas_client_disconnected` (atomic exchange writes a value to memory and also retrieves the previous value before said write in one atomic transaction)  
In this situation, that would slightly improve performance by having one less place where threads could contend with each other momentarily, and also because mutex lock & unlock each perform at least one atomic Read-Modify-Write (RMW) operation, so using atomic exchange means we end up doing at least one less atomic RMW operation.

Note that in the suggested changes, I'm using the relaxed flavor of atomic exchange: this only matters for non-x86 cpus, where RMW atomics can be set to act as a partial memory barrier (acquire/release) or no memory barrier (relaxed) instead of a strict full memory barrier (seq_cst).
Using relaxed atomic exchange because there aren’t any writes to other global/shared memory locations by the worker threads. And the purpose of atomic acquire/release/seq_cst is enforcing an ordering between reads/writes to *different* variables. Atomic relaxed RMW should still enforce coherency with other atomic relaxed RMWs to the *same* exact memory address.

```suggestion
static bool onas_client_disconnected        = false;

#ifdef _WIN32
#define ATOMIC_EXCHANGE_BYTE_RELAXED(byte, val) InterlockedExchange8_nf(&(byte), (val))
#else
#define ATOMIC_EXCHANGE_BYTE_RELAXED(byte, val) __atomic_exchange_n(&(byte), (val), __ATOMIC_RELAXED)
#endif

static void onas_client_note_connection_failure(CURLcode curlcode)
{
    if (!ATOMIC_EXCHANGE_BYTE_RELAXED(onas_client_disconnected, true)) {
        logg(LOGG_ERROR, "ClamClient: Connection to clamd failed, %s.\n", curl_easy_strerror(curlcode));
    }
}

static void onas_client_note_connection_success(void)
{
    if (ATOMIC_EXCHANGE_BYTE_RELAXED(onas_client_disconnected, false)) {
        logg(LOGG_INFO, "ClamClient: Connection to clamd re-established.\n");
    }
}

```

PR #1716, opened 2026-04-23
Matcher performance enhancement and minor fixes

val-ms:
## Problem

At low AC depths, longer descendants can interfere with shorter suffix matches in two ways:

- non-leaf trie states were not inheriting fail-chain output lists, which could suppress valid suffix alerts
- the scan loop could replay fail-list heads that were already reachable, which could overcount suffix matches

Separately, repeated-prefix exact patterns and linked output heads were causing extra `ac_findmatch()` work in the hot path.

## Solution

This PR updates the AC matcher to preserve and filter suffix outputs more accurately and more cheaply:

- append `child->fail->list` onto non-leaf child states during trie construction, while avoiding duplicate head linkage
- avoid re-entering the fail-list in `cli_ac_scanbuff()` when the current head is already reachable from that chain
- precompute small per-head metadata for exact and nocase candidates so impossible matches can be rejected before calling `ac_findmatch()`
- shift a narrow class of repeated-prefix exact patterns onto a later static window to reduce unnecessary verification work

## Testing

- add a focused `clamscan` regression in `unit_tests/clamscan/allmatch_test.py`
- cover the `--dev-ac-only --dev-ac-depth 8` case with `Test.space.Dim.kk` and `Test.Dim`
- verify that scanning `hello Dim world ` still reports `Test.Dim` and does not report the longer non-matching signature

## Notes

In the comparison harness, the AC optimization reduced scan time for `libclamav.dll` from about `4.60s` to `3.63s` while preserving match counts. That result is workload-specific, so real-world gains will vary by file and signature mix.

### DIFF
--- a/libclamav/matcher-ac.c
+++ b/libclamav/matcher-ac.c
@@ -394,6 +394,27 @@ static uint16_t select_repeated_prefix_exact_window(const uint16_t *pattern, uin
     return best_start;
 }
 
+static void recalculate_prefix_metadata(struct cli_ac_patt *pattern)
+{
+    uint16_t i, j;
+
+    pattern->special_pattern = 0;
+    pattern->prefix_length[1] = 0;
+    pattern->prefix_length[2] = 0;
+
+    for (i = 0, j = 0; i < pattern->prefix_length[0]; i++) {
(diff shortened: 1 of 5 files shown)


val-ms:
@codex review

PR #1722, opened 2026-05-08
Tests: Stabilize freshclam mock mirror startup

val-ms:
The freshclam tests started a mock HTTP mirror on a fixed port and immediately ran freshclam without waiting for the server to bind. On slower or busier runners, especially macOS, that left a race where the client could connect before the server was ready or collide with stale port state.

Start the mock mirror on an ephemeral loopback port and report the bound address to the parent process through a readiness pipe. Prefer IPv4 loopback and fall back to IPv6 loopback, formatting IPv6 addresses for use in URLs. Tighten teardown so stale mock mirror processes are killed if they do not exit promptly.

### DIFF
--- a/unit_tests/freshclam_test.py
+++ b/unit_tests/freshclam_test.py
@@ -10,6 +10,7 @@ import os
 from pathlib import Path
 import platform
 import shutil
+import socket
 import unittest
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


rsundriyal (approved):
Looks good.

PR #1727, opened 2026-05-28
Add toggle to ignore special files on Unix systems

grische:
Currently, when a Unix socket, FIFO, or device node is passed as a scan target or file-list entry, clamscan logs "Not supported file type" and exits with error code 2, the same as if a file is unreadable.
With clamdscan the same warning also fires for every special file encountered during recursive directory walks, making filesystem scans (`clamdscan /`) routinely fail.

This PR adds three flags `--ignore-socket-errors`, `--ignore-pipe-errors`, `--ignore-device-errors` to both clamscan and clamdscan.
Enabling such a flag still logs the warning, but doesn't cause an return code 2 anymore, so users can distinguish unsupported special files from e.g. access errors. 

Fixes #1672


grische:
Improved the PR as suggested and downgraded the warnings to info messages.

grische:
@val-ms any chance you could have a look at this PR?

PR #1749, opened 2026-06-17
Add option to enable PCRE2 JIT for regex subsignature matching

j20120307:
# Add option to enable PCRE2 JIT for regex subsignature matching

## Summary

`cli_pcre_compile()` compiles regex subsignatures with `pcre2_compile()` and
`cli_pcre_match()` runs them with the interpreted `pcre2_match()`. The PCRE2
JIT compiler is never enabled, even when the linked PCRE2 library is built
with JIT support.

This PR adds an opt-in option to JIT-compile PCRE subsignatures:

- `PCREJit` in `clamd.conf`
- `--pcre-jit` for `clamscan`
- a new `CL_ENGINE_PCRE_JIT` engine setting

It is **disabled by default**, so existing behavior is unchanged on upgrade.
The option is plumbed exactly like the existing `PCREMatchLimit` /
`PCRERecMatchLimit` / `PCREMaxFileSize` settings (engine field, set/get,
settings copy/apply, option table).

## Implementation

- When enabled, `cli_pcre_compile()` performs a best-effort
  `pcre2_jit_compile(pd->re, PCRE2_JIT_COMPLETE)` after the match limits are
  set. `pcre2_match()` automatically dispatches to the JIT code when a pattern
  has been JIT-compiled, so no change to the normal match call is required.
- If JIT compilation fails (PCRE2 built without JIT, or a construct the JIT
  does not support), the failure is logged via `cli_dbgmsg()` and the pattern
  is matched by the interpreter as before.
- To guarantee identical results to the interpreter when the option is on,
  `cli_pcre_match()` retries a match with `PCRE2_NO_JIT` if the JIT matcher
  returns `PCRE2_ERROR_JIT_STACKLIMIT`. The JIT uses a fixed stack and, unlike
  the interpreter, does not grow on demand; this retry keeps behavior
  equivalent rather than surfacing a JIT-only error.

## Motivation

When the Aho-Corasick prefilter promotes a large number of PCRE subsignatures
for full evaluation against a large buffer, the interpreted matcher dominates
scan time. The literal anchors of many filename-style phishing subsignatures
are short and commonly appear in text-heavy input, so a single buffer can
legitimately trigger several hundred PCRE evaluations, each scanning the full
buffer. JIT executes the identical automaton as native code, so this is a pure
latency improvement with no effect on detection.

## Benchmark

Measured with two builds of the same revision — option off vs option on —
using an identical signature database and scan configuration (`ScanArchive`
disabled, cache disabled). The test input was a ~37 MiB archive composed of
many small text files, which the prefilter promotes to roughly several hundred
PCRE subsignatures.

| Path | Default (off) | `PCREJit`/`--pcre-jit` on | Speedup |
|------|--------------:|--------------------------:|--------:|
| `clamscan` (standalone, includes DB load) | ~22.0 s | ~10.5 s | ~2.1x |
| `clamdscan` (resident daemon, scan-only)  | ~16.6 s | ~5.4 s  | ~3.1x |

Phase decomposition (daemon, scan-only) attributes the gain to the PCRE phase:

| Phase | Off | On |
|-------|----:|---:|
| Aho-Corasick / Boyer-Moore literal prefilter | ~0.9 s | ~0.9 s |
| PCRE full-match phase | ~15.8 s | ~4.5 s |
| PCRE subsignatures evaluated | identical | identical |

Detection results were verified identical with the option on and off (EICAR
detected in both; the benign sample reported clean in both).

## Testing

- Ran the `libclamav` unit test suite (`ENABLE_TESTS=ON`) on both this branch
  and the unmodified base revision: both report `Checks: 1235` with an
  identical set of results, so this change introduces no new test failures. The
  `pcre`, `regex`, and `matchers` suites — which exercise the changed compile
  and match paths — pass. `check_matchers.c` was updated to build with the JIT
  path enabled so the suite validates that matches are unchanged.
- Confirmed the option is off by default (scan time and behavior identical to
  the prior release), `--pcre-jit=yes` engages JIT, and `--pcre-jit=no` is
  identical to the default.
- Verified detection is unchanged in both modes with a logical signature
  containing a PCRE subsignature: matching input is detected and non-matching
  input is reported clean, identically with the option on and off. A
  known-malicious test file is detected and benign input reported clean in both
  modes.
- Confirmed both `clamscan` and `clamdscan`/`clamd` honor the option.

### A note on benchmarking

The repository has no in-tree performance/benchmark test harness — the test
suite is correctness-only — so the timing numbers above were measured manually
by timing identical scans with the option off and on. A wall-clock assertion
was intentionally not added, since it would be flaky across CI hardware. The
per-signature `--statistics=pcre` diagnostic can be used to reproduce the
PCRE-phase timing breakdown.

## Compatibility

- Disabled by default; no behavior change unless explicitly enabled.
- New `CL_ENGINE_PCRE_JIT` is appended to the end of the `cl_engine_field`
  enum, so existing enumerator values are unchanged.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,17 @@ ClamAV 1.6.0 includes the following improvements and changes:
 
 ### Other improvements
 
+- Added an option to enable the PCRE2 JIT compiler for regex subsignature
+  matching. When enabled, patterns are JIT-compiled at database load and matched
+  with native code, significantly reducing scan time for content that triggers
+  many PCRE subsignatures. The new option is `PCREJit` in `clamd.conf` and
+  `--pcre-jit` for `clamscan`, and is disabled by default. It requires the
+  linked PCRE2 library to be built with JIT support; otherwise matching
+  transparently falls back to the interpreter. Detection results are unchanged.
+  Note: PCRE2 does not enforce the depth/recursion limit (`PCRERecMatchLimit`)
+  for JIT matches, though `PCREMatchLimit` is still enforced; leave the option
+  disabled if you rely on `PCRERecMatchLimit` to bound recursive backtracking.
+
 ### Bug fixes
 
 ### Acknowledgments
--- a/clamd/clamd.c
+++ b/clamd/clamd.c
@@ -741,6 +741,14 @@ int main(int argc, char **argv)
             }
         }
 
+        if ((opt = optget(opts, "PCREJit"))->active) {
+            if ((ret = cl_engine_set_num(engine, CL_ENGINE_PCRE_JIT, opt->enabled))) {
+                logg(LOGG_ERROR, "cli_engine_set_num(PCREJit) failed: %s\n", cl_strerror(ret));
+                cl_engine_free(engine);
+                return 1;
+            }
+        }
+
         if ((opt = optget(opts, "PCRERecMatchLimit"))->active) {
             if ((ret = cl_engine_set_num(engine, CL_ENGINE_PCRE_RECMATCH_LIMIT, opt->numarg))) {
                 logg(LOGG_ERROR, "cli_engine_set_num(PCRERecMatchLimit) failed: %s\n", cl_strerror(ret));
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -363,6 +363,7 @@ void help(void)
     mprintf(LOGG_INFO, "    --pcre-match-limit=#n                Maximum calls to the PCRE match function.\n");
     mprintf(LOGG_INFO, "    --pcre-recmatch-limit=#n             Maximum recursive calls to the PCRE match function.\n");
     mprintf(LOGG_INFO, "    --pcre-max-filesize=#n               Maximum size file to perform PCRE subsig matching.\n");
+    mprintf(LOGG_INFO, "    --pcre-jit[=yes/no(*)]               JIT-compile PCRE subsignatures for faster matching (requires PCRE2 JIT support; pcre-recmatch-limit is not enforced for JIT matches).\n");
     mprintf(LOGG_INFO, "    --disable-cache                      Disable caching and cache checks for hash sums of scanned files.\n");
     mprintf(LOGG_INFO, "    --hash-hint                          The file hash so that libclamav does not need to calculate it.\n");
     mprintf(LOGG_INFO, "                                         The type of hash must match the '--hash-alg'.\n");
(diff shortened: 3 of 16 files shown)


j20120307:
Thanks — good catch. Confirmed against the PCRE2 docs: `pcre2_set_recursion_limit()` is the deprecated alias for `pcre2_set_depth_limit()`, and per `pcre2jit(3)` "the PCRE2_ERROR_DEPTHLIMIT error code is never returned when JIT matching is used", so `PCRERecMatchLimit` is not honored for JIT matches.

Worth noting the overall `PCREMatchLimit` (`pcre2_set_match_limit`) **is** still enforced under JIT, so total match work stays bounded; it's specifically the separate depth/recursion bound that's lost.

Since this is an opt-in feature (disabled by default), I've documented the tradeoff rather than changing the match path: the caveat is now in the `PCREJit`/`--pcre-jit` option help, `clamd.conf.sample`, `NEWS.md`, and a code comment at the JIT-compile site, advising operators who rely on `PCRERecMatchLimit` to leave JIT disabled.

Happy to instead force the interpreter when `PCRERecMatchLimit` is set below the default if you'd prefer the limit always be enforceable.

juancarlosgd:
Independent verification of this patch, FWIW — evaluating it for a production
security-scanning pipeline (SQS-driven async malware scan, real user-uploaded
Office/PDF files, AWS Lambda arm64).

**Setup:** applied cleanly against `clamav-1.5.3` (no rebase needed — the 4 core
matcher files, `regex_pcre.c/h` + `matcher-pcre.c/h`, are byte-identical between
this PR's base and the 1.5.3 tag; only `NEWS.md` conflicted, dropped from the
vendored patch). Compiled from source (CMake), Rust toolchain via `rustup` (apt's
`cargo`/`rustc` on Debian bookworm is too old for the `Cargo.lock` v4 format).

**Benchmark:** real 74.6MB PPTX (not synthetic/random bytes — file structure
matters a lot for PCRE-heavy scans, flat entropy blobs are misleading), same
container, CPU/memory constrained to match our real Lambda allocation
(`--memory=4096m --cpus=2.32`, i.e. AWS's ~1769MB-per-vCPU ratio):

| Config | Time |
|---|---|
| `PCREJit no` (default) | 56.6-57.2s |
| `PCREJit yes` | 20.6-20.9s |

**~2.75x**, consistent across 3 separate runs (different content each time, to
avoid `clamd`'s hash-based result cache) and across two different CPU/memory
tiers we tested (3072MB/1.74vCPU and 4096MB/2.32vCPU — same speedup regardless
of core count, as expected for a compile-vs-interpret improvement).

Detection verified identical with the option on/off: EICAR detected correctly
in both modes, and the real test file gave the same clean verdict both ways.

Happy to share more detail if useful for review.

PR #1754, opened 2026-07-02
libclamav: update bundled YARA parser behavior

jhumlick:
Refresh ClamAV's YARA parser/lexer support toward current YARA syntax, including hex-string comments, private strings, octal literals, and clearer errors for unsupported modifiers.

Also harden YARA VM edge cases around fmap reads, divide/modulo/shift operations, undefined string operands, and add clamscan sanity tests for the new parser/runtime behavior.

### DIFF
--- a/libclamav/filtering.c
+++ b/libclamav/filtering.c
@@ -298,6 +298,7 @@ static inline unsigned char spec_ith_char(const struct char_spec *spec, unsigned
         cc0   = spec0->negative ? 0 : c0;            \
         cc1   = spec1->negative ? 0 : c1;            \
         for (; cc0 <= c0end; cc0++) {                \
+            cc1 = spec1->negative ? 0 : c1;          \
             for (; cc1 <= c1end; cc1++) {            \
                 uint16_t a = cc0 | (cc1 << 8);       \
                 if (spec0->negative && cc0 == c0)    \
--- a/libclamav/filtering.c
+++ b/libclamav/filtering.c
@@ -524,6 +524,21 @@ int filter_add_acpatt(struct filter *m, const struct cli_ac_patt *pat)
                 spec->end   = 0xf0 | spec->start;
                 spec->step  = 0x10;
                 break;
+            case CLI_MATCH_NOT_BYTE:
+                spec->start = spec->end = (uint8_t)p;
+                spec->step              = 1;
+                spec->negative          = 1;
+                break;
+            case CLI_MATCH_NOT_NIBBLE_HIGH:
+            case CLI_MATCH_NOT_NIBBLE_LOW:
+                /*
+                 * The filter may over-approximate, but must not reject a
+                 * candidate accepted by the full matcher.
+                 */
+                spec->start = 0x00;
+                spec->end   = 0xff;
+                spec->step  = 1;
+                break;
             default:
                 cli_errmsg("filtering: unknown wildcard character: %d\n", p);
                 return -1;
--- a/libclamav/matcher-ac.c
+++ b/libclamav/matcher-ac.c
@@ -463,11 +463,7 @@ cl_error_t cli_ac_addpatt(struct cli_matcher *root, struct cli_ac_patt *pattern)
         }
     }
 
-    /*
-     * Negation may separate every literal byte, as in 41~0042. In that case,
-     * use one literal trie transition and verify the full predicate pattern.
-     */
-    if (len < root->ac_mindepth && !(pattern->has_negation && len == 1)) {
+    if (len < root->ac_mindepth) {
         /* cli_errmsg("cli_ac_addpatt: Signature for %s is too short\n", pattern->virname); */
         return CL_EMALFDB;
     }
@@ -623,17 +619,6 @@ static int ac_maketrans(struct cli_matcher *root)
                     return ret;
             }
         }
-    }
-
-    /*
-     * A negated pattern may use a one-byte literal anchor when no multi-byte
-     * literal window is available. These final root children were leaves throughout
-     * transition construction, so connect them only after the BFS is done.
-     */
-    for (i = 0; i < 256; i++) {
-        node = ac_root->trans[i];
-        if (node != ac_root && IS_FINAL(node) && IS_LEAF(node))
-            node->trans = node->fail->trans;
     }
 
     return CL_SUCCESS;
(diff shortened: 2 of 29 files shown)


libclamav/yara_grammar.y
val-ms, line 1: We should update the YARA copyright at the top of this, and any other yara project derived source files. 
If it's now based off of latest Yara, the license on the yara project changed and is no longer apache-2.0.  It is now BSD-3-Clause. 
https://github.com/VirusTotal/yara/blob/master/COPYING

val-ms (changes requested):

libclamav/matcher-ac.c
```diff
@@ -2369,6 +2426,21 @@ inline static int ac_uicmp(uint16_t *a, size_t alen, uint16_t *b, size_t blen, i
                         return (b[i] & 0x0f) - (a[i] & 0x0f);
                     }
                     break;
+                case CLI_MATCH_NOT_BYTE:
+                    if ((a[i] & 0xff) != (b[i] & 0xff)) {
```
val-ms, line 2430: **High - overlapping negated alternatives can suppress required backtracking**

`ac_uicmp()` returns nonzero for two `CLI_MATCH_NOT_BYTE` units with different excluded values, but those predicates overlap on 254 byte values. `ac_addspecial_add_alt_node()` therefore leaves variable-length branches `unique`, and `ac_findmatch_special()` can commit to the first matching branch without retrying another length.

For `{ 41 42 ( ~00 | ~01 44 ) 45 46 }` over bytes `41 42 02 44 45 46`, `~00` matches first but causes the suffix to fail; `~01 44` would match, but it is never tried. YARA permits negated byte tokens inside alternatives, including branches of differing lengths.

Please make alternate comparison overlap-aware or reject variable-length alternatives containing negation, and add a regression test for this case. Inkie currently takes the latter approach rather than accepting a rule with incorrect match semantics.

libclamav/str.c
```diff
@@ -72,79 +72,118 @@ static inline int cli_hex2int(const char c)
...
+    unsigned int i = 0;
     int c;
 
-    for (i = 0; i < len; i += 2) {
+    while (i < len) {
         val = 0;
 
-        if (hex[i] == '?' && hex[i + 1] == '?') {
-            val |= CLI_MATCH_IGNORE;
+        if (hex[i] == '~') {
```
val-ms, line 85: **Medium - native signature modifiers do not account for three-character negated units**

The existing `wide` preprocessor in `cli_sigopts_handler()` still advances through the body two source characters at a time. Introducing `~HH`/`~H?`/`~?H` as three-character matcher units shifts all following pairs. For example, `41~0042::w` should widen to `4100~00004200`, but the current transform produces `4100~00004002`, so the signature silently matches the wrong bytes.

Please teach the wide transform to copy a negated matcher unit atomically and append `00`, or reject this modifier combination. The `nocase` interaction should also be defined or rejected: current folding applies only to exact literal units, so `~41::i` excludes `A` but still accepts `a`. Inkie supports wide negation and explicitly rejects nocase negation.

libclamav/matcher-ac.c
```diff
@@ -3151,7 +3282,11 @@ cl_error_t cli_ac_addsig(struct cli_matcher *root, const char *virname, const ch
             ppos = nzpos;
         }
 
-        if (plen < root->ac_mindepth) {
+        /*
+         * A negated pattern may have no adjacent literal pair. One literal is
+         * still a valid anchor because the remaining units are verified later.
+         */
+        if (plen < root->ac_mindepth && !(new->has_negation && plen == 1)) {
```
val-ms, line 3289: **Medium - one-byte fallback never examines a sole trailing literal**

This permits a depth-one literal anchor, but the preceding search bound still assumes `ac_mindepth`: `pend = length - ac_mindepth + 1`. For `~0041` at the default depth of 2, only index 0 is examined and the trailing `41` literal is never considered, while the reversed form `41~00` can load.

Please adjust the search range when the one-byte fallback is active and add scan tests for both orientations. The current tests retain a multi-byte exact window, so they do not exercise this path or the new root-leaf transition completion.

val-ms (changes requested):
The matcher-unit representation is the right architecture and the basic whole-byte/nibble negation predicates look correct. These remaining cases can silently miss valid signatures or change native signature semantics, so I am requesting changes for the inline findings.

val-ms (commented):
A second independent pass over parser conversion, matcher construction, verification, and prefiltering found two additional concerns. The first is another false-negative path; the second is a performance-policy regression introduced by the one-byte fallback.

libclamav/filtering.c
```diff
@@ -524,6 +524,21 @@ int filter_add_acpatt(struct filter *m, const struct cli_ac_patt *pat)
                 spec->end   = 0xf0 | spec->start;
                 spec->step  = 0x10;
                 break;
+            case CLI_MATCH_NOT_BYTE:
```
val-ms, line 527: **High - a leading whole-byte negation produces an incomplete prefilter**

This maps `CLI_MATCH_NOT_BYTE` to `negative = 1`, but the existing `SPEC_FOREACH` macro initializes `cc1` only once, outside its outer `cc0` loop. When the first q-gram unit is negative, the inner loop is exhausted during the first `cc0` iteration and is never reset. The filter therefore emits at most one of the 255 byte values accepted by `~HH` (and emits none when that first value is the excluded byte).

For example, a pattern such as `~010000` can use the trailing `00 00` as its valid AC anchor, but if the filter chooses the leading negated q-gram, bytes `02 00 00` can be discarded before the full matcher sees them even though `~01` accepts `02`. The current tests all provide a more attractive adjacent literal pair, so they do not exercise this path.

Please either reset the inner enumeration for every outer value or conservatively over-approximate whole-byte negation as is already done for negated nibbles. Add a prefilter-enabled regression test where an accepted byte is not the first enumerated value.

libclamav/matcher-ac.c
```diff
@@ -463,7 +463,11 @@ cl_error_t cli_ac_addpatt(struct cli_matcher *root, struct cli_ac_patt *pattern)
         }
     }
 
-    if (len < root->ac_mindepth) {
+    /*
+     * Negation may separate every literal byte, as in 41~0042. In that case,
+     * use one literal trie transition and verify the full predicate pattern.
+     */
+    if (len < root->ac_mindepth && !(pattern->has_negation && len == 1)) {
```
val-ms, line 470: **Medium - the negation-only one-byte exception bypasses the configured AC minimum depth**

This accepts low-selectivity matcher heads such as `41~00` even though the equivalent one-literal-plus-wildcard shape is rejected by `ac_mindepth`. A database containing many such rules on a common byte can attach a large candidate list to one root transition and force full-pattern verification at nearly every occurrence. The minimum depth exists to prevent this class of matcher cost, and the PR does not include an adversarial performance test for the exception.

Inkie keeps negation as a native predicate but still requires the normal exact-window minimum. Unless ClamAV intentionally wants to relax that policy, the safer fix is to remove this exception and the root-leaf transition special case, and reject negated patterns with no sufficiently long exact anchor. If the exception is retained, it needs a benchmark demonstrating bounded behavior with many rules sharing a frequent one-byte anchor.

PR #1758, opened 2026-07-05
CLAM-3049 perf(matcher): avoid slow AC heads with repeated bytes

val-ms:
## Summary

This PR extends ClamAV's existing AC matcher prefix-shift optimization so patterns that begin with a repeated exact byte do not use that low-selectivity repeated-byte window as the trie head.

The existing matcher already shifts poor heads when the first AC window contains wildcards or is all zero bytes. This adds the same treatment for repeated exact-byte prefixes, such as long runs of `0x20` spaces, while preserving the skipped bytes as a backwards-checked prefix.

## Problem

Some production hex signatures start a sub-pattern with many copies of the same byte. When that byte is common in scanned content, the AC trie produces a large number of candidate branches before the rest of the pattern can reject the match.

The motivating case is a real sample scanned with `main.ldb`, where several Valyria signatures start with repeated spaces. The scan spent minutes in this matcher path even though the file ultimately scanned clean.

Representative slow signatures from `main.ldb` included:

- `Xls.Dropper.Valyria-6680530-0`
- `Xls.Dropper.Valyria-6680531-0`
- `Doc.Dropper.Valyria-6680532-0`
- `Doc.Dropper.Valyria-6680533-0`
- `Doc.Malware.Valyria-6691358-0`

## Solution

The new helper detects when a pattern starts with a run of the same static byte at least as long as the AC trie depth. It then scans for a later exact static window of the same depth and chooses the best window using these tie-breaks:

1. fewer occurrences of the repeated prefix byte,
2. more distinct bytes,
3. later position when the first two are tied.

The skipped leading bytes are stored as the pattern prefix and verified by the existing backwards-match path, so match semantics remain unchanged. The previous wildcard/all-zero prefix-shift path now uses the same shared prefix helper.

## Benchmark Results

Release builds from the same base commit were compared on macOS with local `main.ldb` and a known slow sample. Both scans returned `OK`.

Focused five-signature Valyria repro:

- main: `5.846 sec` ClamAV summary, `5.85 real`
- branch: `2.134 sec` ClamAV summary, `2.14 real`
- result: about `2.7x` faster

Full `main.ldb` repro:

- main: `262.176 sec` ClamAV summary, `262.18 real`, `254.78 user`
- branch: `3.302 sec` ClamAV summary, `3.31 real`, `2.20 user`
- result: about `79x` faster

Benchmark command shape:

```sh
clamscan -d ~/main-unpacked/main.ldb ~/Downloads/c72ca7651909b7bd89bc05dbd18be88fe8a22c4cbff0b837d016ea1be9caf277
```

Note: I can't provide the test sample here for privacy reasons.

### DIFF
--- a/libclamav/matcher-ac.c
+++ b/libclamav/matcher-ac.c
@@ -248,6 +248,126 @@ static int sort_heads_by_partno_fn(const void *a, const void *b)
     return 0;
 }
 
+static bool ac_static_window_byte(uint16_t value, uint8_t *byte)
+{
+    switch (value & CLI_MATCH_METADATA) {
+        case CLI_MATCH_CHAR:
+        case CLI_MATCH_NOCASE:
+            if (byte) {
+                *byte = (uint8_t)(value & 0xff);
+            }
+            return true;
+        default:
+            return false;
+    }
+}
+
+static bool ac_select_repeated_prefix_exact_window(const uint16_t *pattern, uint16_t length, uint8_t ac_maxdepth, uint16_t *ppos)
+{
+    const size_t depth = ac_maxdepth;
+    uint8_t repeated;
+    size_t repeated_run_len = 1;
+    size_t best_start       = 0;
+    unsigned int best_repeated_count;
+    unsigned int best_distinct_count = 0;
+    bool found                        = false;
+    size_t start;
+
+    if (!pattern || !ppos || 0 == depth || length <= depth) {
+        return false;
+    }
+
+    if (!ac_static_window_byte(pattern[0], &repeated)) {
+        return false;
+    }
+
+    while (repeated_run_len < length) {
+        uint8_t byte;
+
+        if (!ac_static_window_byte(pattern[repeated_run_len], &byte) || byte != repeated) {
+            break;
+        }
+        repeated_run_len++;
+    }
+    if (repeated_run_len < depth) {
+        return false;
+    }
+
+    best_repeated_count = (unsigned int)depth + 1;
+
+    for (start = 1; start <= (size_t)length - depth; start++) {
+        bool distinct[256]             = {false};
+        unsigned int repeated_count    = 0;
+        unsigned int distinct_count    = 0;
+        bool valid                     = true;
+        size_t i;
+
+        for (i = start; i < start + depth; i++) {
+            uint8_t byte;
+
+            if (!ac_static_window_byte(pattern[i], &byte)) {
+                valid = false;
+                break;
+            }
+            if (byte == repeated) {
+                repeated_count++;
+            }
+            if (!distinct[byte]) {
+                distinct[byte] = true;
+                distinct_count++;
+            }
+        }
+
+        if (!valid || repeated_count >= depth) {
+            continue;
+        }
+
+        if (!found ||
+            repeated_count < best_repeated_count ||
+            (repeated_count == best_repeated_count && distinct_count > best_distinct_count) ||
+            (repeated_count == best_repeated_count && distinct_count == best_distinct_count && start > best_start)) {
+            found                = true;
+            best_start           = start;
+            best_repeated_count  = repeated_count;
+            best_distinct_count  = distinct_count;
+        }
+    }
+
+    if (!found) {
+        return false;
+    }
+
+    *ppos = (uint16_t)best_start;
+    return true;
+}
+
+static void ac_shift_pattern_prefix(struct cli_ac_patt *pattern, uint16_t ppos)
+{
+    uint16_t i;
+    uint16_t j;
+
+    pattern->prefix           = pattern->pattern;
+    pattern->prefix_length[0] = ppos;
+    for (i = 0, j = 0; i < pattern->prefix_length[0]; i++) {
+        if ((pattern->prefix[i] & CLI_MATCH_WILDCARD) == CLI_MATCH_SPECIAL)
+            pattern->special_pattern++;
+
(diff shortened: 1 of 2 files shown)


val-ms:
@TheRaynMan point of interest: this optimization is a backport from one Codex found for me when optimizing Inkie's modified Rust port of the ClamAV AC matcher. Really exciting to me I already had a fix available in that other project for this real world issue.

libclamav/matcher-ac.c
```diff
@@ -248,6 +248,126 @@ static int sort_heads_by_partno_fn(const void *a, const void *b)
     return 0;
 }
 
+static bool ac_static_window_byte(uint16_t value, uint8_t *byte)
+{
+    switch (value & CLI_MATCH_METADATA) {
+        case CLI_MATCH_CHAR:
+        case CLI_MATCH_NOCASE:
```
TheRaynMan, line 255: This would not be case-insensitive ascii-match ... 0x41 through 0x5A matching 0x61 through 0x7A ...

TheRaynMan: So here is a review question: Should the NOCASE part of this switch allow for spotting a long string of AaaAaaAAAAAaa or similar, or only on same fixed exact byte (more common)?

val-ms: I agree with you about nocase repeating patterns being a concern. I asked codex to evaluate if there's an issue with the PR in this regard and it told me:

The question is:

> Should the NOCASE part of this switch allow for spotting a long string of `AaaAaaAAAAAaa` or similar, or only on same fixed exact byte?

My read: **yes, for a nocase signature, mixed `A/a` runs should count as a repeated run.** And the current code already does that, because of when this helper runs.

In `cli_ac_addsig()`, nocase patterns are canonicalized before the repeated-prefix helper is called:

```c
if (sigopts & ACPATT_OPTION_NOCASE) {
    for (i = 0; i < new->length[0]; i++)
        if ((new->pattern[i] & CLI_MATCH_METADATA) == CLI_MATCH_CHAR) {
            new->pattern[i] = CLI_NOCASE(new->pattern[i] & 0xff);
            new->pattern[i] += CLI_MATCH_NOCASE;
        }
}
```

So a nocase pattern like `AaaAaaAAAAAaa` is represented as repeated canonical `a` bytes with `CLI_MATCH_NOCASE` metadata. Then `ac_static_window_byte()` returns that canonical byte, so the repeated-prefix selection sees it as a repeated `a` run.

Matching remains correct after shifting:

- The shifted AC trie still handles nocase through `cli_ac_addpatt_recursive()`, which adds the alternate ASCII-case transition for alpha bytes.
- The stored prefix is checked later by `AC_MATCH_CHAR`, where `CLI_MATCH_NOCASE` compares the canonical pattern byte against `CLI_NOCASE(buffer_byte)`.
- So moving `A/a` bytes into the prefix does not make them case-sensitive.

The important distinction is:

- `AaaAaa...` with a nocase signature option: should be treated as repeated, and currently is.
- `AaaAaa...` without nocase: should not be treated as repeated, because those are fixed distinct bytes in a case-sensitive pattern.

I do think the naming/comment could be clearer. `ac_select_repeated_prefix_exact_window()` says “exact,” but it actually handles static canonical bytes, including already-folded `CLI_MATCH_NOCASE` bytes. A good response would be:

```text
Yes, for a nocase signature I think mixed A/a runs should be eligible. The helper sees CLI_MATCH_NOCASE after cli_ac_addsig() has already canonicalized the byte with CLI_NOCASE(), so AaaAaa is treated as the same repeated canonical byte. Prefix verification still uses CLI_MATCH_NOCASE matching, so shifting those bytes does not change semantics.

The helper name/comment may be misleading though; “exact” really means a static byte window after ClamAV’s pattern normalization. I can add a comment or rename it to make that invariant explicit.
```

I would not change behavior unless you want to make the helper self-contained by applying `CLI_NOCASE()` in the `CLI_MATCH_NOCASE` case too. That would be defensive, but redundant under the current construction path.

csbg-draynor: Fair, needed confirmation it was dealt with elsewhere before reaching this code path. No comment explained this.

jhumlick (dismissed):

libclamav/matcher-ac.c
```diff
@@ -248,6 +248,126 @@ static int sort_heads_by_partno_fn(const void *a, const void *b)
...
+                repeated_count++;
+            }
+            if (!distinct[byte]) {
+                distinct[byte] = true;
+                distinct_count++;
+            }
+        }
+
+        if (!valid || repeated_count >= depth) {
+            continue;
```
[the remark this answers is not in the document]
val-ms: Fixed in 3042685ed. The selector now rejects replacement windows that contain only one distinct byte value, so `ffffff000000` shifts to the transition window `ff0000` instead of replacing the `ffffff` head with `000000`.

I also added a structural matcher regression that asserts the selected prefix length and trie window, then verifies the complete signature still matches.

Validation:

- `make -j12 check_clamav`
- `CK_RUN_CASE=matchers ./build/unit_tests/check_clamav` (9 checks passed)
- clang-format check on both changed files

The equivalent Inkie fix is in [Cisco-Talos/inkie#76](https://github.com/Cisco-Talos/inkie/pull/76); its 8 MiB zero-filled probe improved from 121.2 ms to 21.8 ms while the medium real-world corpus remained neutral.

val-ms:
@codex review

jhumlick (dismissed):

jhumlick (approved):

PR #1764, opened 2026-07-10
Tests: widen PNG decoder Valgrind suppression

val-ms:
## Summary

- Widen the existing fuzzy image hash PNG decoder Valgrind suppression so it matches both `std::io::cursor::Cursor` and `core::io::cursor::Cursor` stack spellings.
- Keep the suppression scoped to the known `fuzzy_hash_calculate_image` path.

## Rationale

Debian 11 arm64 Valgrind jobs with newer Rust output report the same known PNG decoder conditional-use warning, but the current suppression only matches the older `std::io` symbol spelling. The newer stack uses `core::io`, so Valgrind reads the suppression file but does not apply it.

## Validation

- `git diff --check`


val-ms:
For context, this fixes the clamav package build pipeline which started failing a previously suppressed valgrind issue on debian:11 arm64. It only started failing this AM, so I suspect it's related to the new Rust version publkished today. 

I observed the failure on both of: 
* https://github.com/Cisco-Talos/clamav/pull/1758
* https://github.com/Cisco-Talos/clamav/pull/1761

So I figure we should fix it here and then rebase all PRs.

TheRaynMan (approved):
Simple enough suppression for valgrind alert. Approved.

PR #1769, opened 2026-07-11
unit_tests: Don't set TMP

sebastianas:
Setting TMP to CMAKE_CURRENT_BINARY_DIR can result in a "long" path such as "/build/reproducible-path/clamav-1.4.5+dfsg/obj-arm-linux-gnueabihf/unit_tests" together with the temporary filename the testsuite can abort with "OSError: AF_UNIX path too long" during "TC.mock_mirror.start()" because the resulting unix socket name gets too long.

That is because python3.14 switch from fork to a forkserver in Process.start().

Without setting, it ends up under /tmp/ which is brief enough.

### DIFF
--- a/unit_tests/CMakeLists.txt
+++ b/unit_tests/CMakeLists.txt
@@ -214,7 +214,10 @@ else()
 
     set(SOURCE             ${CMAKE_SOURCE_DIR})
     set(BUILD              ${CMAKE_BINARY_DIR})
-    set(TMP                ${CMAKE_CURRENT_BINARY_DIR})
+    set(TMP                $ENV{TMPDIR})
+    if(NOT TMP)
+        set(TMP "/tmp")
+    endif()
 
     set(NEW_PATH "$ENV{PATH}")
 


val-ms (changes requested):
Thank you for your patience while we ran this through the different test environments. The underlying problem described by this PR is real, and moving the test temporary directories out of potentially long build paths is the right direction. There are two issues with the current implementation that we need to address before merging, though.

First, using the macOS `TMPDIR` directly introduces a path-alias mismatch. macOS supplies paths under `/var/folders/...`, while ClamAV canonicalizes the same paths to `/private/var/folders/...`. The quarantine operations themselves succeed, but the tests construct their expected copy/move messages from the non-canonical `TC.path_tmp` value and therefore do not recognize ClamAV's canonical destination path.

This consistently fails:

- `test_quarantine_directory_replacement_does_not_redirect_copy_target`
- `test_quarantine_directory_replacement_does_not_redirect_move_target`

I reproduced both failures locally, and the macOS CI failure shows the same behavior. The output has the test paths under `/var/folders/...` and ClamAV reporting successful quarantine destinations under `/private/var/folders/...`. This is a test-path consistency problem rather than a failure of the quarantine operation itself.

Second, `$ENV{TMPDIR}` is evaluated when CMake configures the build and is then embedded literally in the generated CTest properties as `TMP`. If configuration and testing occur at different times or in different sandboxes, that directory may have been removed or may no longer be writable. Because `testcase.py` passes `TMP` explicitly to `tempfile.mkdtemp()`, every Python test then fails during setup instead of allowing Python to choose another usable temporary directory. I reproduced that behavior with a removed configure-time directory.

My recommendation is:

1. Do not add `TMP` to the CTest environment on non-Windows. Let Python's `tempfile` module choose from the environment and platform defaults when each test process actually runs. The existing Windows-specific `TMP` behavior can remain conditional.
2. In `testcase.py`, create the directory without an explicit `dir=os.getenv("TMP")` on non-Windows and retain the canonical path, for example:

   ```python
   cls.path_tmp = Path(
       tempfile.mkdtemp(prefix=(cls.__name__ + "-"))
   ).resolve()
   ```

   Canonicalizing the returned path keeps the paths passed to ClamAV consistent with the paths ClamAV reports on macOS. The exact implementation can vary, but we need both runtime selection and canonical path handling.
3. After the update, we will rerun the full CTest suite in macOS CI to confirm that the two quarantine TOCTOU tests pass; no local macOS environment is expected from you. Separately, a platform-neutral regression check for a removed or unusable configure-time `TMPDIR` would be valuable if practical, since that verifies the runtime-selection behavior without requiring macOS.

The rest of the behavior looks sound: `mkdtemp()` continues to give each test class a unique directory, parallel test isolation is preserved, and the existing cleanup still removes those directories unless `KEEPTEMP` is enabled.

Thanks again for your patience and for working through the portability details here.

PR #1780, opened 2026-07-28
Jenkinsfile: throttle concurrent builds

val-ms:


val-ms:
Superseded by #1782.

The replacement uses the runtime `throttle()` step so newly discovered multibranch jobs are throttled before their first agent allocation. A new PR and head branch were required to exercise the first-run behavior described in jenkinsci/throttle-concurrent-builds-plugin#448.

PR #1782, opened 2026-07-29
Jenkinsfile: throttle concurrent builds

val-ms:
This replaces #1780 with runtime throttling that also covers the first build of a newly discovered multibranch job.

A Jenkinsfile-defined `throttleJobProperty` is not present when Jenkins queues the first run of a new branch or pull-request job. Consequently, that first run can bypass the global category limit, as described in [jenkinsci/throttle-concurrent-builds-plugin#448](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/issues/448).

Use the plugin's runtime `throttle()` step instead. The pipeline now has no top-level agent. A parent sequential stage named `ClamAV pipeline` enters `throttle(['clamav'])` before allocating the existing `default` agent, then runs the existing Generate Tarball, Build, and Tests stages within that agent and throttle scope.

This structure holds one category slot for the complete orchestration run, preserves the existing internal Tests parallelism, and prevents first runs from acquiring an executor while another ClamAV orchestration run is active.

Validation:

- `git diff --check`
- Confirmed the branch contains one commit changing only `Jenkinsfile`
- Matched the plugin's tested Declarative first-run pattern: top-level `agent none`, then stage-level `agent` and `options { throttle(...) }`
- Used Jenkins-supported sequential nested stages so the throttle remains active for the whole run
- Confirmed that no `agent` is declared on the `Tests` stage containing `parallel`
- Confirmed the obsolete `throttleJobProperty` is absent

The replacement PR intentionally uses a new head branch so Jenkins creates a new multibranch PR job and exercises first-run throttling.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -58,176 +58,185 @@ properties(
 
 pipeline {
 
-    agent {
-        label "default"
-    }
+    agent none
 
     stages {
-
-        stage('Generate Tarball') {
-            steps {
-                cleanWs()
-
-                checkout scm
-
-                dir(path: 'clamav_documentation') {
-                    git(url: 'https://github.com/Cisco-Talos/clamav-documentation.git', branch: "gh-pages")
-                }
-
-                dir(path: 'docs/html') {
-                    sh """# Move the clamav-documentation here.
-                        cp -r ../../clamav_documentation/* .
-                        # Clean-up
-                        rm -rf ../../clamav_documentation
-                        rm -rf .git .nojekyll CNAME Placeholder || true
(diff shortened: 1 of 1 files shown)


val-ms:
Updated after Jenkins rejected an `agent` directly on the parallel `Tests` stage. The throttle and `default` agent now belong to a parent sequential stage containing all existing stages. This holds the throttle for the complete run while preserving the existing Tests parallelism. Jenkins build #2 has parsed the amended revision and is now pending.

rsundriyal (approved):

PR #1783, opened 2026-07-29
libclamav: Harden XLM formula record parsing

jhumlick:
Validate XLM formula record sizes before parsing and honor the declared token-stream length instead of parsing all trailing record data.

Correct the FORMULA record accounting and add bounds checks for token operands, strings, attributes, function arguments, and cell references. Malformed or truncated tokens now terminate parsing without consuming data beyond the declared formula buffer.

Add unit and clamscan regression tests covering boundary-length records, truncated tokens, malformed strings, and scan-map edge cases.

Reported by Alessandro Greco (GitHub user 'aleff-github').

CLAM-3011

### DIFF
--- a/unit_tests/clamscan/xlm_test.py
+++ b/unit_tests/clamscan/xlm_test.py
@@ -121,6 +121,7 @@ class TC(testcase.TestCase):
     def test_formula_record_boundaries(self):
         self.step_name('Test XLM FORMULA records at the BIFF8 size boundary')
 
+        valid_string = b'XLM_VALID_FORMULA'
         testfiles = [
             TC.path_tmp / 'formula-8228.xls',
             TC.path_tmp / 'formula-8227.xls',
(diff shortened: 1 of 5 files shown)


unit_tests/clamscan/xlm_test.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        assert output.err.count('[cli_extract_xlm_macros_and_images] Extracting macros to') == len(testfiles) - 1
+
+        scanmap_command = '{} {}'.format(
+            TC.check_xlm_scanmap,
+            ' '.join(str(path) for path in testfiles),
+        )
+        scanmap_output = self.execute_command(scanmap_command)
+
+        assert scanmap_output.ec == 0
```
val-ms, line 174: Please add a positive assertion proving that at least one valid FORMULA record was successfully disassembled and its extracted output was scanned. As written, every generated workbook is expected to return OK; the debug-line count only proves that XLM extraction started, and check_xlm_scanmap only checks for CL_SUCCESS. This would all still pass if parse_formula() rejected every formula.

One robust approach would be to generate a valid ptgStr whose disassembly contains a distinctive value such as ptgStrA, write a temporary signature matching that normalized extracted output (the raw workbook would not contain that ASCII sequence), and assert detection through clamscan. Ideally, load the same test signature in the scan-map helper and assert CL_VIRUS there as well.

val-ms (changes requested):

val-ms (approved):
Code looks good and local testing went well. I'm waiting for it to run through Jenkins, but I'm happy with it.

aleff-github:
Thanks for the credits <3

PR #1789, opened 2026-08-03
CMake: Fix Windows shared dependency handling

val-ms:
## Summary

- use Windows import libraries, rather than runtime DLLs, as Rust linker inputs
- model dynamic libcheck imports with separate `.lib` and `.dll` locations
- search dependency target directories and sibling `bin` directories when collecting runtime DLLs
- keep PCRE2 in module discovery when config discovery is preferred globally
- derive Windows package architecture names for non-Visual Studio generators
- rerun the Rust build script when dependency-library variables change

## Problem

Windows dependency packages conventionally install import libraries under `lib` and runtime DLLs under `bin`. Some of ClamAV's CMake wiring used `TARGET_FILE` where a linker input was required and searched only one side of that layout while collecting runtime dependencies.

With Curl's exported shared target, `TARGET_FILE` supplied `bin/libcurl.dll` to `libclamav_rust/build.rs`. The build script converted that filename into a request for `libcurl.lib`, but the actual import library is `lib/libcurl_imp.lib`. The resulting Rust test link failed with `LNK1181`.

Dynamic libcheck installations had a related issue: the existing fallback target represented `checkDynamic.lib` as an unknown imported target and did not model or validate the corresponding runtime DLL.

## Solution

Use `TARGET_LINKER_FILE` for the dependency paths passed to Rust. Represent a manually discovered dynamic libcheck target as `SHARED IMPORTED`, with `IMPORTED_IMPLIB` pointing to the `.lib` file and `IMPORTED_LOCATION` pointing to the DLL. Runtime dependency collection now checks both target runtime directories and the sibling `bin` directories derived from linker-file locations.

The patch also keeps PCRE2 in module mode alongside LibXml2, supplies a package-architecture fallback for generators that do not define `CMAKE_VS_PLATFORM_NAME`, and declares the Rust build script's library environment dependencies.

## Validation

- `git diff --check` passes
- clean x64 build completed all 394 targets
- all six x64 CTest groups passed
- x64 ZIP and MSI packages generated successfully from a dependency prefix with runtime DLLs under `bin`
- dynamic Check discovery resolved `lib/checkDynamic.lib` and `bin/checkDynamic.dll` separately
- the complete 13-recipe Mussels ARM64 dependency chain built successfully
- current ClamAV source built and installed for ARM64; `clamscan.exe` reports machine type `AA64`
- ARM64 ZIP and MSI packages generated successfully

ARM64 tests were compiled but not executed because validation ran on an x64 Windows host.

CLAM-3051

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,7 +349,29 @@ if (WIN32)
 
     set(CPACK_GENERATOR                  "ZIP;WIX")
     set(CPACK_PACKAGE_VENDOR             "Cisco Systems, Inc.")
-    set(CPACK_PACKAGE_FILE_NAME ${CPACK_PACKAGE_FILE_NAME}.win.${CMAKE_VS_PLATFORM_NAME})
+    # Visual Studio generators provide CMAKE_VS_PLATFORM_NAME, but Ninja does
+    # not. Derive the package architecture from the compiler when necessary so
+    # documented Ninja builds do not produce package names ending in "win.".
+    set(CLAMAV_PACKAGE_ARCHITECTURE "${CMAKE_VS_PLATFORM_NAME}")
+    if(NOT CLAMAV_PACKAGE_ARCHITECTURE)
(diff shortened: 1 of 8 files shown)


val-ms:
@codex review

PR #1801, opened 2026-08-14
libfreshclam: fix use-after-free in fc_cleanup on init failure

qxp930712:
fc_cleanup() frees g_signVerifier but does not set it to NULL. When fc_initialize() fails, fc_cleanup() is called internally and then again by freshclam at exit, causing a double free that results in a segmentation fault.

Set g_signVerifier to NULL after freeing it, matching the pattern used by other pointers in fc_cleanup().

Triggered when freshclam cannot write to the database directory.


val-ms (approved):
Thank you for the fix! This looks great

PR #1803, opened 2026-08-16
libclamav: report skipped PCRE evaluation through AlertExceedsMax

### LINKED ISSUE #1785: PCREMaxFileSize (default 100MB) silently skips PCRE signature evaluation — verdict stays "OK", and AlertExceedsMax does NOT cover this case
PCREMaxFileSize (default 100MB) silently skips PCRE signature evaluation — verdict stays "OK", and AlertExceedsMax does NOT cover this case

## Summary

`PCREMaxFileSize` (default 100MB, distinct from `MaxFileSize`/`MaxScanSize`) causes
ClamAV to silently skip ALL PCRE-based signature evaluation for files above the
threshold — including PCRE signatures used for exploit detection embedded in images
and other complex content. The scan verdict still reports `OK` (clean), and unlike
`MaxFileSize`/`MaxScanSize`, this is **not** surfaced even with `AlertExceedsMax yes`.

## How we found this

Benchmarking scan time across file sizes (75MB/150MB/etc.), we noticed a
counter-intuitive inversion: a 150MB file consistently scanned FASTER than a 75MB
file of similar content profile. Investigated with `clamscan --statistics=pcre`:

- Run under 100MB (75MB file): statistics table showed ~443 real PCRE subsignature
  evaluations.
- Run over 100MB (150MB file): statistics table was empty — zero PCRE rules were
  evaluated, confirming the size threshold silently disabled the entire matcher
  for that scan, with verdict still `OK`.

## Root cause (traced in source)

- `libclamav/matcher.c:214-221` — when `PCREMaxFileSize` is exceeded, the code logs
  via `cli_dbgmsg()` only (invisible without `--debug`) and returns `CL_EMAXSIZE`.
- `libclamav/scanners.c:~4483` — `CL_EMAXSIZE` falls into the generic "MAX
  conditions" bucket (alongside `CL_EMAXREC`/`CL_EMAXFILES`) that's explicitly
  converted to `CL_SUCCESS` so the scan "finishes as much as is allowed" — no
  alert is raised.
- Compare to `MaxFileSize`, which has explicit call sites
  (`scanners.c:6523`/`6637`) invoking
  `engine->cb_virus_found(..., "Heuristics.Limits.Exceeded.MaxFileSize", ...)` —
  that's what `AlertExceedsMax` actually hooks into. **No equivalent
  `"Heuristics.Limits.Exceeded.PCRE..."` call site exists anywhere in the
  codebase** (confirmed via code search) — so `AlertExceedsMax yes` does not help
  here, even though `PCREMaxFileSize` sits right next to `MaxFileSize`/
  `MaxScanSize` in `clamd.conf` and looks like it should behave the same way.

## Impact

For any deployment scanning large legitimate files (we scan user-uploaded
documents up to several hundred MB), this means files above 100MB get
systematically weaker detection than smaller ones, with zero operational signal —
`--statistics=pcre` is the only way we found to detect it, and it's not part of
default output, logs, or `AlertExceedsMax`.

## Suggestion

Based on our validation above, it might be worth considering whether
`PCREMaxFileSize` should raise `Heuristics.Limits.Exceeded.PCREMaxFileSize`
(mirroring how `MaxFileSize` already does), so `AlertExceedsMax` could cover this
case too — though there may be a reason it was left out that we're not aware of.
If that's intentional, it might still help to note in
`clamd.conf.sample`/the man page that `PCREMaxFileSize` behaves differently from
its neighbors in this respect. Happy to be corrected if we're missing context.

## Environment

ClamAV 1.5.x, clamd (daemon), Debian bookworm.

AAH20:
## Summary

Report `Heuristics.Limits.Exceeded.PCREMaxFileSize` through the existing
`AlertExceedsMax` mechanism when the PCRE matcher skips subsignature evaluation
because a file or buffer exceeds `PCREMaxFileSize`.

Closes #1785.

## Root cause

Both PCRE size-limit branches return `CL_EMAXSIZE`. The generic scanner result
handler intentionally converts non-fatal max conditions to success so remaining
analysis can continue. Unlike `MaxFileSize` and `MaxScanSize`, the PCRE-specific
limit did not first append an exceeds-max heuristic or metadata record. A scan
could therefore report `OK` without an operational indication that PCRE
subsignatures were not evaluated.

## Changes

- Report the limit through `cli_append_potentially_unwanted_if_heur_exceedsmax`
  for both fmap and buffer PCRE paths.
- Preserve existing behavior when `AlertExceedsMax` is disabled.
- Add regression coverage for alert enabled, alert disabled, and normal
  below-limit PCRE evaluation.
- Document that `AlertExceedsMax` covers `PCREMaxFileSize` and that exceeding it
  reduces detection coverage.

## Validation

- Full native ClamAV debug build completed successfully on macOS arm64.
- `python3 -m unittest -v clamscan.regex_test`: 6 tests passed.
- New regression test confirms:
  - above limit + alert enabled: heuristic reported, exit 1;
  - above limit + alert disabled: compatibility behavior retained, exit 0;
  - below limit: PCRE signature evaluated and detected.
- `git diff --check` passed.

### DIFF
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -339,7 +339,7 @@ void help(void)
     mprintf(LOGG_INFO, "    --alert-encrypted-archive[=yes/no(*)] Alert on encrypted archives.\n");
     mprintf(LOGG_INFO, "    --alert-encrypted-doc[=yes/no(*)]    Alert on encrypted documents.\n");
     mprintf(LOGG_INFO, "    --alert-macros[=yes/no(*)]           Alert on OLE2 files containing VBA macros.\n");
-    mprintf(LOGG_INFO, "    --alert-exceeds-max[=yes/no(*)]      Alert on files that exceed max file size, max scan size, or max recursion limit.\n");
+    mprintf(LOGG_INFO, "    --alert-exceeds-max[=yes/no(*)]      Alert on files that exceed max file size, max scan size, max recursion, or PCRE max file size.\n");
     mprintf(LOGG_INFO, "    --alert-phishing-ssl[=yes/no(*)]     Alert on emails containing SSL mismatches in URLs.\n");
     mprintf(LOGG_INFO, "    --alert-phishing-cloak[=yes/no(*)]   Alert on emails containing cloaked URLs.\n");
     mprintf(LOGG_INFO, "    --alert-partition-intersection[=yes/no(*)] Alert on raw DMG image files containing partition intersections.\n");
--- a/clamscan/clamscan.c
+++ b/clamscan/clamscan.c
@@ -339,7 +339,7 @@ void help(void)
     mprintf(LOGG_INFO, "    --alert-encrypted-archive[=yes/no(*)] Alert on encrypted archives.\n");
     mprintf(LOGG_INFO, "    --alert-encrypted-doc[=yes/no(*)]    Alert on encrypted documents.\n");
     mprintf(LOGG_INFO, "    --alert-macros[=yes/no(*)]           Alert on OLE2 files containing VBA macros.\n");
-    mprintf(LOGG_INFO, "    --alert-exceeds-max[=yes/no(*)]      Alert on files that exceed max file size, max scan size, max recursion, or PCRE max file size.\n");
+    mprintf(LOGG_INFO, "    --alert-exceeds-max[=yes/no(*)]      Alert on files that exceed configured scan limits.\n");
     mprintf(LOGG_INFO, "    --alert-phishing-ssl[=yes/no(*)]     Alert on emails containing SSL mismatches in URLs.\n");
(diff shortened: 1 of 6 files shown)


val-ms (changes requested):
I am open to this change, although I would ask that you not call out the PCRE limit in the documentation/config/help-string. We already have similar alerts for other limits, not just the size and time limits. You can find a short list in the older release notes: https://github.com/Cisco-Talos/clamav/blob/main/NEWS.md#01041 So if anything, those documentation/config/help-string messages should really be more general about alerts on exceeded limits overall because I wouldn't want to call out every possible limit in those places.

AAH20 (commented):
Thanks @val-ms — good call. I've generalized all four doc/help-string/config-sample locations so they no longer single out PCREMaxFileSize (or any individual limit). The libclamav reporting itself is unchanged — it still emits `Heuristics.Limits.Exceeded.PCREMaxFileSize` through the existing `AlertExceedsMax` path, matching the pattern the other limit heuristics already use.

Happy to iterate further.

PR #1816, opened 2026-09-02
libclamav: replace the LZMA SDK with Rust decoders

jhumlick:
Replace the vendored C LZMA SDK with lzma-rust2 and sevenz-rust2 implementations for 7z archives, raw LZMA streams, and standalone XZ files. Add support for bzip2-compressed 7z entries while preserving the existing C-facing raw-LZMA interface used by embedded format scanners.

Stream decompressed output in bounded 64 KiB chunks and enforce scan limits using bytes actually produced instead of archive-declared sizes. Scan partial output before reporting decoder errors, and continue with later independent 7z blocks after a block-local size limit.

Harden the vendored decoders against untrusted allocations, falsified sizes, malformed coder graphs, integer overflow, excessive dictionary memory, and early callback termination. Bound decoder memory to 128 MiB and use a single decoding thread.

Preserve legacy 7z metadata ordering, indexes, compressed-size values, encrypted-entry metadata, and unsupported-codec handling. Return CL_EFORMAT for malformed, CRC-failing, truncated, and memory-limited archives, but defer that result so raw detections and more serious scan errors take precedence.

Move concatenated XZ scanning into Rust, explicitly finish decoding at physical EOF EOF, and scan every valid stream including inter-stream padding. Avoid temporary files for zero-length 7z output and correct Windows CRT descriptor ownership for extracted temporary files.

Contain panics at every new Rust FFI boundary and add regression coverage for codecs, actual-output limits, false sizes, later-entry scanning, metadata, encryption, malformed archives, concatenated XZ streams, temporary-file lifecycle, and panic handling.

CLAM-2123


libclamav/scanners.c
```diff
@@ -5173,8 +5076,20 @@ cl_error_t cli_magic_scan(cli_ctx *ctx, cli_file_t type)
             break;
 
         case CL_TYPE_7Z:
-            if (SCAN_PARSE_ARCHIVE && (DCONF_ARCH & ARCH_CONF_7Z))
+            if (SCAN_PARSE_ARCHIVE && (DCONF_ARCH & ARCH_CONF_7Z)) {
                 ret = cli_7unz(ctx, 0);
+                if (CL_EFORMAT == ret) {
+                    /*
+                     * The former 7z SDK scanner logged malformed, CRC,
+                     * truncated, and decoder-memory errors and returned clean.
```
val-ms, line 5084: Probably don't need to mention what the previous implementationd did, but I do appreciate the comment explaining why error handling is deferred. 

That said, I think the method of deferring an error adds some risk of losing the error if there's a `goto done` in between here and the error-restoration logic. Worse--it risks making a non-serious error into a fatal error. The `result_should_goto_done()` ignores non-fatal error codes. 

If the previous implementation previously returned clean/success after logging errors, we should consider doing same here. Or doing the logging and error filtering in the rust code.

libclamav_rust/vendor/lzma-rust2/.github/workflows/cifuzz.yml
val-ms, line 1: We shouldn't vendor the lzma-rust2 or sevenz rust code. Just add them as dependencies in the libclamav_rust Cargo.toml

val-ms (changes requested):

libclamav/scanners.c
```diff
@@ -5609,6 +5524,10 @@ cl_error_t cli_magic_scan(cli_ctx *ctx, cli_file_t type)
         status = CL_SUCCESS;
     }
 
+    if ((CL_SUCCESS == status) && (CL_SUCCESS != deferred_format_error)) {
+        status = deferred_format_error;
```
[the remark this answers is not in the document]
val-ms: Oh, codex had the same concern I did... but like 45 minutes prior. Whoops.

PR #1818, opened 2026-09-03
libclamav: avoid duplicate HandlerType alerts after retyping

val-ms:
## Summary

HandlerType logical signatures could alert again when the retyped child layer was evaluated, including duplicate alerts for Target:0 signatures. The child scan now returns clean for the already-retyped layer instead of falling through to the alert path.

Adds a clamscan regression using a Target:0 HandlerType signature with --allmatch.

Jira: CLAM-3096

## Testing

- ctest --test-dir build -V -R '^clamscan$'\n- Exact reproducer scans clean with no FOUND output\n\n## Backport\n\nPlease backport this fix to the supported release branches.

### DIFF
--- a/libclamav/matcher.c
+++ b/libclamav/matcher.c
@@ -936,6 +936,10 @@ static cl_error_t lsig_eval(cli_ctx *ctx, struct cli_matcher *root, struct cli_a
             (void)cli_recursion_stack_pop(ctx); /* Restore the parent fmap */
 
             goto done;
+        } else {
+            // The retyped layer was already scanned as the requested type. Do not alert on it again.
+            status = CL_SUCCESS;
+            goto done;
         }
     }
 
--- a/libclamav/matcher.c
+++ b/libclamav/matcher.c
@@ -925,6 +925,13 @@ static cl_error_t lsig_eval(cli_ctx *ctx, struct cli_matcher *root, struct cli_a
                 goto done;
             }
 
+            /*
+             * The clean-cache key does not include the HandlerType. Do not let
+             * this retyped scan cache a result that could suppress a later scan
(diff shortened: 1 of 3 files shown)


jhumlick (changes requested):
[P1] Prevent clean-cache poisoning for retyped layers — [libclamav/matcher.c:941](https://github.com/Cisco-Talos/clamav/blob/08f1fdce32ff3bfc5db584db28231abf51645e7a/libclamav/matcher.c#L941)
Returning CL_SUCCESS makes the retyped child eligible for the clean cache. The cache key contains only the content hash, size, and recursion level—not the forced HandlerType. A subsequent handler scan at the same depth can therefore be skipped, missing type-specific detections. In a focused reproducer, the PR reports OK, while the same scan with --disable-cache reports the expected Target:1 detection.

val-ms:
Thanks for catching this. You were right: returning clean for a HandlerType match on the retyped child allowed that child to populate the clean cache, and the next HandlerType scan of the same bytes could then be skipped before its type-specific signatures ran.

I fixed this in eed64fd94 by making HandlerType-retyped layers bypass clean-cache lookup and marking the duplicated fmap as non-cacheable so it cannot populate the cache either. I also added a regression with ZIP and PE HandlerType signatures followed by a PE Target:1 signature; it fails as a false negative without the cache fix and now detects correctly while continuing to suppress the duplicate HandlerType alerts.

Validation: the clamscan build succeeded and `ctest --test-dir build -V -R '^clamscan$'` passed all 136 tests plus 2 subtests. When you have a chance, I would appreciate a re-review.

jhumlick (approved):

PR #1819, opened 2026-09-07
clamonacc: watch a whole filesystem so writes from other mount namespaces are scanned

AdzerKI:
A mount mark reports events only for the mount it was placed on. A process in its own mount namespace reaches the same superblock through its own copy of the mount, so `OnAccessMountPath` yields no event and the file is never scanned. That silently covers containers and any systemd unit with private mounts — on our hosting box it was every process that writes into the web roots, while `clamonacc` stayed `active` and reported nothing.

Reproducer on a tree watched with `OnAccessMountPath`:

```
# event delivered, file scanned
python3 -c 'open("/watched/plain","w").write(EICAR)'

# same write from its own mount namespace, no event
unshare -m --propagation private python3 -c 'open("/watched/nsprobe","w").write(EICAR)'
```

`FAN_MARK_FILESYSTEM` (Linux 4.20) marks the superblock, and the event arrives whichever mount the writer used. This adds `OnAccessFilesystemPath`, taking the same argument as `OnAccessMountPath` and watching the whole filesystem containing it: notification only, DDD preempted, prevention disallowed. Builds whose headers lack the flag keep compiling and report the option as unavailable.

### DIFF
--- a/clamonacc/fanotif/fanotif.c
+++ b/clamonacc/fanotif/fanotif.c
@@ -82,13 +82,16 @@ cl_error_t onas_setup_fanotif(struct onas_context **ctx)
     onas_fan_fd      = (*ctx)->fan_fd;
     (*ctx)->fan_mask = fan_mask;
 
-    if (optget((*ctx)->clamdopts, "OnAccessPrevention")->enabled && !optget((*ctx)->clamdopts, "OnAccessMountPath")->enabled) {
+    int whole_fs = optget((*ctx)->clamdopts, "OnAccessMountPath")->enabled ||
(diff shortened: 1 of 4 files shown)


AdzerKI:
Both points addressed.

Sizing now goes through the event descriptor for fanotify events, so a file reached through a mount that has a different path — or none — in clamonacc's namespace is no longer skipped with ENOENT; the scan itself already passes that descriptor with `--fdpass`.

The two mark kinds are installed independently, so a config carrying `OnAccessMountPath` and `OnAccessFilesystemPath` together gets both instead of silently dropping the filesystem marks.

